PR #34, opened 2026-03-20
fix: filter JS/TS builtins, enable callers_of, improve multi-word search

Samian89:
Three fixes that make the graph more useful for JavaScript/TypeScript projects:

  1. callers_of now works — Was broken for all users. CALLS edges store unqualified names (generateTestCode) but
  callers_of searched by fully qualified name (file.ts::generateTestCode), so it always returned 0 results. Added
   fallback that matches by plain function name.
  2. Builtin noise filtered at query-time, not parse-time — Asking "who calls .map()?" returns hundreds of
  useless hits. Common JS/TS builtins are now skipped in callers_of queries. But they're kept in the graph so
  callees_of still shows them — knowing a function calls .map() and .filter() helps the AI understand what it
  does.
  3. Multi-word keyword search — "whisper transcribe" returned 0 results because it searched for the exact
  string. Now splits into individual terms and matches ANY, ranked by match count.

  What changed

  - graph.py — Added search_edges_by_target_name() + multi-word search_nodes
  - tools.py — callers_of fallback to unqualified name + builtin skip list
  - parser.py — No changes (builtins kept in graph)

### DIFF
--- a/code_review_graph/graph.py
+++ b/code_review_graph/graph.py
@@ -268,6 +268,21 @@ class GraphStore:
         ).fetchall()
         return [self._row_to_edge(r) for r in rows]
 
+    def search_edges_by_target_name(self, name: str, kind: str = "CALLS") -> list[GraphEdge]:
+        """Search for edges where target_qualified matches an unqualified name.
+
+        CALLS edges often store unqualified target names (e.g. ``generateTestCode``)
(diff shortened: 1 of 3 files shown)


tirth8205:
Checking whether callers_of builtin filtering and unqualified name fallback are in main — these are present (search_edges_by_target and _JS_BUILTINS found in query.py). This PR appears to have been integrated. No revival needed.

PR #37, opened 2026-03-20
fix: support multi-word keyword search with AND logic

n24q02m:
## Summary

`search_nodes()` currently treats multi-word queries like `"firebase auth"` as a single literal substring, which matches nothing because no node name contains that exact phrase.

This PR splits the query into individual words and requires **all** words to match (case-insensitive) against the node `name` or `qualified_name`.

**Before:** `search_nodes("firebase auth")` returns 0 results.

**After:** `search_nodes("firebase auth")` matches:
- `verify_firebase_token` (contains "firebase" and "auth" via "token" -> no, via name match)
- `FirebaseAuth` (contains "firebase" and "auth")

Single-word queries behave identically to before.

## Changes

- `code_review_graph/graph.py`: Modified `search_nodes()` to split query into words and build AND conditions.

## Test plan

- [ ] `search_nodes("firebase auth")` returns nodes containing both words
- [ ] `search_nodes("RAG pipeline")` returns empty when no node contains both words
- [ ] Single-word queries like `search_nodes("firebase")` still work as before
- [ ] Existing tests pass

### DIFF
--- a/code_review_graph/graph.py
+++ b/code_review_graph/graph.py
@@ -275,12 +275,29 @@ class GraphStore:
         return [r["file_path"] for r in rows]
 
     def search_nodes(self, query: str, limit: int = 20) -> list[GraphNode]:
-        """Simple keyword search across node names."""
-        pattern = f"%{query}%"
-        rows = self._conn.execute(
-            "SELECT * FROM nodes WHERE name LIKE ? OR qualified_name LIKE ? LIMIT ?",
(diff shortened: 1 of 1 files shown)


tirth8205:
Multi-word keyword search with AND logic is already in main — graph.py:search_nodes() at line 312 uses words = query.lower().split() with AND conditions. This PR has been integrated. No revival needed.

PR #38, opened 2026-03-20
fix: resolve same-file call targets to qualified names

### LINKED ISSUE #20: query_graph callers_of/callees_of returns empty — edge targets stored as bare names
## Summary

`query_graph("callers_of", ...)` and related relational queries return 0 results even when the graph contains matching CALLS edges. The root cause is that call edge targets are stored as bare function names (e.g., `createTask`) rather than qualified names (e.g., `/path/to/file.ts::createTask`), so the `get_edges_by_target(qualified_name)` lookup in `query_graph` never matches.

## Reproduction

```bash
# Clone any TypeScript project with cross-file function calls
git clone https://github.com/ericlitman/pipeline-test-1.git /tmp/test-repo
cd /tmp/test-repo
uvx code-review-graph build

# Query callers — returns 0 results
uvx --from code-review-graph python3 -c "
from code_review_graph.tools import query_graph
result = query_graph('callers_of', '/tmp/test-repo/src/models/task.ts::createTask', repo_root='/tmp/test-repo')
print(result)
"
```

Returns `"summary": "Found 0 result(s)"` despite the edge data clearly showing `createTaskRoutes` calls `createTask`.

## Root cause

Raw edge data in SQLite shows:
```
CALLS: /tmp/test-repo/src/routes/tasks.ts::createTaskRoutes -> createTask
```

The target is the bare name `createTask`, not the qualified name `/tmp/test-repo/src/models/task.ts::createTask`. When `query_graph` calls `store.get_edges_by_target(qualified_name)`, the qualified name doesn't match the bare name stored in the edge.

The same issue affects `importers_of` — import edge targets are stored as relative paths (`../models/task`) rather than resolved qualified names.

Note: `get_impact_radius` is NOT affected because it uses a different traversal path (NetworkX graph).

## Affected queries

- `callers_of` — always returns empty
- `importers_of` — always returns empty
- `tests_for` (via TESTED_BY edges) — no edges created (see related issue)

## Expected behavior

Edge targets should be resolved to qualified names during parsing, matching the format used in node `qualified_name` fields. Alternatively, `query_graph` should fall back to matching against bare names when qualified name lookup fails.

## Environment

- code-review-graph v1.8.2
- Python 3.14.3
- TypeScript codebase (Bun + Hono)
- macOS

n24q02m:
## Summary

Fixes #20.

`callers_of` and `callees_of` queries return empty results because the parser stores CALLS edges with **bare** target names (e.g. `FirebaseAuth`) while graph queries look up **qualified** names (e.g. `main.go::FirebaseAuth`).

This PR adds a post-parse resolution step that builds a symbol table from parsed nodes and qualifies bare call targets that match local definitions.

**Before:** `callers_of("main.go::FirebaseAuth")` returns 0 results, even though `setupRoutes()` calls `FirebaseAuth()` in the same file.

**After:** The CALLS edge target is resolved from `FirebaseAuth` to `main.go::FirebaseAuth`, so `callers_of` correctly returns `setupRoutes`.

External calls (names not defined in the file) remain bare -- this is correct since they reference symbols in other files or libraries.

## Changes

- `code_review_graph/parser.py`:
  - Added `_resolve_call_targets()` method that builds a symbol table and qualifies bare targets
  - Called at the end of `parse_bytes()` after tree extraction

## Test plan

- [ ] Go file with `func FirebaseAuth()` and `func setupRoutes() { FirebaseAuth() }` produces a CALLS edge with qualified target `main.go::FirebaseAuth`
- [ ] Python file with local function calls produces qualified CALLS targets
- [ ] External library calls (e.g. `json.loads()`) remain bare (no `::` in target)
- [ ] Existing tests pass

### DIFF
--- a/code_review_graph/parser.py
+++ b/code_review_graph/parser.py
@@ -247,7 +247,49 @@ class CodeParser:
             tree.root_node, source, language, file_path_str, nodes, edges
         )
 
+        # Resolve bare call targets to qualified names using same-file definitions
+        edges = self._resolve_call_targets(nodes, edges, file_path_str)
+
         return nodes, edges
+
+    def _resolve_call_targets(
+        self,
+        nodes: list[NodeInfo],
+        edges: list[EdgeInfo],
+        file_path: str,
+    ) -> list[EdgeInfo]:
+        """Resolve bare call targets to qualified names using same-file definitions.
+
+        After parsing, CALLS edges store bare function names (e.g. ``FirebaseAuth``)
+        as targets. This method builds a symbol table from the parsed nodes and
+        qualifies any bare target that matches a local definition, so that
+        ``callers_of`` / ``callees_of`` queries produce correct results.
+
+        External calls (names not defined in this file) remain bare.
+        """
+        # Build symbol table: bare_name -> qualified_name
+        symbols: dict[str, str] = {}
+        for node in nodes:
+            if node.kind in ("Function", "Class", "Type", "Test"):
+                bare = node.name
+                qualified = self._qualify(bare, file_path, node.parent_name)
+                if bare not in symbols:
+                    symbols[bare] = qualified
+
+        resolved: list[EdgeInfo] = []
+        for edge in edges:
+            if edge.kind == "CALLS" and "::" not in edge.target:
+                if edge.target in symbols:
+                    edge = EdgeInfo(
+                        kind=edge.kind,
+                        source=edge.source,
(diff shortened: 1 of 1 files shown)


tirth8205:
Same-file call target resolution (_resolve_call_targets) is already in main — parser.py has _resolve_call_targets at line 822 and it's called at lines 376 and 692. This PR has been integrated. No revival needed.

PR #39, opened 2026-03-20
feat: add output pagination to impact radius

n24q02m:
## Summary

On large codebases, `get_impact_radius` can return thousands of nodes, producing 500K+ character MCP responses that blow up context windows and cause timeouts.

This PR adds output pagination with truncation metadata:
- `max_results` parameter (default: 500) caps impacted nodes returned
- `truncated` boolean flag indicates when results are incomplete
- `total_impacted` integer shows the full count before truncation
- Summary includes truncation notice when active

The BFS traversal itself already had a `max_nodes` parameter in `graph.py` but it was not exposed to the tools layer, and the results lacked truncation metadata.

## Changes

- `code_review_graph/graph.py`: `get_impact_radius()` now returns `truncated` and `total_impacted` in results dict
- `code_review_graph/tools.py`: `get_impact_radius()` adds `max_results` parameter, passes to graph layer, includes truncation info in summary and response

## Test plan

- [ ] Impact radius with 100+ impacted nodes and `max_results=10` returns exactly 10 nodes with `truncated=True`
- [ ] Impact radius with few impacted nodes returns all with `truncated=False`
- [ ] Default `max_results=500` does not change behavior for small codebases
- [ ] Existing tests pass

### DIFF
--- a/code_review_graph/graph.py
+++ b/code_review_graph/graph.py
@@ -346,11 +346,17 @@ class GraphStore:
             if node:
                 impacted_nodes.append(node)
 
+        # Truncation: cap impacted nodes and report total
+        total_impacted = len(impacted_nodes)
+        truncated = total_impacted > max_nodes
+        if truncated:
+            impacted_nodes = impacted_nodes[:max_nodes]
+
         impacted_files = list({n.file_path for n in impacted_nodes})
 
         # Collect relevant edges in a single batch query
         relevant_edges = []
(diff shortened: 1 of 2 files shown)


tirth8205:
Impact radius pagination (max_results parameter, truncated flag, total_impacted count) is already in main — graph.py has truncated and total_impacted at lines 458 and 525. This PR has been integrated. No revival needed.

Issue #46, opened 2026-03-20
Build & Embed operations hang silently on Windows 11 with uvx


pvnsai73:
<html>
<body>
<hr>
<h1>Bug Report — code-review-graph on Windows 11</h1>
<p><strong>Build and embed operations hang silently. No error, no timeout. Affects a ~6-file JS project (~240 nodes).</strong></p>
<hr>
<h2>Environment</h2>

Property | Value
-- | --
OS | Windows 11 Enterprise 10.0.26200
Architecture | x86_64-pc-windows-msvc
uvx | 0.10.12 (build 00d72dac7, 2026-03-19)
Python runtime | cpython-3.14.3 (uvx-managed; no system Python present)
code-review-graph | 1.8.4 with [embeddings] extra
sentence-transformers | 3.4.1
PyTorch | 2.10.0
transformers | 4.57.6
Claude Code plugin | 1.8.3 (marketplace)
Required env var | UV_NATIVE_TLS=1 (without this, MCP server fails on corporate network)

<hr>
<h2>Suggested Improvements</h2>
<ol>
<li><strong>Add progress indicators and timeouts</strong> for both <code>build</code> and <code>embed</code> so stalls surface rather than silently hang.</li>
<li><strong>Document the <code>[embeddings]</code> extra requirement</strong> in the README — especially for Windows/uvx users who might install sentence-transformers separately.</li>
<li><strong>Document <code>UV_NATIVE_TLS=1</code></strong> as a known requirement for corporate Windows environments.</li>
<li><strong>Investigate the node/edge discrepancy</strong> (310→242 nodes, 3,560→1,808 edges) — the reconciliation phase is likely where the build stall originates.</li>
</ol></body></html>
</body>
</html>

JF10R:
Temporary workaround: Ask Claude Code to invoke the commands through a bash command instead of directly.

tirth8205:
This is the Windows async event loop hang — same root cause as #136. On Windows with ProactorEventLoop, synchronous MCP tool handlers block the entire event loop during build and embed operations.

PR #49 proposed wrapping all tool handlers with loop.run_in_executor() to fix this but was closed without merging.

The corporate TLS workaround (UV_NATIVE_TLS=1) is a valid configuration tip for Windows corporate environments. The wrong install method note (use [embeddings] extra, not separate pip install) is also important.

This remains an open bug primarily affecting Windows users.

tirth8205:
[...]

PR #49, opened 2026-03-20
fix(mcp): unblock async event loop and fix progress logging API

e-savca:
## Problem
All 9 MCP tool handlers were synchronous functions called directly from
FastMCP's async event loop. This blocked the event loop for the duration
of every SQLite query, file parse, and git subprocess call — especially
problematic on Windows (ProactorEventLoop).

Additionally, `build_or_update_graph_tool` used `ctx.session.send_log_message()`
which does not exist in FastMCP, causing an AttributeError on every call.

## Changes
- Wrap all 9 tool handlers with `loop.run_in_executor()` so blocking work
  runs in a thread pool without freezing the event loop
- Add `on_progress` callback to `full_build` and `incremental_update` to
  stream build progress back to the MCP client
- Use a `asyncio.Queue` + `call_soon_threadsafe` bridge to safely relay
  progress messages from the worker thread to the event loop
- Replace `ctx.session.send_log_message("info", msg)` with `ctx.info(msg)`
  (correct FastMCP Context API)
- Fix `ctx: Context = None` annotation to `ctx: Context | None = None`

### DIFF
--- a/code_review_graph/main.py
+++ b/code_review_graph/main.py
@@ -40,7 +40,7 @@ async def build_or_update_graph_tool(
     full_rebuild: bool = False,
     repo_root: Optional[str] = None,
     base: str = "HEAD~1",
-    ctx: Context = None,
+    ctx: Context | None = None,
 ) -> dict:
     """Build or incrementally update the code knowledge graph.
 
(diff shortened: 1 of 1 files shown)


tirth8205:
The async event loop fix (run_in_executor for blocking MCP tool handlers) is NOT in main — main.py has no run_in_executor calls. This PR addresses the Windows hang bug (#46, #136) where synchronous handlers block ProactorEventLoop. Worth reviving — this is a real bug affecting Windows users. The on_progress callback and asyncio.Queue bridge for streaming build progress are also valuable additions.

PR #58, opened 2026-03-23
feat: add .mjs/.astro extension support and fix IMPORTS_FROM path resolution

zoneghost7:
## Summary

Two changes to `parser.py`:

### 1. Extension mappings for `.mjs` and `.astro` files

- **`.mjs` → `javascript`**: Standard ES module files (`.mjs`) are widely used for test suites (`*.test.mjs`), configs (`astro.config.mjs`, `rollup.config.mjs`), and any project preferring explicit ESM. These were silently skipped during parsing. Since `.mjs` is standard JavaScript with ES module semantics, mapping to `javascript` is correct.

- **`.astro` → `typescript`**: [Astro](https://astro.build/) framework files have TypeScript frontmatter between `---` fences followed by an HTML template section. Tree-sitter's TypeScript grammar parses the frontmatter cleanly — extracting imports, function definitions, and call expressions — while silently skipping the template section (no parse errors). This gives useful graph coverage (imports, functions, calls) for Astro projects without requiring a dedicated Astro grammar.

### 2. Fix IMPORTS_FROM edge path resolution

**Problem:** `IMPORTS_FROM` edges store the raw import string as `target_qualified` (e.g., `../../lib/data-service`). However, query patterns like `importers_of` and `find_dependents` look up edges by absolute file path. This mismatch causes:

- `importers_of` returns **zero results** for every file — the query searches for `/abs/path/to/file.ts` but the database contains `../../lib/data-service`
- `find_dependents()` in `incremental.py` fails to identify downstream files during incremental rebuilds — editing a core library file won't trigger re-parsing of its importers
- Blast radius analysis misses all `IMPORTS_FROM` traversal paths, only finding connections via `CALLS` edges

**Root cause:** `_resolve_module_to_file()` is called when creating `CALLS` edges (via `_resolve_call_target`) but is never called for `IMPORTS_FROM` edges.

**Fix:** Call `_resolve_module_to_file()` when creating `IMPORTS_FROM` edges (lines 601-607). If resolution succeeds, store the absolute path; if it fails (node built-ins like `node:fs`, external packages like `react`, virtual modules like `astro:content`), fall back to the raw string. This is the same graceful-fallback pattern used for `CALLS` edges.

**Before:**
```
importers_of('src/lib/data-service.ts') → 0 results
find_dependents('/abs/path/data-service.ts') → 0 files
```

**After:**
```
importers_of('src/lib/data-service.ts') → WorkLayout.astro, [slug].astro
find_dependents('/abs/path/data-service.ts') → 2 files
```

## Edge cases handled

| Import type | Resolution | Stored as |
|---|---|---|
| Relative: `../../lib/utils` | Resolves via `_resolve_module_to_file` | Absolute path |
| Relative with ext: `./Component.astro` | Resolves (exact match) | Absolute path |
| Index barrel: `./utils` (dir with index.ts) | Resolves (index file) | Absolute path |
| Node built-in: `node:fs` | Returns `None` → fallback | `node:fs` (raw) |
| External package: `react` | Returns `None` → fallback | `react` (raw) |
| Virtual module: `astro:content` | Returns `None` → fallback | `astro:content` (raw) |
| Missing file: `./deleted` | Returns `None` → fallback | `./deleted` (raw) |

## Test plan

- [ ] Full rebuild on a TypeScript/JavaScript project with `.mjs` test files — verify `.mjs` files appear in graph stats
- [ ] Full rebuild on an Astro project — verify `.astro` files are parsed, imports and functions extracted, no errors
- [ ] `importers_of` query returns correct results for files with relative imports
- [ ] `find_dependents` correctly identifies downstream files
- [ ] Blast radius for a library file includes files connected via `IMPORTS_FROM` edges
- [ ] External/unresolvable imports (`node:fs`, `react`, etc.) stored as raw strings (no regression)
- [ ] Incremental rebuild after editing a file correctly re-parses its dependents

### DIFF
--- a/code_review_graph/parser.py
+++ b/code_review_graph/parser.py
@@ -74,6 +74,8 @@ EXTENSION_TO_LANGUAGE: dict[str, str] = {
     ".php": "php",
     ".sol": "solidity",
     ".vue": "vue",
+    ".mjs": "javascript",
+    ".astro": "typescript",
 }
 
 # Tree-sitter node type mappings per language
@@ -599,10 +601,14 @@ class CodeParser:
             if node_type in import_types:
                 imports = self._extract_import(child, language, source)
                 for imp_target in imports:
+                    # Resolve relative/module imports to absolute file paths
+                    resolved = self._resolve_module_to_file(
+                        imp_target, file_path, language,
+                    )
                     edges.append(EdgeInfo(
                         kind="IMPORTS_FROM",
                         source=file_path,
-                        target=imp_target,
+                        target=resolved if resolved else imp_target,
                         file_path=file_path,
                         line=child.start_point[0] + 1,
                     ))


tirth8205:
.mjs and .astro extensions are already in main (parser.py has .mjs -> javascript and .astro -> typescript at lines 97-98). The IMPORTS_FROM path resolution fix may or may not be in main — checking whether tsconfig_resolver already handles this. The tsconfig_resolver.py is present in main, which handles alias resolution. This PR appears to be superseded.

PR #71, opened 2026-03-27
feat: add recurse_submodules option for git submodule support

weirdbb91:
Add opt-in `recurse_submodules` parameter to `build_or_update_graph`, `full_build`, `collect_all_files`, and `get_all_tracked_files` so that files inside git submodules are included in the knowledge graph.

When enabled, `git ls-files --recurse-submodules` is used instead of plain `git ls-files`.  The option can be set three ways (highest wins):

  1. Function parameter `recurse_submodules=True`
  2. Environment variable `CRG_RECURSE_SUBMODULES=true`
  3. Default: disabled (existing behavior unchanged)

This unblocks monorepo setups that use git submodules — previously the parent repo's `git ls-files` returned only gitlink entries for submodule directories, causing the graph builder to find zero parseable files.

Tests:
- 4 unit tests (param, default, env-var fallback, param-overrides-env)
- 4 integration tests with real git repos and submodules
- Full suite: 510 passed, 0 failed

### DIFF
--- a/code_review_graph/incremental.py
+++ b/code_review_graph/incremental.py
@@ -131,6 +131,13 @@ def _is_binary(path: Path) -> bool:
 
 _GIT_TIMEOUT = int(os.environ.get("CRG_GIT_TIMEOUT", "30"))  # seconds, configurable
 
+# When True, `git ls-files --recurse-submodules` is used so that files
+# inside git submodules are included in the graph.  Opt-in via env var;
+# can also be overridden per-call through function parameters.
+_RECURSE_SUBMODULES = os.environ.get(
+    "CRG_RECURSE_SUBMODULES", ""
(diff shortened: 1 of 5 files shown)


tirth8205:
git submodule support (recurse_submodules option) is not in main — the current parser.py and incremental.py do not have a recurse_submodules parameter. This is a valid unimplemented feature. Worth reviving — it has 4 unit tests and 4 integration tests, and uses git ls-files --recurse-submodules only when opted in, so existing behavior is unchanged.

weirdbb91:
Resubmitted as a [new PR](https://github.com/tirth8205/code-review-graph/pull/215) rebased on the latest main to resolve merge conflicts from the original branch. No code logic
  changes — only the base was updated.

PR #99, opened 2026-04-04
Feature: voyage ai embedding support

loc4atnt:
I've added Voyage AI embedding support and some guidance for the embedding config in the README file


tirth8205:
Voyage AI embedding support is not in main — embeddings.py only has local (sentence-transformers), Google Gemini, and MiniMax providers. Worth reviving if the implementation quality is good. However, note that this PR has 951 additions and 3492 deletions which is unusual — the deletions count may indicate it touched a lot of existing code. Recommend reviewing the diff carefully before reopening.

PR #106, opened 2026-04-05
fix: improve call graph accuracy -- filter method call noise, add Kotlin support

gzenz:
## Summary

- **Filter method call noise**: `obj.method()`, `response.json()`, `data.get()` were recorded as bare CALLS targets (`method`, `json`, `get`), polluting the graph with unresolvable noise and creating false callers via name collisions. Now only `self`/`cls`/`this`/`super` method calls emit CALLS edges.
- **Add Kotlin call extraction**: Kotlin uses `simple_identifier` and `navigation_expression` in its tree-sitter AST, which were not handled -- producing zero CALLS edges for Kotlin files.

### Impact (measured on this repo, Python-only)

| Metric | Before | After | Change |
|--------|--------|-------|--------|
| Total CALLS edges | 3006 | 1262 | -58% |
| Resolved (qualified) | 403 | 377 | -6% |
| Bare (unresolvable) | 2603 | 885 | -66% |

Almost all meaningful call relationships preserved (6% drop). The 58% total drop is noise removal.

## Test plan

- [x] `test_method_call_filtering_python_self` -- `self.method()` emits CALLS
- [x] `test_method_call_filtering_python_external` -- `obj.method()` does not
- [x] `test_method_call_filtering_python_super` -- `super().method()` emits CALLS
- [x] `test_method_call_filtering_ts_this` -- `this.method()` emits CALLS
- [x] `test_method_call_filtering_ts_external` -- `obj.method()` does not
- [x] Kotlin `test_finds_calls` -- `println()` works, `repo.save()` filtered
- [x] Full suite: 615 passed, 0 failed

### DIFF
--- a/code_review_graph/parser.py
+++ b/code_review_graph/parser.py
@@ -2432,7 +2432,8 @@ class CodeParser:
             return None  # method child not found
 
         # Simple call: func_name(args)
-        if first.type == "identifier":
+        # Kotlin uses "simple_identifier" instead of "identifier".
+        if first.type in ("identifier", "simple_identifier"):
             return first.text.decode("utf-8", errors="replace")
 
         # Perl: function_call_expression / ambiguous_function_call_expression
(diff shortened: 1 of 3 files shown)


tirth8205:
This PR addresses call graph noise (obj.method() creating false CALLS edges) and adds Kotlin call extraction. The changes are substantial (58% reduction in noisy CALLS edges) and all tests pass. Let me check if this has been integrated into main...

tirth8205:
The call graph noise filtering (only emit CALLS for self/cls/this method calls, filter out obj.method()) does not appear to be in main — the current parser.py extracts method names from all member_expression calls without filtering by receiver type. The 58% noise reduction and Kotlin call extraction this PR provides are still not in main. Worth reviving — this would significantly improve call graph accuracy and reduce false positives.

gzenz:
The noise filtering from this PR is fully included in #158 — same core logic (only allow self/cls/this/super receivers) plus additional refinements: uppercase receivers for class/static calls, namespace import receivers checked against the import map, nested member expression resolution (os.path.getsize), and a test-file exception to preserve TESTED_BY edges. Closing in favor of #158.

gzenz:
Noise filtering is included in #158 with additional improvements (see comment above). Closing.

PR #143, opened 2026-04-08
fix: bump fastmcp to patched versions

zerone0x:
## Summary
- bump mcp to >=2.0.0,<3
- bump fastmcp to >=2.14.0,<3 to avoid CVEs in <2.14.0

## Test plan
- not run (dependency-only change)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -25,8 +25,8 @@ classifiers = [
     "Topic :: Software Development :: Quality Assurance",
 ]
 dependencies = [
-    "mcp>=1.0.0,<2",
-    "fastmcp>=0.1.0,<2",
+    "mcp>=2.0.0,<3",
+    "fastmcp>=2.14.0,<3",
     "tree-sitter>=0.23.0,<1",
     "tree-sitter-language-pack>=0.3.0,<1",
     "networkx>=3.2,<4",


tirth8205 (commented):
Thanks for the CVE fix! The version bumps look correct. Two things before we can merge:

1. **Missing `uv.lock` update** — please run `uv lock` and include the updated lockfile.
2. **Major version jump** — `mcp` goes from `>=1.0.0` to `>=2.0.0` and `fastmcp` from `>=0.1.0` to `>=2.14.0`. These are major version bumps that could have breaking API changes. Have you verified the test suite passes with these new versions? The PR body says "not run" for test plan.

Could you confirm tests pass locally with the bumped deps?

tirth8205:
The version bumps are directionally correct — fastmcp >=2.14.0 addresses the CVEs and the mcp constraint aligns with fastmcp's own requirements. However, two blockers remain before this can merge:

1. **Missing `uv.lock` update**: The lockfile in the repo is not updated to reflect the new dependency constraints. Please run `uv lock` and commit the result.

2. **Test suite not verified**: The PR body says tests were not run, and the maintainer flagged this. These are major version jumps (`mcp` 1.x→2.x, `fastmcp` 0.1.x→2.14.x) that have breaking API changes. Please confirm `uv run pytest tests/ --tb=short -q` passes with the bumped dependencies. The `FastMCP` constructor and tool-registration APIs changed significantly between these versions.

Once those two items are addressed, this is a straightforward security fix and should be prioritized for merge.

tirth8205:
Thank you for this @zerone0x — closing as superseded by PR #222, which bumps \`fastmcp\` to \`>=2.14.0,<3\`. That PR covers the same CVE fix plus the transitive \`fakeredis\` breakage (#195), the Windows event-loop hang (#46 / #136), and several unrelated fixes. Verified locally: all 24 tools register on fastmcp 2.14.6, all 695 tests pass, coverage 73.6%.

Really appreciate you flagging this — you were first on the CVE and your PR pushed the conversation forward. #222 will ship in the next release (v2.2.4) once Windows testing lands.

PR #145, opened 2026-04-08
fix: preserve existing hooks on install

zerone0x:
## Summary
- back up existing .claude/settings.json before writing hooks
- merge new hooks with existing entries instead of overwriting them
- add tests for preserving hooks and backup creation

Fixes #114.

## Testing
- uv run pytest tests/test_skills.py -q


tirth8205:
CI passed, but this now has merge conflicts with `skills.py` after PR #152 merged. Could you rebase on latest `main`? Once rebased, we'll merge immediately.

tirth8205:
Thanks for tackling issue #114! The merge logic and backup approach are the right solution. However there are two bugs that need fixing before this can merge:

**1. Missing `encoding="utf-8"` on write (regression)**

The original `install_hooks` wrote with `encoding="utf-8"`:
```python
settings_path.write_text(json.dumps(existing, indent=2) + "\n", encoding="utf-8")
```

The PR drops the encoding parameter:
```python
settings_path.write_text(json.dumps(existing, indent=2) + "\n")
```

On Windows with non-UTF-8 system locale this will break. Please restore `encoding="utf-8"`.

**2. Hook format generated is invalid**

The PR still calls the original `generate_hooks_config()` which emits:
- A `PreCommit` hook — this is not a valid Claude Code hook event
- Flat `{"matcher": ..., "command": ..., "timeout": ...}` structure instead of the correct nested `{"matcher": ..., "hooks": [{"type": "command", "command": ..., "timeout": ...}]}` structure

These bugs were also present before this PR, but since you're changing the merge logic it would be good to fix the format here too. PR #192 has the correct format if you want to pull from it.

The backup creation and merge loop logic itself is correct — just these two things to fix.

PR #141, opened 2026-04-08
fix: generate valid Claude Code hook schema (closes #138)

sebastianbreguel:
Closes #138

## Problem

`generate_hooks_config()` emits a hooks config that Claude Code rejects with schema errors, so `code-review-graph install` corrupts `~/.claude/settings.json`. Three bugs in the same function:

1. **`PreCommit` is not a valid Claude Code hook event** → `Invalid key in record`.
2. **Missing inner `hooks` array wrapper.** Each entry must have `hooks: [{type: "command", command, timeout}]`, but the old output put `command`/`timeout` directly on the entry → `hooks: Expected array, but received undefined` (the exact error in the issue for `SessionStart`).
3. **Timeouts in milliseconds instead of seconds.** `5000`, `3000`, `10000` — the `PostToolUse` hook effectively had an 83-minute timeout.

Verified against the official schema: https://docs.claude.com/en/docs/claude-code/hooks

## Fix

Rewrite `generate_hooks_config()` to match the documented schema:

```json
"PostToolUse": [
  {
    "matcher": "Edit|Write|Bash",
    "hooks": [
      {"type": "command", "command": "code-review-graph update --quiet --skip-flows", "timeout": 5}
    ]
  }
]
```

- Remove the `PreCommit` entry (no equivalent Claude Code event; pre-commit behavior belongs in a git hook).
- Wrap `PostToolUse` and `SessionStart` handlers in `hooks: [{type: "command", ...}]`.
- Convert timeouts from ms to seconds (`5000` → `5`, `3000` → `3`).

## Testing

- [x] `test_has_post_tool_use` / `test_has_session_start` updated for the nested structure
- [x] New `test_does_not_emit_precommit` — regression guard for this issue
- [x] New `test_all_commands_wrapped_in_inner_hooks_array` — structural invariant
- [x] Full suite: 611 passed, 5 skipped, 0 failed
- [x] `ruff check` clean

PR #154, opened 2026-04-08
Fix Claude hook config and add JSX component CALLS edges

GrimoireScribe:
## Summary

This PR does two things:

1. Fixes Claude Code install/config issues:
- removes invalid `PreCommit` hook generation
- changes `PostToolUse` to `code-review-graph update --skip-flows`
- makes generated Claude/settings/skill file reads and writes explicit UTF-8 for Windows safety

2. Improves React/TSX graph coverage:
- emits synthetic `CALLS` edges for JSX component invocations
- supports plain component tags like `<MarkdownMsg />`
- supports namespace/member usage like `<UI.MarkdownMsg />`
- supports nested member chains like `<UI.Messages.MarkdownMsg />`
- verifies behavior through parser tests, `query_graph callers_of`, and impact-radius tests

## Why

React codebases often invoke components through JSX instead of normal function-call syntax. Before this PR, those usages produced no `CALLS` edges, which meant:
- `callers_of` could miss component call sites entirely
- impact radius under-reported React caller relationships
- flow/impact analysis was weak for component migration work

Separately, Claude Code hook generation was out of sync with valid hook events, and Windows installs could fail on Unicode content due to default-encoding file writes.

## Behavioral notes

Current JSX support is intentionally scoped:
- component-like tags are identified by uppercase names
- intrinsic DOM tags like `<div>` are ignored
- namespace imports can resolve JSX member expressions to module files
- unresolved member-style components still degrade gracefully

This improves React caller tracing a lot without trying to solve full React semantic analysis in one PR.

## Verification

Ran:

```bash
uv run --extra dev pytest tests\test_skills.py tests\test_parser.py tests\test_tools.py -q
uv run --extra dev ruff check tests\test_tools.py tests\test_parser.py code_review_graph\parser.py
```

### DIFF
--- a/code_review_graph/parser.py
+++ b/code_review_graph/parser.py
@@ -295,6 +295,7 @@ class CodeParser:
     def __init__(self) -> None:
         self._parsers: dict[str, object] = {}
         self._module_file_cache: dict[str, Optional[str]] = {}
+        self._export_symbol_cache: dict[str, Optional[str]] = {}
         self._tsconfig_resolver = TsconfigResolver()
 
     def _get_parser(self, language: str):  # type: ignore[arg-type]
@@ -971,6 +972,17 @@ class CodeParser:
                     import_map, defined_names, _depth,
                 ):
                     continue
+
+            # --- JSX component invocations ---
+            if (
+                language in ("javascript", "typescript", "tsx")
+                and node_type in ("jsx_opening_element", "jsx_self_closing_element")
+            ):
+                self._extract_jsx_component_call(
+                    child, language, file_path, edges,
+                    enclosing_class, enclosing_func,
+                    import_map, defined_names,
+                )
 
             # --- Solidity-specific constructs ---
             if language == "solidity" and self._extract_solidity_constructs(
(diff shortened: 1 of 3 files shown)


tirth8205 (commented):
This bundles two distinct features: (a) hook schema fixes and (b) JSX/TSX component CALLS edges.

The hook fixes overlap with PR #141 and #152 (both already merged/in progress). The JSX parsing piece is unique and valuable — good test coverage and +285 lines to parser.py.

Could you rebase on latest `main` and consider splitting the JSX CALLS feature into its own PR? The hook fixes are being handled separately. The JSX work stands on its own merit and we'd be happy to merge it.

GrimoireScribe:
Thanks — I rebased the branch on latest `main` and force-pushed a JSX-only version of the PR branch.

I removed the overlapping hook/config changes and kept only the JSX component `CALLS` work plus the parser tests/fixture, so the branch is now focused on the React/TSX caller-tracing feature.

GrimoireScribe:
Thanks — I rebased the branch on latest `main` and force-pushed a JSX-only version of the PR branch.

I removed the overlapping hook/config changes and kept only the JSX component `CALLS` work plus the parser tests/fixture, so the branch is now focused on the React/TSX caller-tracing feature.

I also verified it locally against a real React component (`MarkdownMsg`) after rebuilding the graph. The JSX caller tracing is working end to end now. Two follow-up behaviors showed up outside this PR's scope: `tests_for` needs a query-side fix for `TESTED_BY` direction/fallbacks, and bare-name resolution benefits from preferring the component/function node over matching test nodes. I fixed those separately on my internal branch and can send them as a separate PR if useful.

PR #162, opened 2026-04-08
fix: flush implicit transaction before BEGIN IMMEDIATE (#135)

azizur100389:
## Summary
- Fixes `sqlite3.OperationalError: cannot start a transaction within a transaction` when running `code-review-graph update` on a repo where `git diff` includes deleted files (#135).
- Adds a guard in `store_file_nodes_edges()` to commit any pending implicit transaction before `BEGIN IMMEDIATE`.
- Adds 3 regression tests covering both `incremental_update` and `full_build` code paths.

## Root Cause
Python's `sqlite3` with default `isolation_level=""` auto-begins a `DEFERRED` transaction on the first DML. In `incremental_update` and `full_build`, `remove_file_data()` issues `DELETE` statements for deleted/stale files, silently opening an implicit transaction. The subsequent `store_file_nodes_edges()` then executes `BEGIN IMMEDIATE` inside that already-open transaction, which SQLite rejects.

## Fix
Check `self._conn.in_transaction` before `BEGIN IMMEDIATE` and commit any pending implicit transaction. This preserves atomic replace semantics while being robust against prior implicit transaction state. The guard is a no-op when no implicit transaction is open.

## Tests Added
- `test_store_after_remove_no_transaction_error` — direct GraphStore unit test
- `test_incremental_deleted_and_modified_files` — end-to-end incremental update with mixed deletes and modifications
- `test_full_build_stale_files_then_parse` — end-to-end full build with stale file purging

## Test Plan
- [x] All 3 new regression tests pass
- [x] All existing `test_graph.py` and `test_incremental.py` tests pass (46/46)
- [x] `ruff check` clean on all changed files

### DIFF
--- a/code_review_graph/graph.py
+++ b/code_review_graph/graph.py
@@ -236,6 +236,13 @@ class GraphStore:
         self, file_path: str, nodes: list[NodeInfo], edges: list[EdgeInfo], fhash: str = ""
     ) -> None:
         """Atomically replace all data for a file."""
+        # Flush any implicit transaction started by prior DML (e.g.
+        # remove_file_data DELETEs during incremental_update/full_build).
+        # Python sqlite3 auto-begins a DEFERRED transaction on the first
+        # DML; BEGIN IMMEDIATE inside an active transaction raises
+        # OperationalError.  See GitHub issue #135.
+        if self._conn.in_transaction:
+            self._conn.commit()
         self._conn.execute("BEGIN IMMEDIATE")
         try:
             self.remove_file_data(file_path)
--- a/tests/test_graph.py
+++ b/tests/test_graph.py
@@ -242,3 +242,29 @@ class TestImpactRadiusSql:
         assert result["changed_nodes"] == []
         assert result["impacted_nodes"] == []
         assert result["total_impacted"] == 0
+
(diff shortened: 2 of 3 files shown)


tirth8205:
Closing in favor of #205 which was just merged and provides a more complete fix.

Your PR correctly identifies the problem and the `in_transaction` guard in `store_file_nodes_edges()` is the right instinct. The three regression tests (direct unit test, incremental update, full build) are thorough and well-structured.

However, PR #205 goes further:
1. **Root cause elimination**: sets `isolation_level=None` on all `sqlite3.connect()` calls — disabling implicit transactions project-wide so the guard becomes a true last-resort safety net rather than the primary defense.
2. **Full incremental path**: also commits after the deleted-file loop in `incremental_update()` (your `incremental.py` diff only patches `full_build()`).
3. **Other connect() sites**: your fix only touches `graph.py`, while #205 also patches `registry.py` and `embeddings.py`.
4. **Atomicity across batch ops**: wraps `store_flows()`, `store_communities()`, and `_compute_summaries` blocks in explicit transactions.

Your end-to-end regression tests were valuable — they were the most comprehensive integration coverage among all the competing PRs. Thank you for the contribution.

Issue #174, opened 2026-04-09
No warning before cloud embeddings send codebase to external APIs


dev-arctik:
## Problem

When Google Gemini or MiniMax embeddings are configured, the tool sends codebase content to external APIs with no prominent warning at the point of invocation. The risk is only mentioned in optional dependency docs — not surfaced at runtime.

## Impact

- Developers working on proprietary, client, or regulated codebases may unknowingly exfiltrate source code to third-party APIs
- There is no confirmation step before code leaves the machine
- This is especially dangerous when `embed_graph_tool` is called by an AI assistant autonomously without the developer realising an external API call is being made

## Example Scenario

A developer installs with `pip install code-review-graph[google-embeddings]`, runs `install`, and their AI assistant later calls `embed_graph_tool` during a review session — sending internal business logic to Google's API without any explicit action from the developer.

## Suggested Fix

1. At `serve` startup, print a clear warning if a cloud embedding provider is configured: `"Warning: Gemini embeddings are enabled — code will be sent to Google APIs"`
2. On first invocation of `embed_graph_tool` with a cloud provider, require explicit confirmation before proceeding
3. Add a `--local-only` flag that prevents any network egress entirely

tirth8205:
Valid security/privacy concern. When Gemini or MiniMax embeddings are configured, code content is sent to external APIs with no runtime warning. This is especially problematic in agentic workflows where embed_graph_tool may be called autonomously.

Suggested minimum fix: print a clear warning at serve startup if a cloud embedding provider is configured, e.g. 'WARNING: Cloud embeddings enabled (provider=gemini) — code content will be sent to Google APIs.' A --local-only flag would also be valuable.

This is a valid open issue.

PR #179, opened 2026-04-09
Add user confirmation before sending code to external embedding APIs isuue #174

Bakul2006:
-Added a confirmation prompt before invoking external embedding APIs
- Ensures that code is not sent outside the local environment without explicit user consent

  Behavior
- If user confirms → API call proceeds
- If user declines → operation is aborted
- In non-interactive environments (e.g., AI agents, CI) → call is blocked by default

  Impact
- Prevents accidental code exfiltration
- Improves transparency and user control
- Keeps behavior aligned with expectations of local-first tooling


Bakul2006:
This change adds a minimal confirmation step before sending code to external embedding APIs. Happy to adjust the behavior based on feedback.

tirth8205:
## Code Review — Concept valid, implementation has a critical bug

The goal of preventing accidental code exfiltration to cloud embedding APIs is correct. However the implementation has a critical bug and should not be merged as-is.

### Critical Bug: `input()` corrupts the MCP stdio protocol

The confirmation prompt in `code_review_graph/embeddings.py` (or wherever you placed it — the diff shows it in `tools/docs.py`) calls `input()`, which writes to **stdout**. The MCP server uses stdio transport, where **stdout is the JSON-RPC channel**. Writing a user-facing prompt to stdout corrupts the protocol and breaks any MCP client connected to the server.

Example: Claude Code calling `embed_graph` via MCP would receive a corrupted response like:
```
WARNING: This will send codebase content to google APIs. Continue? (yes/no): {"result": ...}
```
The JSON parser would fail on the prefix.

### Correct approach

For non-interactive environments (MCP server, CI, AI agents), the confirmation should **refuse by default** without any stdout output. In interactive TTY environments, the prompt must go to **stderr** (not stdout):

```python
def _confirm_cloud_usage(provider_name: str) -> bool:
    """Confirm cloud provider usage. Safe for both MCP and CLI use."""
    if not sys.stdin.isatty():
        return False  # Non-interactive: always refuse
    # Write to stderr so MCP's stdout channel is not corrupted
    sys.stderr.write(
        f"\nWARNING: embed_graph will send codebase content to {provider_name} APIs.\n"
        "Continue? (yes/no): "
    )
    sys.stderr.flush()
    try:
        return sys.stdin.readline().strip().lower() in ("yes", "y")
    except (EOFError, KeyboardInterrupt):
        return False
```

### Other issues

1. **Placement**: The confirmation logic belongs in `tools/docs.py`'s `embed_graph()` function (the MCP tool layer), not inside `EmbeddingStore` itself. `EmbeddingStore` is a library class used in tests and scripts — injecting a confirmation prompt there breaks all automated use.

2. **`EmbeddingStore.provider.name.startswith("local:")`**: The check for whether a provider is local uses a string prefix. This is fragile. Better to check the provider type: `isinstance(emb_store.provider, LocalProvider)` or add an `is_local` property.

3. **CI/testing**: Any test that creates a non-local `EmbeddingStore` will now hang waiting for user input unless `EOFError` is raised. Your `try/except EOFError` handles this, but only if `input()` raises it — which it does in non-TTY test environments. The `sys.stdin.isatty()` check is the reliable way to detect this.

### What to keep from this PR

The concept: before invoking a cloud embedding provider, check and block in non-interactive environments, warn in interactive ones. The `EmbeddingStore` wrapping approach in `embed_graph()` is the right place. The `EOFError`/`KeyboardInterrupt` handling is correct.

Please fix the `stdout` corruption issue and resubmit. The change is small (move to `sys.stdin.isatty()` + `sys.stderr.write()`) but critical.

Bakul2006:
Addressed all review feedback and resolved conflicts.
Ready for review.

tirth8205:
Skipping auto-merge: after updating against current `main`, CI shows the following problems:

**ruff (13 errors)** — mostly `W293` (blank line whitespace) and `F841` (unused variables); the diff wasn't run through `ruff check code_review_graph/` before push.

**mypy (1 error):** `code_review_graph/main.py:739: "FastMCP" has no attribute "_local_only"` — accessing a private attribute that doesn't exist.

All 4 test jobs (3.10–3.13) pass, so the feature itself looks fine; just needs a lint/type-check pass. Please run `ruff check --fix code_review_graph/` and `mypy code_review_graph/ --ignore-missing-imports --no-strict-optional` and push a new commit, and I'll re-review.

tirth8205:

Issue #172, opened 2026-04-09
Why v2.2.2 still has the "Hooks use a matcher + hooks array"  error


Onetaway:
I have update to the v2.2.2 

after run ``code-review-graph install`` command，the generate ``/.claude/settings.json`` is still

  "hooks": {
    "PostToolUse": [
      {
        "matcher": "Edit|Write|Bash",
        "command": "code-review-graph update --skip-flows",
        "timeout": 5000
      }
    ],
    "SessionStart": [
      {
        "command": "code-review-graph status",
        "timeout": 3000
      }
    ],
    "PreCommit": [
      {
        "command": "code-review-graph detect-changes --brief",
        "timeout": 10000
      }
    ]
  }

after that, run ``claude``, still has this error

hooks
     ├ PostToolUse
     │ └ 0
     │   └ hooks: Expected array, but received undefined
     ├ PreCommit: Invalid key in record
     └ SessionStart
       └ 0
         └ hooks: Expected array, but received undefined

akhilv77:
I am getting the same issue. when i do claude command.

summy00:
Claude帮我修复了:
原来 code-review-graph install 生成的 hooks 格式不对，Claude Code 要求每个事件下的条目包含 matcher 和 hooks 数组（每个 hook 有
  type、command、timeout），而不是直接把 command 放在顶层。

  修改内容：
  - PostToolUse / SessionStart 的 hook 结构改为 { matcher, hooks: [{ type: "command", command, timeout }] } 格式
  - 移除了 PreCommit（不是 Claude Code 支持的 hook 事件）

```
 Added 1 line, removed 1 line
       6          "hooks": [
       7            {
       8              "type": "command",
       9 -            "command": "code-review-graph update --quiet --skip-flows",
       9 +            "command": "code-review-graph update --skip-flows",
      10              "timeout": 5000
      11            }
      12          ]
  ⎿  PostToolUse:Edit hook returned blocking error  
```

tirth8205:
Confirmed: v2.2.2 still has this bug. The skills.py:generate_hooks_config() function has not been updated to use the correct Claude Code hook schema. This is the same issue as #97, #163, #188, #191, #201.

PR #203 provides a complete fix but was closed without merging. The core problem in the current code is:
1. Missing nested hooks: [{type: command, command: ...}] array in each hook entry
2. PreCommit is not a valid Claude Code hook event
3. SessionStart is not a valid Claude Code hook event

[...]

Issue #163, opened 2026-04-08
Settings Error: hooks schema validation failures in .claude/settings.json


anouar-bm:
## Bug Report

When using the code-review-graph MCP server with Claude Code, the `.claude/settings.json` file generated or expected by the plugin fails schema validation with the following errors:

```
Settings Error

  /Users/.../1337/.claude/settings.json
   └ hooks
     ├ PostToolUse
     │ └ 0
     │   └ hooks: Expected array, but received undefined
     ├ PreCommit: Invalid key in record
     └ SessionStart
       └ 0
         └ hooks: Expected array, but received undefined
```

## Expected Behavior

Hooks should use a `matcher` + `hooks` array structure. The correct format is:

```json
{
  "PostToolUse": [
    {
      "matcher": "Edit|Write",
      "hooks": [{"type": "command", "command": "echo Done"}]
    }
  ]
}
```

## Actual Behavior

The `hooks` field inside hook entries is `undefined` (missing), and `PreCommit` is not a valid key in the hooks record.

## Impact

> Files with errors are skipped entirely, not just the invalid settings.

This means all hooks configured in `.claude/settings.json` are silently ignored when the file contains these errors, breaking any SessionStart or PostToolUse hooks (e.g., auto graph updates on file changes).

## Steps to Reproduce

1. Install the code-review-graph MCP server
2. Follow setup instructions that configure `.claude/settings.json` hooks
3. Open Claude Code — the settings error is shown on startup

## Suggested Fix

Update the hooks configuration in the setup instructions / generated config to use the correct schema:

```json
{
  "hooks": {
    "PostToolUse": [
      {
        "matcher": "Edit|Write",
        "hooks": [{"type": "command", "command": "<graph update command>"}]
      }
    ],
    "SessionStart": [
      {
        "matcher": "",
        "hooks": [{"type": "command", "command": "<startup command>"}]
      }
    ]
  }
}
```

## Reference

- [Claude Code Hooks Documentation](https://code.claude.com/docs/en/hooks)

tirth8205:
Same hooks schema bug as #97, #172, #188, #191, #201. The generated settings.json still uses the old invalid format. See #201 for full details and the correct schema. PR #203 has a complete fix waiting to be merged.

tirth8205:
Fixed in **v2.2.3**, now live on PyPI: https://pypi.org/project/code-review-graph/2.2.3/

[...]

PR #180, opened 2026-04-09
fix: emit valid Claude Code hook schema in settings.json

pashak1207:
Claude Code rejects the file generated by `code-review-graph install` with "Expected array, but received undefined" because each entry under PostToolUse/SessionStart needs an inner `hooks: [{type, command}]` array, not a flat `command:` field. On top of that, the generator emits a `PreCommit` block — which isn't a Claude Code event at all.

Consequence: the entire settings.json is discarded, so neither the PostToolUse auto-update hook nor the SessionStart status hook ever fires.

This rewrites generate_hooks_config() to produce the correct schema, drops the bogus PreCommit block, and narrows the PostToolUse matcher from `Edit|Write|Bash` to `Edit|Write`. Bash calls in Claude Code don't modify source files directly, so running an incremental graph update after every shell command was just noise.

Tests in TestGenerateHooksConfig were updated to assert the new schema, plus a new regression test (test_entries_use_claude_code_hook_schema) guards against the flat-command shape recurring.

Fixes #97, #163, #172.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## [Unreleased]
+
+### Fixed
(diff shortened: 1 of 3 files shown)


ZzAve:
FWIW: I applied these fixed to my local setup and can confirm `code-review-graph` to be working with claude.

tirth8205:
Closing in favor of #208 which was just merged and provides a more complete fix.

Your PR correctly fixes the two critical bugs: nested `hooks` array wrapper and removal of the invalid `PreCommit` event. The narrowing of the PostToolUse matcher from `Edit|Write|Bash` to `Edit|Write` is a reasonable call too.

However, there is one remaining issue in this PR: the timeouts (`5000`, `3000`) are still in milliseconds. According to the Claude Code docs, timeout values are in **seconds**. Leaving them at 5000 would mean an 83-minute timeout on the PostToolUse hook. PR #208 correctly converts these to 30s and 10s.

PR #208 also adds a new `install_git_hook()` function that restores the pre-commit intent (which was lost when removing the invalid `PreCommit` hook event) by writing a proper `.git/hooks/pre-commit` script instead — with tests for create, append, idempotent, and no-.git-dir cases.

Thank you for the contribution — identifying the timeout unit bug was a key insight.

PR #181, opened 2026-04-09
fix: rebuild crashes with "cannot start a transaction within a transaction"

realkotob:
## The problem

Running `code-review-graph build` on a large repo (~9k files) after an earlier interrupted build crashes almost immediately with:

```
```

I hit this on a big source tree on my Macbook M1 Pro. The first build ran through the parse phase to completion, then I Ctrl-C'd it while community detection was stuck (separate perf issue — the file-based fallback is quadratic on large graphs). On the retry, the rebuild blew up on the very first file:

```
INFO: Progress: ... files parsed ... (never gets here)
  ...
  File ".../incremental.py", line 397, in full_build
      store.store_file_nodes_edges(...)
  File ".../graph.py", line 239, in store_file_nodes_edges
      self._conn.execute("BEGIN IMMEDIATE")
```

## Why it happens

`full_build` starts by purging any files that used to exist in the DB but no longer exist on disk:

```python
for stale in existing_files - current_abs:
    store.remove_file_data(stale)
```

`remove_file_data` issues two `DELETE` statements and returns without committing. Under Python's default (legacy) `sqlite3` isolation mode, those `DELETE`s silently open an implicit transaction. The very next `store_file_nodes_edges()` call then runs an explicit `BEGIN IMMEDIATE` — and SQLite rejects it because one is already open.

The reason this went undetected is that the purge loop is usually a no-op: if every file still exists on disk, `existing_files - current_abs` is empty and the buggy path never runs. My interrupted build had left the DB with thousands of file rows whose absolute paths no longer matched the set the rebuild was computing, so every single one was treated as stale and the connection was left dirty before parsing even started.

## The fix

Two parts:

1. **`incremental.py`** — `full_build` now calls `store.commit()` once after draining the purge loop, so the parse phase starts with a clean connection.
2. **`graph.py`** — `store_file_nodes_edges` checks `conn.in_transaction` and flushes any dangling implicit transaction before its `BEGIN IMMEDIATE`. Defensive: if any future code path leaves DML uncommitted, the atomic per-file replace still works instead of crashing the build.

## Test plan

- [x] New regression test `test_store_file_nodes_edges_after_dangling_dml` primes an implicit transaction via `remove_file_data` and asserts `store_file_nodes_edges` still succeeds. Fails on `main`, passes with the fix.
- [x] `uv run pytest tests/test_graph.py tests/test_incremental.py` — 45 passed
- [x] `uv run pytest tests/test_tools.py tests/test_integration_v2.py` — 66 passed
- [x] Manually reproduced the crash on a real repo and confirmed the fix allows the rebuild to proceed past the purge loop.

### DIFF
--- a/code_review_graph/graph.py
+++ b/code_review_graph/graph.py
@@ -236,6 +236,11 @@ class GraphStore:
         self, file_path: str, nodes: list[NodeInfo], edges: list[EdgeInfo], fhash: str = ""
     ) -> None:
         """Atomically replace all data for a file."""
+        # Flush any pending implicit transaction opened by earlier DML on this
+        # connection so the explicit BEGIN IMMEDIATE below doesn't trip SQLite's
+        # "cannot start a transaction within a transaction" error.
+        if self._conn.in_transaction:
+            self._conn.commit()
         self._conn.execute("BEGIN IMMEDIATE")
         try:
             self.remove_file_data(file_path)
--- a/code_review_graph/incremental.py
+++ b/code_review_graph/incremental.py
@@ -352,8 +352,13 @@ def full_build(repo_root: Path, store: GraphStore) -> dict:
     # Purge stale data from files no longer on disk
     existing_files = set(store.get_all_files())
     current_abs = {str(repo_root / f) for f in files}
-    for stale in existing_files - current_abs:
-        store.remove_file_data(stale)
+    stale_files = existing_files - current_abs
+    if stale_files:
+        for stale in stale_files:
+            store.remove_file_data(stale)
+        # Commit the implicit transaction opened by the DELETEs so the
+        # subsequent BEGIN IMMEDIATE in store_file_nodes_edges succeeds.
+        store.commit()
 
     total_nodes = 0
     total_edges = 0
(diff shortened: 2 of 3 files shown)


tirth8205:
This was closed in favor of #205 which was just merged and provides a more complete fix.

Your PR correctly identifies the two-part fix: commit after the stale-file purge loop in `full_build()`, and an `in_transaction` guard in `store_file_nodes_edges()`. Both of those changes are also present in #205.

However, PR #205 goes further:
1. **Root cause elimination**: sets `isolation_level=None` on all `sqlite3.connect()` calls in `graph.py`, `registry.py`, and `embeddings.py` — so implicit transactions can never leak again.
2. **Incremental path too**: adds the corresponding `store.commit()` after the deleted-file loop in `incremental_update()`.
3. **Atomicity**: wraps `store_flows()`, `store_communities()`, and all three `_compute_summaries` blocks in explicit transactions.
4. **More regression tests**: 2 new tests covering both single and bulk remove-then-store scenarios.

Your root-cause analysis was excellent and helped clarify the problem for everyone. Thank you for the contribution.

Issue #201, opened 2026-04-10
Bug: Claude Code fails to start — invalid hooks schema in .claude/settings.json


pavitra93:
**Affects: code-review-graph install --platform claude-code on Claude Code v1.x+**

**Description**
After running code-review-graph install --platform claude-code, the generated .claude/settings.json uses an outdated hook schema that is incompatible with the current Claude Code hooks format. As a result, Claude Code throws a Settings Error on startup and skips the entire settings file.

**Steps to reproduce**
1.  brew install pipx
2.  pipx install code-review-graph
3.  code-review-graph install --platform claude-code
4.  code-review-graph build
5.  Run claude in any project terminal
Error
Settings Error — /path/to/project/.claude/settings.json
 └ hooks
   ├ PostToolUse › 0 › hooks: Expected array, but received undefined
   ├ PreCommit: Invalid key in record
   └ SessionStart › 0 › hooks: Expected array, but received undefined

unix4lyf:
Getting the same issue

Legys:
Yeah the same

Razbotics:
I asked claude to fix it and it modified it to this
```
  "hooks": {
    "PostToolUse": [
      {
        "matcher": "Edit|Write|Bash",
        "hooks": [
          {
            "type": "command",
            "command": "flock -n /tmp/code-review-graph.lock code-review-graph update --skip-flows",
            "timeout": 5000
          }
        ]
      }
    ],
    "SessionStart": [
      {
        "hooks": [
          {
            "type": "command",
            "command": "code-review-graph status",
            "timeout": 3000
          }
        ]
      }
    ],
    "Stop": [
      {
        "hooks": [
          {
            "type": "command",
            "command": "code-review-graph detect-changes --brief",
            "timeout": 10000
          }
        ]
      }
    ]
  }
}
```

pavitra93:
Yeah, that can be solved but this should be fixed from code itself.

tirth8205:
This is still a valid open bug. The current code in `skills.py:generate_hooks_config()` still generates the old invalid schema:

```json
{
  "PostToolUse": [{"matcher": "Edit|Write|Bash", "command": "...", "timeout": 5000}],
  "PreCommit": [{...}],
  "SessionStart": [{...}]
}
```

The correct Claude Code schema requires a nested `hooks` array:

```json
{
  "PostToolUse": [{
    "matcher": "Edit|Write",
[...]

Issue #191, opened 2026-04-09
claude can't start after `uvx code-review-graph build`


Legys:
I wanted to try the tool, installed via the instruction in the readme, but when I run `claude` in the repository I installed it, I see some hooks with errors. I haven't found existing issues and my setup doesn't have existing hooks.

`.claude/settings.json` Looks like this:

```
{
  "hooks": {
    "PostToolUse": [
      {
        "matcher": "Edit|Write|Bash",
        "command": "code-review-graph update --skip-flows",
        "timeout": 5000
      }
    ],
    "SessionStart": [
      {
        "command": "code-review-graph status",
        "timeout": 3000
      }
    ],
    "PreCommit": [
      {
        "command": "code-review-graph detect-changes --brief",
        "timeout": 10000
      }
    ]
  }
}

```
This error appears when I run claude:
```
 └ hooks
     ├ PostToolUse
     │ └ 0
     │   └ hooks: Expected array, but received undefined
     ├ PreCommit: Invalid key in record
     └ SessionStart
       └ 0
         └ hooks: Expected array, but received undefined

  Hooks use a matcher + hooks array. The matcher is a string: a tool name ("Bash"), pipe-separated list ("Edit|Write"), or empty to match all. Example: {"PostToolUse": [{"matcher": "Edit|Write", "hooks": [{"type": "command",
  "command": "echo Done"}]}]}
  Learn more: https://code.claude.com/docs/en/hooks


  Files with errors are skipped entirely, not just the invalid settings.
```

tirth8205:
This is the same hooks schema bug reported in #97, #163, #172, #188, #201. The current version of skills.py still generates the old invalid hook format missing the nested hooks array. PreCommit and SessionStart are also not valid Claude Code hook event types.

The fix is to update generate_hooks_config() in skills.py to use the correct schema:
- Each hook entry needs a hooks: [{type: command, command: ...}] nested array
- Remove PreCommit (not a valid Claude Code hook event)
- Remove SessionStart (not a valid Claude Code hook event)

**Workaround:** After installing, manually edit .claude/settings.json to use the correct format. See the Claude Code hooks docs at https://docs.anthropic.com/en/docs/claude-code/hooks

tirth8205:
Fixed in **v2.2.3**, now live on PyPI: https://pypi.org/project/code-review-graph/2.2.3/

[...]

Issue #188, opened 2026-04-09
claude settings.json in wrong format


laoshaw:
after the install, claude code complains about invalid settings.json.

<img width="774" height="750" alt="Image" src="https://github.com/user-attachments/assets/777f7ba0-c33b-4791-ae8d-448fa5ab165c" />

svikas641:
does it work after applying the suggested changes ?

tirth8205:
Same issue as #97, #163, #172, #191, #201. The install command generates settings.json with an outdated hook schema that Claude Code rejects. The fix is pending — see #201 for details and the correct schema format.

**Quick workaround:** Edit .claude/settings.json to use:
```json
{"hooks": {"PostToolUse": [{"matcher": "Edit|Write", "hooks": [{"type": "command", "command": "code-review-graph update --skip-flows", "timeout": 5000}]}]}}```

For igraph: install it with pip install igraph or uv add igraph. The WARNING is shown once when the build starts; after installing igraph, restart the build and it should be picked up.

tirth8205:
Fixed in **v2.2.3**, now live on PyPI: https://pypi.org/project/code-review-graph/2.2.3/

**Root cause**: in v2.2.2, \`generate_hooks_config()\` emitted an invalid Claude Code hook schema — flat \`{matcher, command, timeout}\` entries without the required nested \`hooks: [{type, command, timeout}]\` array, timeouts in milliseconds instead of seconds, and a \`PreCommit\` event that isn't a real Claude Code event. PR #208 rewrote the generator to emit the correct v1.x+ schema and moved pre-commit checks into an actual git pre-commit hook via \`install_git_hook()\`.

**Upgrade and re-run install:**

\`\`\`bash
uvx --reinstall code-review-graph
# or
pip install -U code-review-graph

# Then rewrite .claude/settings.json:
code-review-graph install
\`\`\`

Closing. If you still see the error after upgrading and re-running \`install\`, please reopen with the contents of your \`.claude/settings.json\` — the old broken entries may have been merged in rather than overwritten.

PR #203, opened 2026-04-10
fix: merge-based hook installation preserves existing user hooks

lngyeen:
## Summary

`install_hooks()` now **merges** new entries into existing hooks instead of overwriting them via `existing.update()`. Creates a `settings.json.bak` backup before modification.

## Motivation

The `existing.update(hooks_config)` call in `install_hooks()` replaces the entire `hooks` key, silently destroying user-defined hooks (e.g., linters, formatters, safety checks).

> **Note**: The hook schema fix (invalid nested arrays, unsupported `PreCommit` event, `Bash` matcher) was already resolved upstream in PR #208. This PR's scope is narrowed to the **merge logic**, **backup**, and **regression test** only.

Closes #114

## Changes

| File | Change |
|------|--------|
| `skills.py:install_hooks()` | Merge-based hook installation with backup |
| `tests/test_skills.py` | Schema regression test + merge behavior test |

## Test plan

- [x] All skills tests pass
- [x] Full test suite passes with zero regressions
- [x] Verified hooks append correctly alongside existing user hooks


tirth8205:
This PR addresses the hooks schema bug (#97, #114, #172, #188, #191, #201) which is still not fixed in main. The current skills.py:generate_hooks_config() still generates the invalid schema. This PR provides exactly what's needed: proper nested hooks arrays, removal of PreCommit/SessionStart (invalid events), narrowed PostToolUse matcher, and merge-based installation with backup. Worth reviving — this is the #1 most-reported bug in the issue tracker.

lngyeen:
Rebased onto latest `main` — conflicts resolved, all 46 skill tests passing. The hook schema changes from #208 are now upstream, so this PR's scope is narrowed to the **merge logic** (append-don't-overwrite user hooks) + **backup** (`settings.json.bak`) + a detailed schema regression test. Ready for re-review.

tirth8205:
Thanks for the thorough fix! #257 was merged which addresses the same clobbering issue. Your approach with backup + array merging is more robust — if you'd like to rebase on main and contribute the improvements on top of #257, that would be welcome.

PR #142, opened 2026-04-08
fix: --platform target now filters skills, hooks, and instruction files (#133)

sebastianbreguel:
## Problem

Running \`code-review-graph install --platform cursor\` (or any non-\"all\" target) still wrote Claude-specific files and every platform rule file, ignoring the flag:

- \`CLAUDE.md\` was injected
- \`.claude/settings.json\` hooks were installed
- \`AGENTS.md\`, \`GEMINI.md\`, \`.cursorrules\`, \`.windsurfrules\` were all written

This made \`--platform\` effectively cosmetic — the only thing it actually scoped was the MCP server config.

Fixes #133.

## Fix

- \`skills.py\`: \`inject_platform_instructions\` now takes a \`target\` param and filters files via an ownership map:
  - \`AGENTS.md\` → \`cursor\`, \`opencode\`, \`antigravity\`
  - \`GEMINI.md\` → \`antigravity\`
  - \`.cursorrules\` → \`cursor\`
  - \`.windsurfrules\` → \`windsurf\`
  Default \`target=\"all\"\` preserves existing behavior. \`target=\"claude\"\` writes nothing (CLAUDE.md is handled separately).
- \`cli.py\` \`_handle_init\`: \`inject_claude_md\` and \`install_hooks\` are now gated on \`target in (\"claude\", \"all\")\`, and \`target\` is forwarded to \`inject_platform_instructions\`.

Note: the fix does not remove stale files from prior runs — e.g. if a user ran with \`--platform all\` and later re-runs with \`--platform cursor\`, existing \`GEMINI.md\` / \`.windsurfrules\` stay on disk. Deleting user files silently felt unsafe; happy to revisit if you'd prefer a cleanup step.

## Testing

- New \`TestInjectPlatformInstructionsFiltering\` covers \`all\`, default, \`claude\`, \`cursor\`, \`windsurf\`, \`antigravity\`, \`opencode\`.
- \`uv run pytest\` — 618 passed
- \`uv run ruff check\` — clean on touched files

PR #204, opened 2026-04-10
feat: preview and confirm before injecting instructions into CLAUDE.md

### LINKED ISSUE #173: install silently rewrites CLAUDE.md / platform rules without preview
## Problem

The `code-review-graph install` command injects instructions into platform rules (e.g. `CLAUDE.md` for Claude Code) without showing the user what it's writing before it happens. There is no `--dry-run` flag, no diff preview, and no confirmation prompt.

## Impact

- Users discover unexpected content in their `CLAUDE.md` after the fact with no audit trail of what was added
- In team repos, `CLAUDE.md` is checked into git — a silent rewrite shows up as an unexplained diff in `git status`
- The injected instructions affect AI assistant behavior globally, which is a meaningful change that deserves explicit user consent

## Suggested Fix

1. Add a `--dry-run` flag that prints the exact changes that would be made without applying them
2. Before writing, show a diff of what will be added and prompt for confirmation (can be skipped with `--yes`)
3. At minimum, print the injected content to stdout after install so users know what was added

lngyeen:
## Summary

Addresses #173 — `install` silently rewrites CLAUDE.md / platform rules without preview.

Three changes:

1. **`--dry-run` now covers instruction injection**: shows the full content that would be appended to CLAUDE.md and platform rule files, then exits without writing.

2. **Interactive confirmation**: before injecting, the installer prints the full content and list of target files, then prompts `Inject into CLAUDE.md, AGENTS.md, ...? [Y/n]`. Non-TTY environments (CI, piped input) auto-confirm for backwards compatibility.

3. **New flags**:
   - `--yes` / `-y`: auto-confirm without prompting
   - `--no-instructions`: skip injection entirely

## Example output

```
Graph instructions will be appended to: CLAUDE.md, AGENTS.md (new), .cursorrules (new)
Content to inject:
  <!-- code-review-graph MCP tools -->
  ## MCP Tools: code-review-graph
  ...

Inject into CLAUDE.md, AGENTS.md (new), .cursorrules (new)? [Y/n]
```

## Test plan

- [x] All 615 tests pass
- [x] `--dry-run` shows preview without writing
- [x] `--no-instructions` skips injection
- [x] `--yes` auto-confirms
- [x] Non-TTY defaults to confirm (backwards compat)
- [x] Idempotent: already-injected files are skipped

Closes #173

### DIFF
--- a/code_review_graph/cli.py
+++ b/code_review_graph/cli.py
@@ -94,6 +94,63 @@ def _print_banner() -> None:
 """)
 
 
+def _preview_instructions(
+    repo_root: Path,
+    section: str,
+    marker: str,
+    platform_files: dict[str, str],
+) -> list[str]:
+    """Print a preview of files that would be modified by instruction injection.
+
+    Returns list of file names that would be modified (not already containing
+    the marker).
+    """
+    targets: list[str] = []
+
+    claude_md = repo_root / "CLAUDE.md"
+    if claude_md.exists():
+        content = claude_md.read_text(encoding="utf-8")
+        if marker not in content:
+            targets.append("CLAUDE.md")
+    else:
+        targets.append("CLAUDE.md (new)")
+
+    for label, filename in platform_files.items():
+        path = repo_root / filename
+        if path.exists():
+            content = path.read_text(encoding="utf-8")
+            if marker not in content:
+                targets.append(label)
+        else:
+            targets.append(f"{label} (new)")
+
+    if targets:
+        print(f"\nGraph instructions will be appended to: {', '.join(targets)}")
+        print("Content to inject:")
+        for line in section.strip().splitlines():
+            print(f"  {line}")
+        print()
+    return targets
+
+
+def _confirm_inject(targets: list[str]) -> bool:
+    """Prompt user to confirm instruction injection.
+
+    Returns True if user confirms, False otherwise.  Non-interactive
+    environments (no TTY) default to True to preserve backwards
+    compatibility with CI/automation.
+    """
+    import sys
+    if not sys.stdin.isatty():
+        return True
+    try:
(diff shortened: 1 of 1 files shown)


tirth8205:
The preview-and-confirm UX is exactly what issue #173 asked for — good design choices: dry-run shows preview, --no-instructions opt-out, --yes for CI, non-TTY auto-confirms. The implementation is clean.

Two things to address before merging:

**1. Logic bug: inject_claude_md not counted in 'injected' list**

In this block:
```python
inject_claude_md(repo_root)
updated = inject_platform_instructions(repo_root)
injected = ["CLAUDE.md"] + updated if (repo_root / "CLAUDE.md").exists() else updated
```

The conditional check is wrong due to Python operator precedence — this evaluates as:
  injected = ["CLAUDE.md"] + (updated if condition else updated)
which always includes "CLAUDE.md". It should be:
  injected = (["CLAUDE.md"] + updated) if already_confirmed_we_wrote_it else updated

More importantly, inject_claude_md also creates CLAUDE.md if it doesn't exist, so checking .exists() post-write isn't the right guard anyway. Simplest fix: just always include "CLAUDE.md" in the list since inject_claude_md was unconditionally called.

**2. Type incompatibility with PR #142**

_preview_instructions expects platform_files as dict[str, str] (label -> filename), but if PR #142 merges first, _PLATFORM_INSTRUCTION_FILES becomes dict[str, tuple[str, ...]] (filename -> owners tuple). The for loop 'for label, filename in platform_files.items()' will then unpack incorrectly. Please check for ordering dependency or make _preview_instructions robust to the updated type.

Everything else looks good — the --no-instructions skip path, the --yes flag, the dry-run preview, and the non-TTY backwards-compat default are all correct.

tirth8205:

PR #156, opened 2026-04-08
added configurable data directory support using CRG_DATA_DIR environment variable #155

Bakul2006:
The .code-review-graph directory is currently hardcoded at the repository root, with no way to customize its location. This causes issues with project structure, CI/CD workflows, and monorepo setups.

Introduced support for a configurable data directory using the CRG_DATA_DIR environment variable
Centralized path resolution via a helper function (get_data_dir)
Updated all relevant usages:
Database (graph.db)
Wiki directory
HTML graph output
Registry path addressing the issue #155

PR #207, opened 2026-04-10
feat: add CRG_DATA_DIR and CRG_REPO_ROOT env vars for zero-artifact

yashmewada9618:
Introduces two environment variables that together allow storing all generated files outside the repository, keeping the working tree completely clean (no .code-review-graph/ directory, no .gitignore entries, nothing).

CRG_DATA_DIR — relocate the data directory
  - New get_data_dir(repo_root) helper returns .code-review-graph/ by default, or the path from CRG_DATA_DIR when set (absolute or relative to repo_root).
  - get_db_path(), cli.py visualize/wiki commands, and tools/docs.py wiki tools all use get_data_dir(), so every generated artifact (graph.db, graph.html, wiki/) respects the override.
  - The auto-generated .gitignore is only written when using the default in-repo location; an external directory is left untouched.

CRG_REPO_ROOT — anchor the MCP server to a specific repository
  - find_project_root() checks CRG_REPO_ROOT before walking up via .git, so the MCP server process can be started from any working directory (e.g., a user-level IDE launcher) and still resolve the correct repo.
  - Without this, code-review-graph serves anchors to Path.cwd(), which is often / or the home directory when launched by an MCP host.

Typical zero-artifact setup (e.g., in ~/.claude.json mcpServers):

  CRG_DATA_DIR=~/.local/share/code-review-graph/my-project
  CRG_REPO_ROOT=/path/to/my-project

Adds 10 tests covering both env vars (default behaviour, absolute path, relative path, no. gitignore-in-external-dir, fallback cases).

Co-Authored By: Claude 
Relates to #155 and #156


tirth8205:
## Code Review — Mostly good, two blocking issues

This PR is the preferred implementation over #156 for `CRG_DATA_DIR`. The `get_data_dir()` + `CRG_REPO_ROOT` core is correct and well-tested. However there are two blocking issues in the bundled embedding confirmation feature:

### Blocking Issue 1: `input()` inside an MCP tool corrupts the stdio protocol

`_confirm_cloud_usage` in `tools/docs.py` calls `input()`, which writes to stdout. The MCP server uses **stdio transport** — stdout is the JSON-RPC channel. Writing a confirmation prompt to stdout will corrupt the protocol stream and break the MCP client.

```python
def _confirm_cloud_usage(provider_name: str) -> bool:
    try:
        response = input(  # ← BROKEN: writes to MCP's stdout channel
            f"WARNING: This will send codebase content to {provider_name} APIs. "
            "Continue? (yes/no): "
        ).strip().lower()
```

In non-interactive mode the `EOFError` path returns `False` (blocking the call), which is the right default. But the `input()` call itself is what corrupts the stream. Fix: use `sys.stderr` for warnings and check `sys.stdin.isatty()` before calling `input()`. In non-TTY environments, refuse by default. In TTY environments, write the prompt to `sys.stderr` and read from `sys.stdin` directly:

```python
def _confirm_cloud_usage(provider_name: str) -> bool:
    if not sys.stdin.isatty():
        return False  # Non-interactive: refuse by default
    sys.stderr.write(
        f"WARNING: This will send codebase content to {provider_name} APIs. "
        "Continue? (yes/no): "
    )
    sys.stderr.flush()
    try:
        return sys.stdin.readline().strip().lower() in ("yes", "y")
    except (EOFError, KeyboardInterrupt):
        return False
```

### Blocking Issue 2: `mcp._local_only` dynamic attribute is fragile

```python
mcp._local_only = local_only  # dynamic attribute on FastMCP instance
# ...
if getattr(mcp, '_local_only', False):  # read via getattr with default
```

This works today but is fragile — FastMCP could add a `_local_only` attribute with different semantics in a future version, causing silent misbehavior. A better approach is to use a module-level variable:

```python
_local_only: bool = False  # module-level in main.py

def main(repo_root: str | None = None, local_only: bool = False) -> None:
    global _local_only
    _local_only = local_only or os.environ.get("CRG_LOCAL_ONLY", "").lower() in ("1", "true", "yes")
```

Then in `embed_graph_tool`:
```python
from . import main as _main_module
if _main_module._local_only and provider and provider != "local":
    ...
```

Or, simpler: check `os.environ.get("CRG_LOCAL_ONLY")` directly in `embed_graph_tool` since the env var is already the documented interface.

### Minor issues

- `uv.lock` shows version `2.2.1` but current `main` is at `2.2.2` per the recent version bump commit. This will cause a conflict — drop the `uv.lock` change or rebase to pick up the current version.
- The `_check_cloud_embedding_config()` warning (prints to stderr on server startup) is fine and useful, but the wording "embeddings may send code" is imprecise — it only sends code if `embed_graph_tool` is called. Consider: "If `embed_graph` is called, code will be sent to external APIs."

### What's correct and should be kept

- `get_data_dir()` implementation — correct
- `.gitignore` skip for external dirs in `get_db_path()` — correct (this is better than #156)
- `CRG_REPO_ROOT` in `find_project_root()` — correct and valuable
- `--local-only` CLI flag concept — correct
- `CRG_EMBEDDING_PROVIDER` env var in `get_provider()` — correct
- All 10 tests for `get_data_dir` and `find_project_root` — correct

**Action:** Fix the two blocking issues (stdin/stdout corruption in `_confirm_cloud_usage`, module-level var for `_local_only`), rebase to pick up current `main` version, and this is ready.

tirth8205:

PR #234, opened 2026-04-11
feat: add Elixir (.ex/.exs) parsing support (#112)

azizur100389:
## Summary
Register `.ex` and `.exs` with tree-sitter-elixir. Extract modules as `Class` nodes (with dot-qualified names for nested modules), `def`/`defp`/`defmacro`/`defmacrop` as `Function` nodes with a visibility modifier, `alias`/`import`/`require`/`use` as `IMPORTS_FROM` edges, and ExUnit `test "name" do ... end` blocks as `Test` nodes attached to their enclosing module.

Closes #112.

## Root cause
Elixir was not listed in `EXTENSION_TO_LANGUAGE` and had no entries in any of the type mapping dicts, so `.ex` / `.exs` files were completely skipped by `build_or_update_graph`. Elixir's AST is unusual — **every structural construct is a `call` node** (`defmodule`, `def`, `alias`, `test`), so it cannot be handled by the normal `_FUNCTION_TYPES` / `_CLASS_TYPES` / `_IMPORT_TYPES` dict-mapping pattern and needs a dedicated constructs handler.

## Implementation approach
1. `EXTENSION_TO_LANGUAGE`: `.ex` and `.exs` → `"elixir"`
2. `_CLASS_TYPES["elixir"] = []`, `_FUNCTION_TYPES["elixir"] = []`, `_IMPORT_TYPES["elixir"] = []` — handled via constructs handler
3. `_CALL_TYPES["elixir"] = ["call"]` — regular calls still flow through the generic path
4. New `_extract_elixir_constructs()` dispatched next to `_extract_lua_constructs`. It intercepts `call` nodes whose first child is an `identifier` with text matching one of the Elixir structural keywords:
   - `defmodule` → `_handle_elixir_defmodule` (Class node + recurse into `do_block`)
   - `def` / `defp` / `defmacro` / `defmacrop` → `_handle_elixir_def` (Function node with visibility stored in `.modifiers`; recurses into `do_block` or short-form `keywords > pair`; handles `def foo, do: body` shortcut)
   - `alias` / `import` / `require` / `use` → `_handle_elixir_import` (IMPORTS_FROM edge with the original keyword recorded in `extra`)
   - `test` (only inside an enclosing module) → `_handle_elixir_test` (Test node named by the string literal)
5. For any other `call` node, the handler returns `False` and the generic CALLS machinery fires — so `Logger.info(x)`, `Helpers.log(y)`, and `assert x == y` all produce regular CALLS edges.
6. `_get_call_name` — new elixir branch for `first.type == "dot"` which returns the rightmost identifier (the method name in `Module.func()`).
7. **Nested defmodule**: qualified name is built by joining `enclosing_class` with the inner module name, so `defmodule Helpers` inside `Calculator` produces the class name `"Calculator.Helpers"`.

## What's captured
- Top-level and nested modules (`defmodule Calculator do ... defmodule Helpers do ... end end`)
- Public and private functions with `.modifiers` = `"public"` / `"private"`
- Short-form `def foo, do: body`
- Pipe-operator chains (`list |> Enum.filter(...) |> Enum.sum()`)
- All 4 import keywords: `alias` / `import` / `require` / `use`
- ExUnit `test "name" do ... end` → Test node; calls inside resolve correctly
- Local alias resolution: `Helpers.log(x)` inside a module that does `alias Calculator.Helpers` resolves to the qualified local `Calculator.Helpers.log` node
- External `Module.function(x)` calls → CALLS edge with the method name

## Tests added (`tests/test_multilang.py::TestElixirParsing` — 14 tests)
- `test_detects_language` (.ex and .exs)
- `test_finds_top_level_modules`
- `test_finds_nested_module_with_qualified_name`
- `test_finds_public_functions` (including pipe-operator and short-form def)
- `test_finds_private_function_with_visibility_modifier`
- `test_public_functions_marked_public`
- `test_finds_nested_module_functions` (parent = `"Calculator.Helpers"`)
- `test_finds_aliases_and_imports` (all 4 import keywords)
- `test_finds_module_dot_function_calls` (external bare name + local alias resolution)
- `test_finds_local_qualified_calls_in_tests`
- `test_detects_exunit_test_blocks`
- `test_tests_belong_to_calculator_test_module`
- `test_nodes_have_elixir_language`
- `test_finds_contains_edges` (file → module → function hierarchy)

**New fixture:** `tests/fixtures/sample.ex` exercises `@moduledoc`, all 4 import keywords (with options), `def`, short-form `def`, `defp`, pipe operator, nested `defmodule`, and a full ExUnit test module with three `test` blocks.

## Test results

| Stage | Result |
|---|---|
| Stage 1 — new targeted tests | **14/14 passed** |
| Stage 2 — `tests/test_multilang.py` full | **144/144 passed** — zero regressions across any language |
| Stage 3 — adjacent `tests/test_parser.py` | **67/67 passed** |
| Stage 4 — full suite | **709 passed**, 6 pre-existing Windows failures in `test_incremental`/`test_notebook` (verified identical on unchanged `main`) |
| Stage 5 — `ruff check` on changed files | **clean** |
| Stage 6 — fixture smoke parse | 13 nodes, 30 edges — all modules, nested module, public/private functions, and all 3 ExUnit tests detected |

**Zero regressions.** All new code is gated on the `elixir` language check so existing languages are untouched.

### DIFF
--- a/code_review_graph/parser.py
+++ b/code_review_graph/parser.py
@@ -104,6 +104,8 @@ EXTENSION_TO_LANGUAGE: dict[str, str] = {
     ".xs": "c",  # Perl XS: parsed as C to capture functions/structs/includes
     ".lua": "lua",
     ".luau": "luau",
+    ".ex": "elixir",
+    ".exs": "elixir",
     ".ipynb": "notebook",
 }
 
@@ -140,6 +142,7 @@ _CLASS_TYPES: dict[str, list[str]] = {
     "dart": ["class_definition", "mixin_declaration", "enum_declaration"],
     "lua": [],  # Lua has no class keyword; table-based OOP handled via constructs handler
     "luau": ["type_definition"],  # Luau type aliases; table-based OOP via constructs handler
+    "elixir": [],  # Elixir modules are `call` nodes, handled via constructs handler
 }
 
 _FUNCTION_TYPES: dict[str, list[str]] = {
@@ -175,6 +178,8 @@ _FUNCTION_TYPES: dict[str, list[str]] = {
     "dart": ["function_signature"],
     "lua": ["function_declaration"],
     "luau": ["function_declaration"],
+    # Elixir: def/defp are `call` nodes, handled via constructs handler
+    "elixir": [],
 }
 
 _IMPORT_TYPES: dict[str, list[str]] = {
@@ -201,6 +206,8 @@ _IMPORT_TYPES: dict[str, list[str]] = {
     # Lua/Luau: require() is a function_call, handled via _extract_lua_constructs
     "lua": [],
     "luau": [],
+    # Elixir: alias/import/require/use are `call` nodes, handled via constructs handler
+    "elixir": [],
 }
 
 _CALL_TYPES: dict[str, list[str]] = {
@@ -227,6 +234,7 @@ _CALL_TYPES: dict[str, list[str]] = {
     "solidity": ["call_expression"],
     "lua": ["function_call"],
     "luau": ["function_call"],
+    "elixir": ["call"],
 }
 
 # Patterns that indicate a test function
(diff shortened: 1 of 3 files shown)


azizur100389:
Closing this — v2.3.0 ships Elixir support as part of your super-PR #228, so there's nothing left to do here. Happy this one aligned with what you had in mind.

If you spot anything in my fixture/test coverage that isn't in the shipped version (e.g. short-form `def foo, do: body`, visibility stored in `modifiers`, local-alias resolution checks for `Module.function` calls), I'm happy to open a targeted follow-up issue rather than a competing PR.

Thanks for all the work on v2.3.0 and v2.3.1 — it's clear you're moving fast on the language-support backlog.

PR #236, opened 2026-04-11
feat(bash): add .ksh extension to bash parser (#235)

azizur100389:
## Summary
Register `.ksh` (Korn shell) with `tree-sitter-bash` alongside the existing `.sh` / `.bash` / `.zsh` entries shipped in #227. Korn shell is close enough to Bash syntactically that `tree-sitter-bash` handles the structural features the graph captures (function definitions, commands, `source`/`.` includes) correctly.

Closes #235.

## Why this PR
In the close comment on #230 you explicitly flagged this as worth a follow-up:

> If there's anything your implementation does that mine doesn't (e.g. .ksh extension, additional test coverage), please open a follow-up issue pointing at the specific feature and I'll merge it in. **The .ksh extension in particular looks worth adding — I didn't include it in #227.**

This PR is exactly that: the tracking issue (#235) plus the minimal change to address it.

## Why it matters
Korn shell is still used in legacy AIX/Solaris operations, IBM internal tooling, and enterprise CI scripts. Repositories that ship `.ksh` scripts currently index to `0 nodes` because the extension is unrecognized — the same failure mode that motivated #197.

## Implementation
A single line added to `EXTENSION_TO_LANGUAGE` in `code_review_graph/parser.py`:

```python
".ksh": "bash",  # Korn shell — close enough to bash for tree-sitter-bash (#235)
```

All of the bash parsing machinery shipped in #227 (`_FUNCTION_TYPES`, `_CALL_TYPES`, `_extract_bash_source_command`, name/call resolution) already handles any file routed through the `"bash"` language path — so no further wiring is needed.

## Tests added (`tests/test_multilang.py::TestBashParsing`)
1. **`test_detects_language`** — extended with a `.ksh` assertion to lock the mapping in as a regression guard.
2. **`test_ksh_extension_parses_as_bash`** — end-to-end regression test that copies `tests/fixtures/sample.sh` to a temp `legacy.ksh`, parses it through `CodeParser.parse_file()`, and asserts:
   - Every node's `language` field is `"bash"`
   - The set of extracted `Function` names is identical to the `.sh` run
   - The `CONTAINS` / `CALLS` / `IMPORTS_FROM` edge counts per kind match

The second assertion specifically proves `.ksh` is fully wired through the same structural extraction path as `.sh`, not a degenerate zero-result read.

## Test results

| Stage | Result |
|---|---|
| Stage 1 — new targeted tests | **2/2 passed** |
[...]

PR #238, opened 2026-04-11
feat(parser): shebang-based language detection for extension-less scripts (#237)

### LINKED ISSUE #237: feat(parser): shebang-based language detection for extension-less scripts
## Summary
`CodeParser.detect_language()` is currently extension-only (`path.suffix.lower()` lookup against `EXTENSION_TO_LANGUAGE`). Extension-less Unix scripts that start with a shebang (`#!/bin/bash`, `#!/usr/bin/env python3`, `#!/usr/bin/env node`, etc.) are silently skipped by the graph builder.

## Why it matters
A lot of production repos have critical scripts with no extension:

| Common location | Typical shebang | Today's result |
|---|---|---|
| `.git/hooks/pre-commit` | `#!/bin/bash` | 0 nodes |
| `bin/myapp` (project entry point) | `#!/usr/bin/env python3` | 0 nodes |
| `scripts/deploy` | `#!/bin/sh` | 0 nodes |
| `.husky/pre-push` | `#!/usr/bin/env sh` | 0 nodes |
| `tools/bootstrap` | `#!/usr/bin/env node` | 0 nodes |
| `installer`, `configure` | `#!/bin/bash` | 0 nodes |

These files are currently invisible to `callers_of`, `get_impact_radius`, `detect_changes`, and architecture mapping — even though the underlying bash / python / node parsers already handle them perfectly once routed.

## Proposed change
When the extension-based lookup in `detect_language()` returns `None` **and** the file has no extension (`path.suffix == ""`), read the first ~256 bytes and check for a shebang. Map common interpreters to the language that is already registered:

| Shebang interpreter | Maps to |
|---|---|
| `bash`, `sh`, `zsh`, `ksh`, `dash`, `ash` | `bash` |
| `python`, `python2`, `python3`, `pypy`, `pypy3` | `python` |
| `node`, `nodejs` | `javascript` |
| `ruby` | `ruby` |
| `perl` | `perl` |
| `lua` | `lua` |
| `Rscript` | `r` |
| `php` | `php` |

The check handles both direct form (`#!/bin/bash`) and env form (`#!/usr/bin/env bash`), tolerates trailing flags (`#!/bin/bash -e`), and short-circuits on missing/binary content.

## Non-goals
- **Does not** override extension-based detection. `.py` still means Python even if the first line is `#!/bin/bash` (which would be a malformed file).
- **Does not** add new languages — only routes extension-less files to languages that are *already* registered.
- **Does not** read bytes for files that already have a known extension (no perf regression on large repos).

## Follow-up PR
A PR implementing this with:
1. A new `_detect_language_from_shebang(path)` helper on `CodeParser`
2. `detect_language()` falls back to it for extension-less files
3. Targeted tests in `tests/test_parser.py` covering: direct shebang, env-form shebang, each mapped interpreter, missing shebang, binary content, and end-to-end `parse_file()` on a shebang-only script producing the expected function nodes.

azizur100389:
## Summary
Adds a shebang fallback to `CodeParser.detect_language()` so extension-less Unix scripts are routed to the correct tree-sitter grammar based on their first line. Closes #237.

| Common path | Typical shebang | Before | After |
|---|---|---|---|
| `.git/hooks/pre-commit` | `#!/bin/bash` | 0 nodes | Parsed as bash |
| `bin/myapp` | `#!/usr/bin/env python3` | 0 nodes | Parsed as python |
| `scripts/deploy` | `#!/bin/sh` | 0 nodes | Parsed as bash |
| `.husky/pre-push` | `#!/usr/bin/env sh` | 0 nodes | Parsed as bash |
| `tools/bootstrap` | `#!/usr/bin/env node` | 0 nodes | Parsed as javascript |

## Root cause
`detect_language()` was a single-line lookup against `EXTENSION_TO_LANGUAGE`. Any file with no extension returned `None`, which filters it out of both `incremental_update()` and `full_build()` before parsing. Real-world repos rely heavily on extension-less scripts for entrypoints, git hooks, CI installers, and shell tooling — all currently invisible to `callers_of`, `get_impact_radius`, `detect_changes`, and architecture mapping.

## Fix
1. **New module-level `SHEBANG_INTERPRETER_TO_LANGUAGE` table** mapping common interpreter basenames to languages **already registered** in `EXTENSION_TO_LANGUAGE`. This file strictly *routes* extension-less scripts to existing grammars — it does **not** introduce new languages:
   - `bash` / `sh` / `zsh` / `ksh` / `dash` / `ash` → `"bash"`
   - `python` / `python2` / `python3` / `pypy` / `pypy3` → `"python"`
   - `node` / `nodejs` → `"javascript"`
   - `ruby`, `perl`, `lua`, `Rscript`, `php`

2. **New `_SHEBANG_PROBE_BYTES = 256`** — maximum bytes read from the head when probing. Enough for any reasonable shebang line while keeping worst-case I/O tiny.

3. **New `CodeParser._detect_language_from_shebang(path)`** static method. Opens the file, reads up to 256 bytes, verifies `#!` prefix, splits on the first newline AND first NUL byte (defensive against binary), decodes UTF-8 strictly so malformed content returns `None` instead of raising. Handles:
   - Direct form: `#!/bin/bash`
   - `/usr/bin/env` indirection: `#!/usr/bin/env bash`
   - Linux `env -S` flag: `#!/usr/bin/env -S node --experimental-vm-modules`
   - Trailing flags: `#!/bin/bash -e`
   - Interpreter basename extraction from any absolute path
   - CRLF line endings

4. **`detect_language(path)`** now tries extension lookup first, and if it returns `None` **and** `path.suffix == ""`, falls back to the shebang probe. Files with a *known* extension are **never re-read** — extension-based detection remains authoritative.

## Non-regressions guaranteed by design
- `.py` files still parse as Python even if the first line is a misleading `#!/bin/bash` (locked in by `test_detect_shebang_does_not_override_extension`)
- Extension-less README / LICENSE files return `None` after a cheap 256-byte read that finds no shebang.
- Binary files whose first bytes are not `#!` return `None` without raising.
- Unknown interpreters (e.g. `#!/usr/bin/env ocaml`) return `None` — same semantics as an unmapped extension.
- No performance impact on files with a known extension — the shebang probe **only** runs for the `path.suffix == ""` branch.

## Tests added (`tests/test_parser.py::TestCodeParser` — 16 tests)
- `test_detect_shebang_bin_bash`
- `test_detect_shebang_bin_sh_routed_to_bash`
- `test_detect_shebang_env_bash`
- `test_detect_shebang_env_python3`
- `test_detect_shebang_direct_python`
- `test_detect_shebang_node`
- `test_detect_shebang_env_dash_s_flag`
- `test_detect_shebang_ruby`
- `test_detect_shebang_perl`
- `test_detect_shebang_with_trailing_flags`
- `test_detect_shebang_missing_returns_none`
- `test_detect_shebang_empty_file_returns_none`
- `test_detect_shebang_binary_content_returns_none`
- `test_detect_shebang_unknown_interpreter_returns_none`
- `test_detect_shebang_does_not_override_extension` (regression guard against extension override)
- `test_parse_shebang_script_produces_function_nodes` — end-to-end `parse_file()` check: an extension-less bash script is detected AND parsed into File + Function nodes, all tagged `language="bash"`.

## Test results

| Stage | Result |
|---|---|
| Stage 1 — new targeted shebang tests | **16/16 passed** |
| Stage 2 — `tests/test_parser.py` full | **83/83 passed** |
| Stage 3 — adjacent `tests/test_multilang.py` | **151/151 passed** |
| Stage 4 — full suite | **748 passed** (up from 733 baseline — +15 net), 8 pre-existing Windows failures in `test_incremental`/`test_main`/`test_notebook` (verified identical on unchanged `main`) |
| Stage 5 — `ruff check` | `code_review_graph/parser.py`: **clean**. `tests/test_parser.py`: 1 pre-existing `F841` on line 1038 (in `test_map_dispatch_qualified_reference`, unrelated to this PR — reproducible on unchanged `main` at line 901) |

**Zero regressions.** Purely additive fallback that only fires for files with no extension.

### DIFF
--- a/code_review_graph/parser.py
+++ b/code_review_graph/parser.py
@@ -119,6 +119,41 @@ EXTENSION_TO_LANGUAGE: dict[str, str] = {
     ".jl": "julia",
 }
 
+# Shebang interpreter → language mapping for extension-less Unix scripts.
+# Each key is the **basename** of the interpreter path as it appears after
+# ``#!`` (or after ``#!/usr/bin/env``).  Only languages already registered
+# above are listed — this file strictly routes extension-less scripts, it
+# does NOT introduce new languages on its own.  See issue #237.
+SHEBANG_INTERPRETER_TO_LANGUAGE: dict[str, str] = {
+    # POSIX / bash-compatible shells — all routed through tree-sitter-bash
+    "bash": "bash",
+    "sh": "bash",
+    "zsh": "bash",
+    "ksh": "bash",
+    "dash": "bash",
+    "ash": "bash",
+    # Python (every common variant)
+    "python": "python",
+    "python2": "python",
+    "python3": "python",
+    "pypy": "python",
+    "pypy3": "python",
+    # JavaScript via Node
+    "node": "javascript",
+    "nodejs": "javascript",
+    # Ruby / Perl / Lua / R / PHP
+    "ruby": "ruby",
+    "perl": "perl",
+    "lua": "lua",
+    "Rscript": "r",
+    "php": "php",
+}
+
+# Maximum bytes to read from the head of a file when probing for a shebang.
+# 256 is enough for any reasonable shebang line (``#!/usr/bin/env python3 -u\n``
+# is ~30 chars) while keeping the worst-case read tiny even on fat binaries.
+_SHEBANG_PROBE_BYTES = 256
+
 # Tree-sitter node type mappings per language
 # Maps (language) -> dict of semantic role -> list of TS node types
 _CLASS_TYPES: dict[str, list[str]] = {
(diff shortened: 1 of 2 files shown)


azizur100389:
Closing in favor of bundled PR #276 which combines this with #236 (.ksh extension) into a single parser-improvement PR.

PR #240, opened 2026-04-11
fix: three Windows-only bugs in v2.3.1 (#239)

### LINKED ISSUE #239: Three Windows-only bugs in v2.3.1 (CRLF Databricks detection, UTF-8 gitignore, stale FastMCP test API)
## Summary
While investigating v2.3.1 test output on Windows, I found three distinct bugs that all reproduce on `origin/main` with no modifications. They are not flakiness — each is a deterministic cross-platform bug with a clear root cause.

All three are currently masked by the 8 failing tests on Windows, which are easy to dismiss as "platform noise" but actually reveal real product bugs.

---

## Bug 1 — `get_data_dir()` writes non-UTF-8 `.gitignore` on Windows

**File:** `code_review_graph/incremental.py:128-138` (in `get_data_dir()`)

```python
inner_gitignore.write_text(
    "# Auto-generated by code-review-graph — do not commit database files.\n"
    "# The graph.db contains absolute paths and code structure metadata.\n"
    "*\n"
)
```

`write_text()` is called **without an `encoding=` argument**. The string literal contains an em-dash (`—`, U+2014). On Windows, `Path.write_text` uses the system default codepage (cp1252 in most locales), which encodes U+2014 as byte `0x97`. When anything later reads this file as UTF-8 — including `get_data_dir`'s own test `test_default_uses_repo_subdir` at `tests/test_incremental.py:194` — it raises:

```
UnicodeDecodeError: 'utf-8' codec can't decode byte 0x97 in position 38: invalid start byte
```

**Reproducible on:** Windows 11 / Python 3.10.10 / v2.3.1, clean checkout of `origin/main`. No modifications required.

**Impact:** Any Windows user running `code-review-graph build` generates a `.gitignore` that cannot be parsed by standard UTF-8 tooling (including the tool's own tests and any editor set to UTF-8 strict mode). Unix systems are unaffected because they default to UTF-8.

**Fix:** Add `encoding="utf-8"` to the `write_text()` call. (Same fix that `get_data_dir`'s caller `_ensure_repo_gitignore` at line 170 already applies — it uses `encoding="utf-8"`, but this sibling function was missed.)

---

## Bug 2 — Databricks notebook auto-detection fails on CRLF line endings

**File:** `code_review_graph/parser.py:391-394` (in `parse_bytes()`)

```python
if language == "python" and source.startswith(
    b"# Databricks notebook source\n",
):
    return self._parse_databricks_py_notebook(path, source)
```

The check hard-codes `\n`. On Windows, git checkouts default to `core.autocrlf=true`, which converts the Databricks fixture's line endings to `\r\n` on checkout. The file on disk starts with `b"# Databricks notebook source\r\n"`, so `source.startswith(b"# Databricks notebook source\n")` returns `False`, the file is parsed as regular Python, and **all Databricks-specific handling is bypassed**:

- `notebook_format` metadata is never set on the File node
- `# MAGIC %sql` cells are not parsed as SQL (no `IMPORTS_FROM` edges for table references)
- `# MAGIC %md` cells are not skipped
- Per-cell `cell_index` metadata is missing from function nodes

**Reproducible on:** Windows 11 / v2.3.1, clean checkout, confirmed with a single `python -c` statement reading `tests/fixtures/sample_databricks_export.py` — the file starts with `b'# Databricks notebook source\r\n'`.

**Impact:** Every one of the 4 `TestDatabricksPyNotebook` tests fails on Windows:
- `test_detects_databricks_header` — no `notebook_format`
- `test_extracts_sql_tables` — missing `bronze.events`, `silver.users`, etc.
- `test_skips_magic_md_cells` — extra function from MD cells that should be skipped
- `test_cell_index_tracking` — `cell_index` metadata missing

**Fix:** Parse the first line robustly, stripping any trailing `\r`:

```python
first_line_end = source.find(b"\n")
first_line = source[:first_line_end].rstrip(b"\r") if first_line_end != -1 else source
if language == "python" and first_line == b"# Databricks notebook source":
    return self._parse_databricks_py_notebook(path, source)
```

This matches both `\n` and `\r\n` endings.

---

## Bug 3 — Stale FastMCP API in async regression guard test

**File:** `tests/test_main.py:68-71` (in `TestLongRunningToolsAreAsync::test_heavy_tools_are_coroutines`)

```python
async def test_heavy_tools_are_coroutines(self):
    tools = await crg_main.mcp.get_tools()
```

`FastMCP` does not expose a `get_tools()` method. The available API in `fastmcp>=2.14.0` (the version pinned in `pyproject.toml`) is `list_tools()`. The test dies at `AttributeError: 'FastMCP' object has no attribute 'get_tools'. Did you mean: 'list_tools'?`.

This is not a cosmetic issue — this test is the **regression guard** for #46/#136 (v2.3.1 Windows async fix). The PR description for #231 explicitly says:

> There are regression tests (`test_heavy_tools_are_coroutines` + `test_heavy_tool_source_uses_to_thread`) that will fail at CI collection time if anyone converts one of the 5 tools back to sync in a future refactor, so we shouldn't hit this class of bug again.

But the test has been failing at **runtime** on every platform since it was written, which means the regression guard is **not actually guarding anything**. A future refactor that converts `build_or_update_graph_tool` back to sync would not be caught by CI because this test is already red.

**Reproducible on:** every platform with the pinned `fastmcp>=2.14.0`. Not Windows-specific.

**Fix:** Replace `get_tools()` with `list_tools()`. The rest of the test logic (verifying `asyncio.iscoroutinefunction(tool.fn)` for the heavy 5 tools) is correct.

---

## Follow-up PR
A single PR will fix all three bugs in one focused commit. Each fix is small, each has a direct regression test, and each is verifiable on unchanged `origin/main` before and after.

Note: there are 2 additional failing tests on Windows (`test_returns_none_without_git`, `test_falls_back_to_start` in `TestFindRepoRoot` / `TestFindProjectRoot`) that are true **test environmental flakiness** — they assume no ancestor of `tmp_path` has a `.git` directory, which is false on any Windows user whose home directory contains a git repo (e.g. dotfiles). Those are intentionally **not** addressed here because the fix requires either a product API change (`stop_at` parameter) or invasive monkeypatching — worth a separate discussion.

azizur100389:
## Summary
Three distinct, deterministic bugs that all reproduce on clean `origin/main` on Windows. They are **not flakiness** — each has a clear root cause and a targeted regression test that fails before the fix and passes after. All three were masking real product bugs behind the "CI is just flaky on Windows" narrative.

Closes #239.

---

## Bug 1 — `get_data_dir()` writes non-UTF-8 `.gitignore` on Windows

**File:** `code_review_graph/incremental.py:128-138`

`inner_gitignore.write_text(...)` was called **without an `encoding=` argument**. The string literal contains an em-dash (`—`, U+2014). On Windows, `Path.write_text` uses the system default codepage (cp1252 in most locales), which encodes U+2014 as the single byte `0x97`. Anything that later reads the file as UTF-8 — including `get_data_dir`'s own test `test_default_uses_repo_subdir` — raises:
```
UnicodeDecodeError: 'utf-8' codec can't decode byte 0x97 in position 38
```

**Fix:** Add `encoding="utf-8"` to the `write_text()` call. The sibling function `_ensure_repo_gitignore` at line 170 already uses this — this newer writer was simply missed when the inner `.gitignore` feature was added.

**Regression guard** — `TestDataDir::test_auto_gitignore_is_valid_utf8`:
- Reads the generated file as raw bytes
- Asserts the em-dash is stored as UTF-8 bytes `0xE2 0x80 0x94`
- Asserts cp1252 byte `0x97` does **not** appear
- Round-trips cleanly through strict UTF-8 decoding

---

## Bug 2 — Databricks notebook auto-detection fails on CRLF line endings

**File:** `code_review_graph/parser.py:390-394` (in `parse_bytes`)

The check was hard-coded to LF:
```python
if language == "python" and source.startswith(
    b"# Databricks notebook source\n",
):
```

On Windows, `git config core.autocrlf=true` (the default) rewrites text files to CRLF on checkout. The Databricks fixture starts with `b"# Databricks notebook source\r\n"` on any Windows checkout, so the check returns `False`, the file is parsed as regular Python, and **all Databricks-specific handling is bypassed**:
- `notebook_format` metadata is never tagged on the File node
- `# MAGIC %sql` cells are not parsed as SQL (no `IMPORTS_FROM` edges for table references)
- `# MAGIC %md` cells are not skipped
- Per-cell `cell_index` metadata is missing

This silently breaks **four** tests in `TestDatabricksPyNotebook` on Windows:
- `test_detects_databricks_header`
- `test_extracts_sql_tables`
- `test_skips_magic_md_cells`
- `test_cell_index_tracking`

**Fix:** Parse the first line robustly by finding the first `b"\n"` and stripping any trailing `b"\r"` before comparing to the exact header:

```python
if language == "python":
    first_newline = source.find(b"\n")
    first_line = (
        source[:first_newline].rstrip(b"\r")
        if first_newline != -1
        else source.rstrip(b"\r")
    )
    if first_line == b"# Databricks notebook source":
        return self._parse_databricks_py_notebook(path, source)
```

This matches both LF and CRLF endings **and** is strictly **more restrictive** than `startswith` — a file whose first line is `"# Databricks notebook source code examples"` will correctly **not** trigger Databricks parsing (regression guard `test_databricks_header_prefix_false_positive_rejected` locks this in).

**Regression guards** — `TestDatabricksPyNotebook`:
- `test_databricks_header_crlf_line_endings` — CRLF path (the actual bug)
- `test_databricks_header_lf_line_endings_still_work` — LF path unchanged
- `test_databricks_header_prefix_false_positive_rejected` — guards against a naive "just use `startswith`" fix

---

## Bug 3 — Stale FastMCP API in `test_heavy_tools_are_coroutines`

**File:** `tests/test_main.py:68-71`

`await crg_main.mcp.get_tools()` — `FastMCP.get_tools()` **does not exist** in `fastmcp>=2.14.0` (pinned via `fastmcp>=2.14.0,<3` in `pyproject.toml`). The current API is `list_tools()`, and it returns MCP protocol `Tool` pydantic objects that do **not** expose the underlying Python function at all — so the old lookup pattern cannot be directly ported.

The test dies at runtime with `AttributeError` on **every platform**. This means the async regression guard promised by PR #231 —

> There are regression tests (`test_heavy_tools_are_coroutines` + `test_heavy_tool_source_uses_to_thread`) that will fail at CI collection time if anyone converts one of the 5 tools back to sync in a future refactor, so we shouldn't hit this class of bug again.

— has been silently **inert** since the test was merged. A future refactor that converted any of the 5 heavy tools back to sync would **not** be caught by CI because the guard is already red.

**Fix:** Mirror the sibling `test_heavy_tool_source_uses_to_thread`, which resolves each heavy tool by `getattr(crg_main, name)` — a resilient approach independent of any FastMCP internal surface. Also drop `@pytest.mark.asyncio` on both guards since they no longer need an event loop.

**Regression guard** — `test_regression_guard_does_not_depend_on_fastmcp_internals`:
- AST-walks the two async-guard functions' source
- Fails if they reference `mcp.get_tools` / `mcp._tools` / `mcp.tool_manager` / `mcp._tool_manager` on actual `Attribute` nodes (docstrings ignored via `ast.walk`)
- Fails if any heavy tool cannot be resolved via `getattr(crg_main, name)`

This is a **meta-guard** — it protects the guards themselves from future FastMCP API drift.

---

## Test results

| Stage | Result |
|---|---|
| Stage 1 — new targeted regression tests | **7/7 passed** |
| Stage 2 — `test_incremental.py` + `test_notebook.py` + `test_main.py` | **99 passed + 2 xpassed**, 2 pre-existing failures in `TestFindRepoRoot`/`TestFindProjectRoot` (environmental — not addressed in this PR, see issue comment) |
| Stage 3 — adjacent `tests/test_parser.py` (parser.py touched) | **67/67 passed** |
| Stage 4 — full suite | **743 passed**, 165 pre-existing Windows teardown errors (unchanged). **+10 net tests and -6 pre-existing Windows failures resolved** vs baseline on `main` |
| Stage 5 — `ruff check` on all 5 changed files | **clean** |

## Why this fix is safe
- **Fix 1** (`encoding="utf-8"`) matches the established pattern in the sibling function in the same file. No API change.
- **Fix 2** uses a **more restrictive** match than the original (exact first line, not `startswith`), so it **cannot loosen** prior behavior — only the CRLF false negative is newly accepted. Prefix false positives are explicitly rejected by a test.
- **Fix 3** is test-only. The underlying product code (the 5 heavy tools) is unchanged. The regression guard is now actually enforced.

## Not addressed in this PR (intentional)
Two additional failing tests on Windows (`test_returns_none_without_git`, `test_falls_back_to_start`) are true environmental flakiness — they assume no ancestor of `tmp_path` has a `.git` directory, which is false on any Windows user whose home directory contains a git repo (e.g. dotfiles). Fixing these requires either a product API change (`stop_at` parameter on `find_repo_root`) or invasive monkeypatching. Left for a separate discussion, called out in the tracking issue.


azizur100389:
Closing in favor of bundled PR #274 which combines this with #242 to resolve all 8 Windows test failures in a single commit.

PR #268, opened 2026-04-13
perf: fix build hang on large repos by replacing per-row queries with…

jindalarpit:
### Problem

On large repositories (50K+ files, 333K+ nodes), `code-review-graph build` hangs indefinitely after community detection completes. The build appears stuck with no output after logging "Community detection complete: 34822 communities". This affects both CLI and MCP usage.

### Root Cause

`_compute_summaries()` in `code_review_graph/tools/build.py` was running per-row SQL queries in a loop:

- **Risk index**: 2 × N individual `COUNT(*)` queries (one for caller count, one for test coverage) for every Function/Class/Test node. On a 333K-node graph with 1.8M edges, this meant ~666K individual SQL queries.
- **Community summaries**: 1 heavy JOIN + GROUP BY + ORDER BY query per community (~35K queries for 34,822 communities).
- **Flow snapshots**: 1 `SELECT` per node ID in each flow path.

### Fix

Replaced all per-row queries with bulk operations:

| Phase | Before | After |
|-------|--------|-------|
| Risk index | 2 × N queries (~666K) | 2 bulk `GROUP BY` aggregates |
| Community summaries | 1 JOIN per community (~35K) | 1 window function query |
| Flow snapshots | 1 SELECT per path node | 1 bulk load of all node names |

Also added `logger.info()` calls at each postprocessing phase so users can see what stage the build is in, instead of silence after community detection.

### Changes

- `code_review_graph/tools/build.py`: Rewrote `_compute_summaries()` to use bulk aggregate queries and in-memory lookups instead of per-row SQL. Added progress logging for each phase.

### Testing

All 740 existing tests pass. Linting clean.

```bash
pytest tests/ --tb=short -q   # 740 passed
ruff check code_review_graph/  # All checks passed

### DIFF
--- a/code_review_graph/tools/build.py
+++ b/code_review_graph/tools/build.py
@@ -72,6 +72,7 @@ def _run_postprocess(
     use_incremental = not full_rebuild and bool(changed_files)
 
     try:
+        logger.info("Detecting execution flows...")
         if use_incremental:
             from code_review_graph.flows import incremental_trace_flows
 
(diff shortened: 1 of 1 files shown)


panga:
I can confirm this fix resolves the issue. In a larger repository with 5000+ files, the build no longer hangs.

tirth8205:
Superseded by #184 which covers all 3 sections with comprehensive tests and passing CI.

PR #275, opened 2026-04-14
fix(java): extract method name from identifier, not return type

krmahadevan:
tree-sitter-java places type_identifier (return type) before identifier (method name) in method_declaration nodes.
The generic _get_name() loop matched type_identifier first, causing methods to be indexed under their return type instead of their actual name.

For example:

* `public String getName()` was indexed as "String" instead of "getName", and
* `public ConfigBean getUtilityIngestionBean()`was indexed as "ConfigBean".

This broke callers_of,callees_of, and
children_of queries for any Java method with a non-void, non-generic return type.

Adds a Java-specific branch in _get_name() that returns the first identifier child for method_declaration nodes, following the same pattern as the Go fix (field_identifier) from PR #166.

Kotlin & Scala  is unaffected — its syntax places the name before the return type.

### DIFF
--- a/code_review_graph/parser.py
+++ b/code_review_graph/parser.py
@@ -3357,6 +3357,16 @@ class CodeParser:
             for child in node.children:
                 if child.type == "field_identifier":
                     return child.text.decode("utf-8", errors="replace")
+        # Java methods: tree-sitter-java puts type_identifier or generic_type
+        # (return type) before identifier (method name).  Must run before
+        # the generic loop, which would match the return type's
+        # type_identifier (e.g. "String", "ConfigBean").
+        # Constructors are fine — they have no return type node.
+        # Kotlin is unaffected: its syntax places the name before the type.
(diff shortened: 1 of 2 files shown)


krmahadevan:
@tirth8205 - Can you please help review this PR ? 

I am leveraging `code-review-graph` tool at work, to build a claude skill for code refactoring and I noticed this gap because of which Claude is falling back to using `Grep/Blob/File` instead of using this tool. That will increase token usage and so I was hoping that we could get this PR reviewed. 

This is the other PR https://github.com/tirth8205/code-review-graph/pull/278

krmahadevan:
This is the 3rd PR https://github.com/tirth8205/code-review-graph/pull/280

All of these will basically help us reduce claude token usage when working with java codebases wherein we use `code-review-graph` tool

krmahadevan:
@tirth8205 - ping! Gentle reminder, if you could please help with these PRs.

PR #292, opened 2026-04-15
fix: resolve Windows stdio JSON parsing errors and add Claude configuration docs

Thamindu-Dev:
### Describe the pull request
This PR addresses a critical blocker for Windows users where the MCP server fails to connect to Claude Code, throwing `Invalid JSON: EOF while parsing a value` and `Connection closed` errors during the `stdio` stream initialization.

**Root Causes & Fixes:**
1. **FastMCP Stdio Bug:** Older versions of `fastmcp` (locked at `<3`) fail to properly handle newline characters (`\n`) in Windows `stdio` streams. Bumped the `fastmcp` constraint to `>=3.2.4` and regenerated the `uv.lock` file to pull the upstream fix.
2. **Configuration Docs:** Added a dedicated section in the `README.md` for Windows users. Using the typical `cmd /c` wrapper in Claude's configuration injects unwanted terminal echoes into the stream, crashing the parser. Documented the direct `.exe` execution approach using `"env": {"PYTHONUTF8": "1"}` to ensure stable communication.

**Testing:**
Tested locally on Windows 11 with Claude Code CLI. The server now successfully connects and maintains the stream without parsing crashes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,7 +67,6 @@ Build the code review graph for this project
 
 The initial build takes ~10 seconds for a 500-file project. After that, the graph updates automatically on every file edit and git commit.
 
----
 
 ## How It Works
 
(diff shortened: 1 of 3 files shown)


tirth8205:
Thanks for the Windows fix and docs! However, the fastmcp dependency change from `>=2.14.0,<3` to `>=3.2.4` is a major version bump that could break existing users. Could you keep the docs/README changes but either: (a) keep fastmcp on v2, or (b) ensure backward compatibility with the fastmcp v3 API in a separate PR with proper migration notes?

Thamindu-Dev:
Thanks for reviewing and merging the docs! I understand the concern about the major version bump. I'll take a look at option (b) and see if I can put together a separate PR to safely migrate the codebase to the fastmcp v3 API.

Issue #297, opened 2026-04-15
PHP: method calls and plain function calls not extracted as CALLS edges


kojiromike:
## Summary

When building a graph on a large PHP codebase (~4,400 PHP files, 33K functions, 3K classes), the parser produces only **58 CALLS edges** from PHP files. Method calls (`$this->method()`, `$obj->method()`), static calls (`Class::method()`), and plain function calls (`sqlQuery(...)`, `xl(...)`) are not being captured.

CONTAINS and IMPORTS_FROM edges work well (36K and 12K respectively), so the file/class/function structure is there — but without call edges, blast-radius analysis is effectively disabled for PHP.

## Reproduction

```bash
git clone https://github.com/openemr/openemr.git
cd openemr
uvx code-review-graph build
sqlite3 .code-review-graph/graph.db "SELECT kind, COUNT(*) FROM edges WHERE file_path LIKE '%.php' GROUP BY kind ORDER BY COUNT(*) DESC;"
```

Output:

```
CONTAINS|36770
IMPORTS_FROM|12401
CALLS|58
```

The 58 CALLS edges that do exist are all global function calls with a `\` prefix — `\dirname()`, `\is_array()`, `\is_null()`, etc. Calls without the leading backslash, method calls, and static calls are all missing.

## Expected

PHP call extraction should capture at minimum:

- **Plain function calls**: `sqlQuery(...)`, `xl(...)`, `text(...)`
- **Method calls**: `$this->execute(...)`, `$service->search(...)`
- **Static method calls**: `QueryUtils::fetchRecords(...)`, `EncounterService::create(...)`
- **Scoped calls**: `parent::__construct(...)`, `self::factory(...)`

Tree-sitter's PHP grammar exposes these as `function_call_expression`, `member_call_expression`, `scoped_call_expression`, and `nullsafe_member_call_expression` node types, so the AST data is available.

## Environment

- code-review-graph installed via `uvx` (latest from PyPI as of 2026-04-15)
- Target repo: openemr/openemr at commit `a7a3f0a` (4,770 files, 64,600 nodes)
- macOS, Python 3.12

## Context

OpenEMR is a large open-source healthcare EHR. The PHP call extraction gap is the main blocker — the structural (CONTAINS) and import (IMPORTS_FROM) edges are already useful, but blast-radius analysis needs CALLS edges to be practical.

## Related issues

[...]

Issue #273, opened 2026-04-14
Add PHP support to language index


samuelyadav-a11y:
code-review-graph (v2.2.1) currently only indexes JavaScript, TypeScript, and C languages. PHP files in the codebase are not being parsed or indexed, preventing the MCP tools from searching and analyzing PHP code.

Expected Behavior:

PHP files should be automatically detected and indexed during graph build
PHP classes, functions, and relationships should be searchable via semantic_search_nodes_tool
Impact radius analysis should work for PHP code changes

tirth8205:
PHP support is already included. PR #298 (just merged) further improves PHP call extraction. Closing as already implemented.

PR #298, opened 2026-04-15
Fix PHP CALL extraction: add support for method, static, and unqualified function calls #297

Bakul2006:
Root cause: PHP node child-type mismatch in _get_call_name.
Fix: PHP-specific handling + scoped/nullsafe node coverage + regression tests. 
Addressing the issue #297  and #273 

Changes
Expanded PHP call node coverage in parser call node configuration.
Added PHP-specific call-name extraction logic

### DIFF
--- a/code_review_graph/parser.py
+++ b/code_review_graph/parser.py
@@ -275,7 +275,12 @@ _CALL_TYPES: dict[str, list[str]] = {
     ],
     "kotlin": ["call_expression"],
     "swift": ["call_expression"],
-    "php": ["function_call_expression", "member_call_expression"],
+    "php": [
+        "function_call_expression",
+        "member_call_expression",
+        "scoped_call_expression",
+        "nullsafe_member_call_expression",
+    ],
     "scala": ["call_expression", "instance_expression", "generic_function"],
     "solidity": ["call_expression"],
     "lua": ["function_call"],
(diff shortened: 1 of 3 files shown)


kojiromike:
Tested this PR locally against OpenEMR (4,770 files, a large PHP codebase). Results are dramatic:

| Metric | main | PR #298 |
|--------|------|---------|
| PHP CALLS edges | 58 | **145,423** |
| Total graph edges | 156,005 | 304,632 |

Spot-checking the top targets shows real PHP patterns now resolve correctly — method calls (`addChild`, `text`, `get`, `saveXML`), static calls (`OEGlobalsBag::getInstance`), framework/test calls (`assertEquals`, `assertSame`), and OpenEMR-specific globals (`sqlQuery`, `sqlStatement`, `xl`, `xlt`).

Sample blast-radius query on `sqlStatement` now returns real callers with correct file paths and line numbers — e.g. `C_Document.view_action` at multiple sites, `C_PatientFinder.search_by_FullName`, etc. Pre-PR this returned nothing.

The `scoped_call_expression` handling that constructs `Class::method` targets is especially valuable — it's what makes `OEGlobalsBag::getInstance` (and many other singletons/factories across the codebase) participate in caller graphs. Also good catch on `nullsafe_member_call_expression` (`$obj?->method()`) — that's increasingly common in modern PHP.

One small observation: the `_normalize_php_name` helper is currently defined as a closure inside `_get_call_name`. If this pattern repeats for other languages (e.g. Perl/Ruby also have sigil-prefixed names in some grammars), it might be worth promoting to a module-level helper. Just a thought for a follow-up.

LGTM from an end-user perspective. Thanks for the fast turnaround.

PR #319, opened 2026-04-18
feat(parser): comprehensive Julia language support

danvinci:
## Summary

Comprehensive Julia language support. The existing Julia hook in `parser.py` had a dead node type (`short_function_definition` — never emitted by tree-sitter-julia) and missed most real-world structural constructs, so Julia files collapsed to a single `File` node with raw-text imports. This PR routes Julia through a dedicated `_extract_julia_constructs()` helper (following the R / Lua / Elixir pattern) and fixes the name, bases, and import resolution paths.

## Changes

- **Fix:** remove `short_function_definition` (phantom type). Short-form `f(x) = expr` parses as `assignment` with a `call_expression` LHS — now handled by the new constructs helper.
- **Structs / modules / abstract types:** `module_definition`, `struct_definition`, and `abstract_definition` all produce Class nodes. `<:` inheritance via `type_head > binary_expression` emits INHERITS edges. Parametric types are handled on both sides of `<:`.
- **Functions:** `function_definition` and `macro_definition` as Function nodes. `_get_name` peels `signature > where_expression`, `signature > typed_expression`, and `field_expression` so that `function foo(x) where T`, `function foo(x)::RetT`, and `function Base.show(...)` all resolve to the method name. Parametric constructors (`Foo{T}(x) = ...`) resolve to `Foo`.
- **Qualified defs → cross-module edges:** `function Base.show(...)` emits a REFERENCES edge from the function to `Base` with a `julia_qualified_def` marker, so review tools can find Base extensions.
- **Nested functions:** `parent_name` and CONTAINS edges attach to the enclosing function rather than jumping past it to the module.
- **Imports:** structured `using` / `import` parsing — `using LinearAlgebra`, `using Pkg.Artifacts` (dotted), `using Stats: mean, std` (selective, recorded as `Stats.mean` / `Stats.std`). `include("file.jl")` emits IMPORTS_FROM.
- **Macrocalls:** `macrocall_expression` and `broadcast_call_expression` produce CALLS edges with `@`-prefixed names for macros. Non-testset macrocalls fall through so defs nested under `@inline` / `Base.@kwdef` / `@generated` still get picked up via recursion.
- **`@testset "name" begin ... end`:** Test node with the description in the name; recurses into the body so internal `@test` and helper calls get attributed to the testset.
- **`@enum Foo A B C`:** Class node for the type plus Function nodes for each variant, wired via CONTAINS. `julia_kind="enum"` / `"enum_variant"` extras for downstream filtering.
- **`export` / `public` (Julia 1.11+):** REFERENCES edges with `julia_export` / `julia_public` markers so "public surface of this module" is queryable.
- **Test file patterns:** `test/runtests.jl`, `test/*.jl`.

## Test plan

Verified locally on Python 3.12:

- [x] `uv run ruff check code_review_graph/` → `All checks passed!`
- [x] `uv run mypy code_review_graph/ --ignore-missing-imports --no-strict-optional` → `Success: no issues found in 50 source files`
- [x] `uv run pytest tests/ -q` → **812 passed**, 1 skipped, 2 xpassed (+24 new `TestJuliaParsing` cases)
- [x] Stress-tested on four real-world packages (DataFrames.jl, Agents.jl, Plots.jl, JuMP.jl — 153 files, 3.4k+ Functions, 17k+ edges, **zero parser errors**)
- [ ] CI matrix (3.10 / 3.11 / 3.12 / 3.13)

## Scope notes — intentionally not modeled

- `primitive type` (rare outside stdlib/numerics)
- `const` type aliases (more noise than signal for impact analysis)
- `do`-blocks as distinct lambda nodes (calls already attributed to the enclosing function)
- Multiple-dispatch method resolution (requires type inference; same limitation as Python/Ruby in this tool — each method def is a separate node, call edges resolve by bare name)

---

> **Disclosure:** drafted and implemented with Claude Opus 4.7 from a pre-written spec, then validated locally against the full test suite and four real-world Julia packages before submission.

### DIFF
--- a/code_review_graph/parser.py
+++ b/code_review_graph/parser.py
@@ -1667,13 +1667,19 @@ class CodeParser:
 
     def _julia_short_func_name(self, call_expr) -> Optional[str]:
         """Extract the name from a ``call_expression`` that is the LHS of
-        a short-form function ``f(x) = expr`` or ``Base.f(x) = expr``.
+        a short-form function ``f(x) = expr`` or ``Base.f(x) = expr`` or
+        ``Foo{T}(x) = expr``.
         """
         for child in call_expr.children:
             if child.type == "identifier":
                 return child.text.decode("utf-8", errors="replace")
             if child.type == "field_expression":
                 for ident in reversed(child.children):
+                    if ident.type == "identifier":
+                        return ident.text.decode("utf-8", errors="replace")
+                return None
+            if child.type == "parametrized_type_expression":
+                for ident in child.children:
                     if ident.type == "identifier":
                         return ident.text.decode("utf-8", errors="replace")
                 return None
@@ -1727,6 +1733,13 @@ class CodeParser:
         # ``=`` (plain variable, const) is left to the generic path.
         if node_type == "assignment":
             lhs = child.children[0] if child.children else None
+            # Unwrap typed LHS: ``f(x)::RetT = expr`` parses as
+            # ``assignment > typed_expression > call_expression``.
+            if lhs is not None and lhs.type == "typed_expression":
+                for sub in lhs.children:
+                    if sub.type == "call_expression":
+                        lhs = sub
+                        break
             if lhs is not None and lhs.type == "call_expression":
                 name = self._julia_short_func_name(lhs)
                 if name:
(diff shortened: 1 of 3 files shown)


tirth8205:
This PR now has merge conflicts with main after recent merges to parser.py and related files. Could you rebase on the latest main?

danvinci:
Thanks for handling the rebase and getting this merged - appreciate it.

PR #246, opened 2026-04-12
feat: lang handler refactor + parser improvements

gzenz:
## Summary
- Extract per-language logic from parser.py into 21 handler classes in `code_review_graph/lang/`
- Add `BaseLanguageHandler` with 7-method interface for language-specific parsing
- New language support: Luau, Objective-C, Bash, Elixir
- Parser accuracy: typed variable call resolution, star import expansion, method call noise filtering, Angular template parsing, Dart call extraction

Split from #158 (1/5). Independent -- targets main.

## Test plan
- [x] 690 tests pass (120 new), 0 failures
- [x] Ruff + mypy clean
- [x] All existing parser tests still pass (no regressions)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #158, opened 2026-04-08
feat: parser refactoring, Jedi call resolution, and performance optimizations

gzenz:
## Summary

Major improvements spanning parser architecture, call graph accuracy, build performance, and dead code detection:

- **Parser refactoring**: Extracted 16 per-language handler modules into `code_review_graph/lang/` using a strategy pattern, replacing monolithic conditionals in `parser.py`. Thread-safe parser caches with double-check locking.
- **Jedi-based call resolution**: New `jedi_resolver.py` resolves Python method calls at build time. Pre-scan filtering by project function names reduces enrichment from 36s to 3s on large repos. New `[enrichment]` optional dependency group.
- **PreToolUse search enrichment**: New `enrich.py` module and `code-review-graph enrich` CLI command inject graph context (callers, flows, community, tests) into agent search results passively via hook.
- **Call graph improvements**: Typed variable call enrichment (Python, JS/TS, Kotlin/Java), star import resolution, namespace imports, CommonJS `require()`, Angular template parsing, JSX handler tracking, module-qualified call resolution, function/class references as arguments.
- **Dead code FP reduction**: Framework decorators recognized as entry points, CDK construct methods and abstract overrides excluded, e2e test directories filtered.
- **Community detection 21x speedup**: Bulk node loading + adjacency-indexed cohesion computation (48.6s to 2.3s on 41k-node repos).
- **Build performance**: Batch file storage (50-file transactions), batch risk_index (2 GROUP BY queries replace per-node loops).
- **DB schema v8**: Composite edge index for upsert performance (v7 reserved by PR #127).
- **Other**: Weighted flow risk scoring, transitive TESTED_BY, `--quiet`/`--json` CLI flags, search deduplication, 829+ tests (up from 615).

## Evaluation

Tested against [Gadgetbridge](https://codeberg.org/Freeyourgadget/Gadgetbridge) (41k nodes, 280k edges):
- **8/10 scorecard PASS** (callers_of, callees_of, tests_for, communities, flows, impact radius, risk scores)
- Call resolution rate improved from 28% to 39.6%
- Community detection: 48.6s to 2.3s
- Full build time: ~22s for 3,574 files

## Migration note

Our composite edge index migration is numbered **v8** to avoid conflict with v6 (summary tables, already on main) and v7 (reserved by PR #127).

## Test plan

- [x] `uv run pytest tests/ --tb=short -q` -- 829 passed, 4 skipped
- [x] `uv run ruff check code_review_graph/` -- all checks passed
[...]

PR #334, opened 2026-04-19
feat(query): dedupe results and use transitive test lookup

gzenz:
## Summary

Query-graph returns deduped results and finds tests via transitive edges instead of direct-only.

## Stacked PR approach

This is PR 9/16 in a stack that ports closed #158 into small, reviewable slices based on your feedback. Each branch contains this commit plus every earlier branch in the stack, so it applies cleanly on its own. The intended merge order is bottom-to-top.

Stack (merge order):

2. feat/hardened-tests-v2


tirth8205:
This cumulative head `c5eb771` cannot merge: it is conflicting, Python 3.13 failed, and current main already contains caller/callee deduplication plus corrected bounded transitive-test lookup. I am closing this stale source and porting only the safe remaining `tests_for` behavior, preserving sanitized node responses and marking indirect matches. Silent non-test ambiguity resolution and raw transitive dictionaries will not be carried because they can select the wrong symbol and change the response schema. No contributor action is needed.

tirth8205:
The safe remaining tests_for delta is now completed through replacement PR #636, merged as 102f5a8d46e82840b45efc8939e9104e859d2436 with Gideon Zenz attribution. Current main uses the already-bounded GraphStore traversal, rebuilds results through the sanitized full node schema, marks direct and indirect matches explicitly in standard and minimal output, and keeps ambiguous bare symbols explicit. The unsafe silent production-node preference and stale stacked history were not carried. Focused graph/query tests, the full suite, and all 11 CI checks including Windows-native passed. No contributor action is needed.

PR #337, opened 2026-04-19
feat(parser): filter instance-method CALLS noise

gzenz:
## Summary

Suppress a small frozenset of well-known instance-method names (append, copy, items, keys, values, ...) from CALLS edges to cut cross-file false positives.

## Stacked PR approach

This is PR 12/16 in a stack that ports closed #158 into small, reviewable slices based on your feedback. Each branch contains this commit plus every earlier branch in the stack, so it applies cleanly on its own. The intended merge order is bottom-to-top.

Stack (merge order):

2. feat/hardened-tests-v2


tirth8205:
Head `0536c73` cannot merge because its global instance-method blocklist deletes legitimate user calls such as `get`, `find`, `write`, `remove`, and `update` from production graphs while retaining them in test files. Nested imported receivers can also be lost entirely; the commit adds no tests and includes unrelated `uv.lock` churn. Please preserve uncertain edges with confidence/noise metadata or resolve from imports/types, and provide precision/recall, blocked-name, nested-receiver, and performance regressions.

tirth8205:
Final maintainer disposition: this global method-name blocklist is not being merged.

The patch removes valid production calls named get, find, write, remove, update, and similar methods solely from their spelling, while treating test files differently and losing some nested imported receivers. That creates false negatives that downstream impact and review tools cannot recover.

Merged replacement #649 takes the evidence-based path instead: it qualifies calls when type/import evidence exists and preserves uncertain bare CALLS edges rather than deleting them. It passed 565 broader parser/language tests and all 11 GitHub checks including Windows. Closing this PR as not planned because the deletion policy is unsafe.

PR #338, opened 2026-04-19
feat(parser): enrich CALLS via typed-variable annotations

gzenz:
## Summary

When a variable has a type annotation (Python `x: Foo`, Kotlin `val x: Foo`), resolve `x.method(...)` to `Foo.method`.

## Stacked PR approach

This is PR 13/16 in a stack that ports closed #158 into small, reviewable slices based on your feedback. Each branch contains this commit plus every earlier branch in the stack, so it applies cleanly on its own. The intended merge order is bottom-to-top.

Stack (merge order):

2. feat/hardened-tests-v2


tirth8205:
Thanks — typed-variable resolution is useful, but current head is a cumulative, conflicting stack and cannot merge as-is. Its unique 559-line walker has no focused tests; same-file typed calls can target Type::method while the actual node is file::Type.method, and Java parameters/fields are not collected reliably.

Please rebuild this as one focused PR on current main, preserve lexical scope, collect parameters and fields before resolving calls, use current qualified-name conventions, deduplicate edges, and add Python, Kotlin, Java, and TypeScript tests for same-file, imported, generic, and shadowed types. Then run the full current CI suite.

tirth8205:
Final maintainer disposition: the safe typed-receiver behavior is now rebuilt and merged through replacement PR #649 at bfe1a13ed3aecaea173319964ee04fbf0e35384f.

The merged version:
- handles Python, Kotlin, Java, JavaScript, TypeScript, and TSX;
- collects receiver types with function, class, and block scope;
- requires same-file or repository-local import/type evidence;
- verifies the qualified target and retains uncertain calls instead of guessing;
- preserves the Julia resolver added after this source branch.

It passed 37 focused typed-call plus Julia overlap tests, 565 broader parser/language tests, and all 11 GitHub checks including Windows. Gideon Zenz is retained as co-author on the merged commit. Closing this cumulative branch as superseded.

PR #94, opened 2026-04-04
fix: resolve SQLite transaction bugs, FTS5 sync, and atomic operations

murilloimparavel:
## Summary

Fixes multiple SQLite transaction management issues that cause `cannot start a transaction within a transaction` errors and potential data corruption:

- **BEGIN IMMEDIATE conflict** (`graph.py`): Add `in_transaction` guard before `BEGIN IMMEDIATE` to prevent errors when implicit transactions are left open by prior DML operations
- **FTS5 content sync** (`search.py`): Recreate FTS5 table with `content='nodes'` and use `INSERT INTO nodes_fts(nodes_fts) VALUES('rebuild')` for proper content-sync population (was creating standalone table disconnected from nodes)
- **Atomic batch operations** (`flows.py`, `communities.py`): Wrap DELETE + INSERT loops in `BEGIN IMMEDIATE` / `try` / `COMMIT` / `except ROLLBACK` to prevent partial writes on crash
- **Missing rollback on error** (`tools/build.py`): Add `store.rollback()` in post-processing except blocks to prevent lingering transactions that corrupt the DB on next run
- **Public rollback API** (`graph.py`): Expose `GraphStore.rollback()` method instead of accessing `_conn` directly
- **Diagnostic logging**: Add `logger.warning()` before silent rollbacks in transaction guards for debuggability

## Reproduction

The `cannot start a transaction within a transaction` error occurs when:
1. `store_file_nodes_edges()` calls `BEGIN IMMEDIATE` while a prior operation (e.g., `set_metadata`, `upsert_node`) left an implicit Python sqlite3 transaction open
2. Post-processing steps (signatures, FTS, flows, communities) fail partially without rollback, leaving the DB in a dirty state for the next build

## Test plan

- [x] Full rebuild on monorepo with 8,352 nodes, 50,259 edges — zero transaction errors
- [x] Verified FTS5 content sync works correctly after rebuild
- [x] Verified flows (1,185) and communities (605) populated correctly
- [x] Verified rollback logging appears when uncommitted transactions are detected
- [x] All 29 CLI functional tests passing

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #342, opened 2026-04-19
chore: fix CI lint and type errors on main

gzenz:
## Summary

Main branch CI has been failing since PR #94. This fixes it.

- Remove 10 duplicate macOS-copy module files (`analysis 2.py`, `enrich 2.py`, `enrich 3.py`, `exports 2.py`, `exports 3.py`, `graph_diff 2.py`, `jedi_resolver 2.py`, `memory 2.py`, `memory 3.py`, `token_benchmark 2.py`) that tripped ruff `N999`.
- Sort imports in `main.py` (ruff `I001`).
- Drop unused `parse_git_diff_ranges` import in `tools/review.py` (ruff `F401`).
- Add `# type: ignore[attr-defined]` for `FastMCP._tool_manager` private attribute access in `main.py`.

Verified locally: `uv run ruff check code_review_graph/` and `uv run mypy code_review_graph/ --ignore-missing-imports --no-strict-optional` both pass.

## Context

This is PR 0/16 in a stack that ports closed #158. Merging this first turns main green so the 16 feature PRs (#326-#341) can show clean CI after rebase. See any of those PRs for the full stack description.


gzenz:
Heads-up on pre-existing test failures this PR surfaces (but does not introduce):

Before this PR, `lint` failed on `main`, so test jobs never got to run on PRs. Now that lint passes, four test failures show up on every PR in the stack. I believe all of them are pre-existing on `main`:

1. `AttributeError: 'FastMCP' object has no attribute '_tool_manager'` — upstream `fastmcp` dropped the private `_tool_manager` attribute. The code in `main.py:_apply_tool_filter` and the `--tools` test in `test_main.py` both need updating. Touches: `a3a043b feat(serve): add --tools flag`.
2. `AssertionError: [{'show_banner': ..., 'transport': 'stdio'}] == [{'transport': 'stdio'}]` — `fastmcp.run()` signature drift; test expects the old signature.
3. Parser regressions on the `main` tree (unrelated to this stack):
   - Godot `_ready` top-level function not detected
   - SQL `sqlQuery` target missing (now returns `\dirname`)
   - Generic class detection returning empty set for `SampleManager`, `Item`
   - Java `imports[].target` returning FQN (`com.example.auth.User`) instead of resolved file path
   - Java `getName` method missing from extracted names
   - INHERITS target leaking `implements ` / `extends ` prefix

Happy to fix (1) and (2) in this PR if you want — they're small (update `_tool_manager` usage to the new fastmcp API, align the test signature). (3) is a larger surface and probably belongs in a separate PR once this stack lands.

Tell me how you'd like to proceed.

gzenz:
Hi @tirth8205 — friendly nudge on this stack. Since these PRs were opened on 2026-04-19, ~10 PRs from other contributors have been merged into main (SQL support, Copilot platform, Windows stdio fix, daemon, opencode plugin, etc.) but the v8 stack hasn't received any review yet. I want to check: do you intend to look at #326–#342, or would you prefer I close them so I can plan accordingly?

These aren't cosmetic changes — they materially affect whether the tool is usable on larger codebases. From measurements on real-world repos:

- CALLS resolution rate: **~39% with the stack landed vs ~15% without** (current main)
- Dead-code findings: **~3× more false positives** without the stack — the review output gets dominated by noise
- Cross-file `ClassName.method` resolver lifts Kotlin/Java resolution from near-zero to ~33%

Without these landed, indexing and review on anything larger than a small Python repo gets buried under unresolved bare-name calls and dead-code false positives. Happy to split, rebase, expand tests, or rework anything that needs changes — but I need a signal that there's a path forward before continuing to keep the stack rebased on top of an active main.

Stack is currently rebased on `77284a9`, lint/type/security/schema-sync all green on every PR (the test failures are pre-existing main bugs catalogued [in this comment](https://github.com/tirth8205/code-review-graph/pull/342#issuecomment-4275463293)).

tirth8205:
Thanks @gzenz — and an apology first: this sat 51 days unanswered while we fixed main CI in parallel, exactly the silent overlap you were trying to avoid. That's on us.

Verdict: superseded. Main is now green at 0c9a5ff: `ruff check`, `mypy`, and the full suite (1274 passed) all pass. The duplicate-file removal landed in 0b4b268, and the fastmcp `_tool_manager` drift and parser regressions you catalogued in April are fixed too — your diagnosis was accurate and genuinely useful. This PR no longer applies (conflicts in `cli.py`/`query.py`; all 10 deleted files already gone), so we're closing it.

The stack is a different story: the CALLS-resolution and dead-code-precision improvements you measured are exactly what we want reviewed. Two asks: (1) rebase #326–#341 onto current main so CI signal is real; (2) this branch picked up ~2k lines of non-lint content (`parser.py` +1084, new `test_hardened.py`) — keep that in the feature PRs where it belongs. We want to review the rebased stack.

PR #353, opened 2026-04-20
fix: preserve Kotlin annotations + resolve C# namespace imports (#295, #310)

### LINKED ISSUE #295: Kotlin annotations (e.g. @Composable, @Inject) not preserved on Function/Class nodes
**Version**: code-review-graph 2.3.2 (Python 3.13, Windows 10)

## Summary

Kotlin nodes (Function / Class) have `modifiers=None` and `extra={}` on every row, even when the source contains annotations like `@Composable`, `@HiltViewModel`, `@Inject`, `@Test`, etc. This means consumers cannot filter or pivot queries on Kotlin annotation metadata — common Android use cases like "show me only @Composable functions" or "show me all @HiltViewModel classes" aren't expressible.

## Reproduction

Test repo: `D:\openclaw` (mixed monorepo, Kotlin Android app under `apps/android/`). After `code-review-graph build`:

```sql
-- How many Kotlin nodes have anything in modifiers or extra?
SELECT COUNT(*) FROM nodes
WHERE language='kotlin'
  AND (modifiers IS NOT NULL AND modifiers != '');
-- 0

SELECT COUNT(*) FROM nodes
WHERE language='kotlin'
  AND (extra IS NOT NULL AND extra != '{}' AND extra != '');
-- 0

-- Sanity: they do have signatures
SELECT name, signature FROM nodes
WHERE language='kotlin' AND signature IS NOT NULL
LIMIT 3;
-- CameraHudKind        | class CameraHudKind
-- CameraHudState       | class CameraHudState
-- DeviceNames          | class DeviceNames
```

The `signature` column has the bare `class X` / `fun x()` text but without annotations or visibility modifiers. The `modifiers` column (which on Python nodes sometimes carries decorators) is always NULL. The `extra` JSON blob is `{}`.

## Why this matters

On Android projects, the most common analytical queries are annotation-scoped:
- "All @Composable functions in this module" (UI surface area)
- "All @HiltViewModel classes" (DI graph)
- "All @Test or @JUnitTest functions" (test coverage mapping)
- "All @RequiresApi(N) callers" (min-SDK impact)

None of these are expressible against the current graph. Users have to either grep source files (bypassing the whole graph) or build a parallel light-weight indexer that re-scans Kotlin for annotation regex — defeating the point of having CRG in the pipeline.

## Suggested fix direction

The tree-sitter Kotlin grammar exposes `annotation` nodes directly on function and class declarations (`modifier_list` → `annotation`). The Kotlin extractor could collect annotation names (and optionally argument lists) into either:
1. A string column on nodes: `modifiers = "@Composable @RequiresApi(26)"` — matches the `signature` pattern, cheapest fix.
2. A JSON array in the `extra` column: `extra = {"annotations": ["Composable", "RequiresApi"]}` — more structured, easier to filter.

Option 2 lets FTS5 tokenize the annotation names for free, which is probably the nicest integration. Happy to try a patch if you can point me at the Kotlin extractor file.

## Scope note

This issue is about **Kotlin specifically**. I don't know if Swift has the same gap — Swift extraction works differently and I haven't verified annotation handling on Swift nodes. Filing only the Kotlin case for now.

Related: #294 (Kotlin + Swift CALLS edge extraction missing).

### LINKED ISSUE #310: Bug: C# IMPORTS_FROM edges target namespace strings, not file paths — importers_of / get_impact_radius return empty for .cs files
## Environment

- **code-review-graph version:** 2.3.2
- **Language affected:** C# (`.cs` files)
- **tree-sitter-c-sharp:** (bundled)

---

## Summary

For C# codebases, `query_graph importers_of <some.cs>` returns `[]` even when many other `.cs` files import the same namespace. As a result, `get_impact_radius` and any other feature built on the import graph is broken for `.cs` files.

The `IMPORTS_FROM` edges **are** being created for C# `using` directives — but their `target` is the raw namespace string (e.g. `ACME.AI.Agent.Swarm.Core`) rather than a resolved file path. `importers_of` looks edges up by the absolute file path of the target, so namespace-string targets never match.

```json
// query_graph { "pattern": "importers_of", "target": "...TaskBoard.cs" }
{ "results": [], "edges": [] }
```

The graph reports 3,857 nodes and 12,523 edges across 682 files (including ~200 C# files). The C# IMPORTS_FROM edges exist in the DB; they just don't point at file nodes.

---

## Root Cause (Code Analysis)

The C# `using_directive` is correctly registered as an import type and is correctly extracted as text — the problem is in the **resolution** step. Walking the C# code path in `parser.py` (v2.3.2):

### 1. C# `using_directive` is registered as an import

`parser.py` line 229:

```python
"csharp": ["using_directive"],
```

### 2. The text is extracted correctly

`parser.py` lines 3587–3591 (inside `_extract_import`):

```python
elif language in ("java", "csharp"):
    # import/using package.Class
    parts = text.split()
    if len(parts) >= 2:
        imports.append(parts[-1].rstrip(";"))
```

For `using ACME.AI.Agent.Swarm.Core;` this yields `"ACME.AI.Agent.Swarm.Core"`.

### 3. The edge target falls back to the raw namespace string

`parser.py` lines 2275–2286 (inside `_extract_imports`):

```python
imports = self._extract_import(child, language, source)
for imp_target in imports:
    resolved = self._resolve_module_to_file(
        imp_target, file_path, language,
    )
    edges.append(EdgeInfo(
        kind="IMPORTS_FROM",
        source=file_path,
        target=resolved if resolved else imp_target,   # <-- fallback to raw string
        file_path=file_path,
        line=child.start_point[0] + 1,
    ))
```

`_resolve_module_to_file` has no notion of C# namespaces — there is no namespace→file index — so `resolved` is always `None` and the edge `target` is the namespace string.

### 4. `importers_of` looks edges up by absolute file path

`tools/query.py` lines 249–263:

```python
elif pattern == "importers_of":
    abs_target = (
        str((root / target).resolve()) if node is None
        else node.file_path
    )
    for e in store.get_edges_by_target(abs_target):   # <-- exact file-path match
        if e.kind == "IMPORTS_FROM":
            results.append({...})
```

No edge has the absolute path of `TaskBoard.cs` as its `target` — the targets are namespace strings — so the result is empty. `get_impact_radius` traverses the same edges and is similarly broken.

(`imports_of`, by contrast, uses `get_edges_by_source` and returns `e.target_qualified` directly, so it should still surface the namespace strings rather than returning empty. Worth confirming on a real build — the symptom may be specific to `importers_of` and downstream features.)

---

## Reproduction

Any C# project. Example with a project containing ~200 `.cs` files:

```
# TaskBoard.cs is in namespace ACME.AI.Agent.Swarm.Core
# 40+ other .cs files do: using ACME.AI.Agent.Swarm.Core;

query_graph importers_of TaskBoard.cs   → [] (expected: 40+ files)
get_impact_radius      TaskBoard.cs     → only direct symbols, no importer fan-out
```

Inspecting the SQLite store directly confirms IMPORTS_FROM edges exist with namespace-string targets:

```sql
SELECT source_qualified, target_qualified
FROM edges
WHERE kind = 'IMPORTS_FROM' AND file_path LIKE '%.cs'
LIMIT 5;
-- target_qualified: 'System', 'System.Collections.Generic',
--                   'ACME.AI.Agent.Swarm.Core', ...
-- (namespace strings, not file paths)
```

---

## Expected Behaviour

| Pattern | Expected for C# |
|---|---|
| `importers_of SomeFile.cs` | All `.cs` files that `using` the namespace(s) declared in `SomeFile.cs` |
| `get_impact_radius SomeFile.cs` | Correct blast radius via import chains |
| `imports_of SomeFile.cs` | All namespaces declared in its `using` directives (already partly works, but as namespace strings rather than file targets) |

---

## Suggested Fix

The edge-creation path is fine; what's missing is **C# namespace → file resolution**. Sketch:

1. During parsing, collect every `namespace_declaration` (and file-scoped `namespace`) node from each `.cs` file into a `namespace → [file_paths]` index. (One namespace can span many files, so it's one-to-many.)
2. In `_resolve_module_to_file`, add a C# branch that consults this index and returns the canonical file path of (one of) the files declaring the namespace.
3. Decide the one-to-many policy: either (a) emit one IMPORTS_FROM edge per file in the namespace, or (b) emit a single edge to a representative file plus a follow-up traversal step. Option (a) is simpler and matches how `importers_of` is consumed today.

Happy to send a PR if there's a preferred direction.

---

## Related smaller bug: `_collect_import_names` has no C# handler

While tracing the above, I noticed `_collect_import_names` at `parser.py` line 2940 only handles `python`, `javascript`, `typescript`, `tsx`:

```python
def _collect_import_names(self, node, language, source, import_map):
    if language == "python":
        ...
    elif language in ("javascript", "typescript", "tsx"):
        ...
    # no csharp branch
```

`import_map` is therefore always empty for `.cs` files, which prevents resolving bare call targets (e.g. `Foo.Bar()` where `Foo` came in via a `using`) back to their source namespace. This is **separate** from the import-edge bug above and will need its own handler — but fixing it would round out C# call-graph accuracy alongside the namespace-resolution fix.

Sketch:

```python
elif language == "csharp":
    # using ACME.AI.Agent.Swarm.Core → {Core: ACME.AI.Agent.Swarm.Core}
    for sub in node.children:
        if sub.type == "qualified_name":
            full_ns = sub.text.decode("utf-8", errors="replace")
            short_name = full_ns.rsplit(".", 1)[-1]
            import_map[short_name] = full_ns
            break
```

---

## Additional Context

Tested on a project with ~200 `.cs` files (v2.3.2). After graph build:

- `Nodes: 3857`, `Edges: 12523`
- C# `IMPORTS_FROM` edges **do exist** in the DB — their targets are namespace strings, not file paths
- `callers_of` works (call edges are parsed correctly via `invocation_expression`)
- `importers_of`, `get_impact_radius`, and any other feature that joins through file-path-keyed import targets are the affected surface

Happy to submit a PR for either or both of the above if that would help.

azizur100389:
## Summary
Two parser correctness fixes that restore metadata and resolution paths lost by v2.3.2.

Closes #295, closes #310.

---

### Fix 1 — Kotlin (and Java/C#/Python) annotation preservation (#295)
`_extract_functions` already walked tree-sitter `modifiers > annotation` children to collect decorators, but used the list **only** for test-kind detection and then discarded it. Kotlin `@Composable`, `@HiltViewModel`, `@Inject`; Java `@Entity`, `@Autowired`; C# `[HttpGet]`; Python decorators — all were missing from the persisted node.

`_extract_classes` did not extract class-level annotations at all, so `@HiltViewModel class MyViewModel`, `@AndroidEntryPoint`, etc. lost their annotation metadata entirely.

**Fix:**
- New module-level `_extract_annotation_list(node)` helper that centralises the Java/Kotlin/C# `modifiers` walk and the Python `decorated_definition` walk.
- `_extract_functions` now persists annotations to `node.modifiers` (comma-separated string) AND `node.extra["decorators"]` (list) on every Function / Test node.
- `_extract_classes` calls the same helper to preserve class-level annotations.

Consumers can now write queries like "show me all `@Composable` functions" or "find `@HiltViewModel` classes" — both currently impossible on Kotlin codebases.

### Fix 2 — C# namespace-based `importers_of` resolution (#310)
C# `using ACME.Core;` directives produce an `IMPORTS_FROM` edge whose `target` is the raw namespace string (`"ACME.Core"`), not a file path. `importers_of` looked edges up by file path and never matched a namespace, so it returned `[]` for every `.cs` file — breaking `get_impact_radius`, `detect_changes`, and dead-code analysis on C# codebases.

**Fix:**
- New `_extract_csharp_namespaces(root_node)` helper handles both tree-sitter shapes:
  - block form: `namespace_declaration`
  - C# 10+ file-scoped: `file_scoped_namespace_declaration`
- `parse_bytes` tags each C# File node with the namespaces it declares in `extra["csharp_namespaces"]`.
- `tools/query.py::query_graph` adds a namespace fallback to `importers_of`: when the target is a `.cs` file, look up its declared namespaces and also search `edges_by_target` for each namespace string. Deduped against the primary lookup so duplicates never appear.

This mirrors the existing bare-name fallback already used by `callers_of` and `inheritors_of` (see #87). No parser restructuring needed — just an additional lookup path in the query layer.

---

## Tests added (10 new tests in `tests/test_multilang.py`)

**`TestKotlinAnnotations`** — 5 tests:
| Test | Purpose |
|---|---|
| `test_hilt_viewmodel_annotation_on_class` | `@HiltViewModel` persisted on Class node |
| `test_composable_annotation_on_function` | `@Composable` persisted on Function node |
| `test_multiple_annotations_on_function` | multiple `@Inject` + `@field:JvmField` both preserved |
| `test_unannotated_function_has_none_modifiers` | regression guard: no leak of empty string / empty list |
| `test_test_annotation_still_triggers_test_kind` | guard: pre-existing `@Test` → Test kind promotion still works |

**`TestCSharpNamespaceResolution`** — 5 tests:
| Test | Purpose |
|---|---|
| `test_file_scoped_namespace_tagged_on_file_node` | C# 10+ `namespace Foo;` form |
| `test_block_namespace_tagged_on_file_node` | classic `namespace Foo { ... }` form |
| `test_multiple_namespaces_in_one_file` | both namespaces kept |
| `test_non_csharp_file_has_no_namespace_tag` | scope guard |
| `test_importers_of_resolves_namespace_to_file` | end-to-end: write Core.cs + App.cs + Unrelated.cs, query `importers_of Core.cs`, assert App.cs is found and Unrelated.cs is not |

## Test results

| Stage | Result |
|---|---|
| New targeted tests | **10/10 passed** |
| `tests/test_multilang.py` full | 203 passed, 38 pre-existing failures (verified identical on unchanged `main`) |
| Full suite | **1016 passed**, 73 pre-existing failures (verified **identical set** on `main` via `comm -23` diff — zero regressions) |
| `ruff check` on all 3 changed files | **clean** |
| `mypy` on `parser.py` + `query.py` | **clean** |

**Zero regressions.** All fixes follow patterns already used in the codebase (inheritors_of bare-name fallback, File node extra tagging mirroring the Swift `swift_kind` pattern).

### DIFF
--- a/code_review_graph/parser.py
+++ b/code_review_graph/parser.py
@@ -591,6 +591,69 @@ def _is_test_file(path: str) -> bool:
     return any(p.search(path) for p in _TEST_FILE_PATTERNS)
 
 
+def _extract_csharp_namespaces(root_node) -> list[str]:
+    """Return the list of namespaces declared in a C# compilation unit.
+
+    C# supports two shapes:
+
+    * Block scoped:  ``namespace Foo.Bar { ... }``
+    * File scoped:   ``namespace Foo.Bar;``  (C# 10+)
+
+    A single file can declare multiple namespaces, so this returns a list
+    in source order.  Duplicates are preserved so the caller can reason
+    about multi-namespace files explicitly.  See #310.
+    """
+    namespaces: list[str] = []
+
+    def _walk(node) -> None:
+        # tree-sitter-c-sharp uses separate node types for the two shapes:
+        #  * ``namespace_declaration``              (block form)
+        #  * ``file_scoped_namespace_declaration``  (C# 10+ single-semicolon form)
+        if node.type in (
+            "namespace_declaration", "file_scoped_namespace_declaration",
+        ):
+            for child in node.children:
+                if child.type in ("qualified_name", "identifier"):
+                    text = child.text.decode("utf-8", errors="replace").strip()
+                    if text:
+                        namespaces.append(text)
+                    break
+        for child in node.children:
+            _walk(child)
+
+    _walk(root_node)
+    return namespaces
+
+
+def _extract_annotation_list(node) -> list[str]:
+    """Return the list of decorator / annotation names on *node*.
+
+    Walks:
+    * Java/Kotlin/C#: ``modifiers > annotation | marker_annotation`` child.
+    * Python: parent ``decorated_definition`` with ``decorator`` siblings.
+
+    Each entry has its leading ``@`` stripped.  Order follows source order so
+    the first annotation in the source is first in the returned list.  This
+    function replaces the inlined logic that previously extracted decorators
+    only for test detection and then discarded them (see #295).
+    """
+    deco_list: list[str] = []
+    for sub in node.children:
+        # Java/Kotlin/C#: annotations inside a ``modifiers`` child.
+        if sub.type == "modifiers":
+            for mod in sub.children:
+                if mod.type in ("annotation", "marker_annotation"):
+                    text = mod.text.decode("utf-8", errors="replace")
+                    deco_list.append(text.lstrip("@").strip())
+    # Python: check parent ``decorated_definition`` for decorator siblings.
+    if node.parent and node.parent.type == "decorated_definition":
+        for sib in node.parent.children:
+            if sib.type == "decorator":
+                text = sib.text.decode("utf-8", errors="replace")
+                deco_list.append(text.lstrip("@").strip())
+    return deco_list
+
+
 def _is_test_function(
     name: str, file_path: str, decorators: tuple[str, ...] = (),
 ) -> bool:
@@ -693,6 +756,14 @@ class CodeParser:
 
         # File node
         test_file = _is_test_file(file_path_str)
+        file_extra: dict = {}
+        # C#: capture every namespace this file declares so query-time
+        # fallbacks (importers_of, get_impact_radius) can resolve
+        # namespace-form IMPORTS_FROM targets back to file paths.  See #310.
+        if language == "csharp":
+            ns_list = _extract_csharp_namespaces(tree.root_node)
+            if ns_list:
+                file_extra["csharp_namespaces"] = ns_list
         nodes.append(NodeInfo(
             kind="File",
             name=file_path_str,
@@ -701,6 +772,7 @@ class CodeParser:
             line_end=source.count(b"\n") + 1,
             language=language,
             is_test=test_file,
+            extra=file_extra,
         ))
 
         # Pre-scan for import mappings and defined names
@@ -2741,11 +2813,22 @@ class CodeParser:
         if not name:
             return False
 
+        # Extract class-level annotations (Kotlin/Java/C#/Python).  Without
+        # this, Kotlin `@HiltViewModel`, `@AndroidEntryPoint`, `@Composable`
+        # classes, Java `@Entity`/`@Service` classes, etc. would lose their
+        # annotation metadata.  See #295.
+        class_decorators = _extract_annotation_list(child)
+        class_modifiers: Optional[str] = (
+            ",".join(class_decorators) if class_decorators else None
+        )
+
         # Swift: detect the actual type keyword (class/struct/enum/actor/extension)
         # and store it in extra["swift_kind"] for richer downstream analysis.
         # Tree-sitter maps struct/enum/actor/extension all to class_declaration;
         # protocol uses its own protocol_declaration node type.
         extra: dict = {}
+        if class_decorators:
+            extra["decorators"] = list(class_decorators)
         if language == "swift":
             if child.type == "class_declaration":
                 _swift_keywords = {"class", "struct", "enum", "actor", "extension"}
@@ -2765,6 +2848,7 @@ class CodeParser:
             line_end=child.end_point[0] + 1,
             language=language,
             parent_name=enclosing_class,
+            modifiers=class_modifiers,
             extra=extra,
         )
         nodes.append(node)
@@ -2829,22 +2913,16 @@ class CodeParser:
             if receiver_type:
                 enclosing_class = receiver_type
 
-        # Extract annotations/decorators for test detection
+        # Extract annotations/decorators for test detection AND persistence.
+        # Previously ``deco_list`` was used only for test detection and
+        # discarded — see #295 for Kotlin (``@Composable``, ``@HiltViewModel``,
+        # ``@Inject``) and the same pattern affects Java (``@Test``,
+        # ``@Autowired``), C# (``[HttpGet]``), and Python (``@app.get``).  Now
+        # persisted into ``node.modifiers`` (comma-separated string) and
+        # ``node.extra["decorators"]`` (list) so consumers can filter by
+        # annotation (e.g. "show me all @Composable functions").
         decorators: tuple[str, ...] = ()
-        deco_list: list[str] = []
-        for sub in child.children:
-            # Java/Kotlin/C#: annotations inside a modifiers child
-            if sub.type == "modifiers":
-                for mod in sub.children:
-                    if mod.type in ("annotation", "marker_annotation"):
-                        text = mod.text.decode("utf-8", errors="replace")
-                        deco_list.append(text.lstrip("@").strip())
-        # Python: check parent decorated_definition for decorator siblings
-        if child.parent and child.parent.type == "decorated_definition":
-            for sib in child.parent.children:
-                if sib.type == "decorator":
-                    text = sib.text.decode("utf-8", errors="replace")
-                    deco_list.append(text.lstrip("@").strip())
+        deco_list: list[str] = _extract_annotation_list(child)
         if deco_list:
             decorators = tuple(deco_list)
 
@@ -2853,6 +2931,13 @@ class CodeParser:
         qualified = self._qualify(name, file_path, enclosing_class)
         params = self._get_params(child, language, source)
         ret_type = self._get_return_type(child, language, source)
+
+        modifiers_str: Optional[str] = (
+            ",".join(deco_list) if deco_list else None
+        )
+        extra: dict = {}
+        if deco_list:
+            extra["decorators"] = list(deco_list)
 
         node = NodeInfo(
             kind=kind,
@@ -2864,7 +2949,9 @@ class CodeParser:
             parent_name=enclosing_class,
             params=params,
             return_type=ret_type,
+            modifiers=modifiers_str,
             is_test=is_test,
+            extra=extra,
         )
         nodes.append(node)
 
--- a/code_review_graph/parser.py
+++ b/code_review_graph/parser.py
@@ -785,6 +785,72 @@ def _is_test_function(
     return False
 
 
+def _modifier_annotation_names(node) -> list[str]:
+    """Return annotation names from a ``modifiers`` child of *node*.
+
+    Covers Java/Kotlin/C# where annotations live inside a ``modifiers``
+    node as ``annotation`` / ``marker_annotation`` children. The leading
+    ``@`` is stripped. See: #295
+    """
+    names: list[str] = []
+    for sub in node.children:
+        if sub.type == "modifiers":
+            for mod in sub.children:
+                if mod.type in ("annotation", "marker_annotation"):
+                    text = mod.text.decode("utf-8", errors="replace")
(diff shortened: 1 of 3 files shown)


azizur100389:
Heads up on CI: the 3 failing checks (lint, type-check, test) are **pre-existing baseline failures on `upstream/main`**, not regressions from this PR. Verified by running each check on my branch *and* on a clean checkout of `b0f8527` (the merge base) and getting identical failure sets.

| Check | Failure source | Mine? |
|---|---|---|
| `lint` | 9× `N999 Invalid module name` for accidentally-committed macOS Finder duplicates (`memory 2.py`, `analysis 2.py`, `enrich 2.py`, `exports 2.py/3.py`, `graph_diff 2.py`, `jedi_resolver 2.py`, `token_benchmark 2.py`, `memory 3.py`) + 1× `I001` import sort in `main.py:9` + 1× `F401` unused import in `tools/review.py:9` | No |
| `type-check` | `main.py:945: "FastMCP[Any]" has no attribute "_tool_manager"` | No |
| `test (3.10/.11/.12/.13)` | 61 failures, including 6 errors in `test_main.py::TestApplyToolFilter` from the same `_tool_manager` issue, plus the same set that fails on unmodified `main` | No |

All three are addressed by @gzenz's existing **#342 "chore: fix CI lint and type errors on main"**.

For the actual changes in this PR:
- 10/10 new regression tests in `TestKotlinAnnotations` and `TestCSharpNamespaceResolution` pass locally on Python 3.10
- Full `tests/test_multilang.py` shows 38 failures = same set that fails on unmodified `main` (zero regressions, verified via `comm -23` diff)
- `ruff check` clean on all 3 files I touched (`parser.py`, `tools/query.py`, `tests/test_multilang.py`)
- `mypy` clean on `parser.py` and `tools/query.py`

Happy to rebase once #342 lands.

tirth8205:
@azizur100389 Thanks — this is careful work: the parse->store->query regression test for #310 and the modifiers=None guard are exactly the kind of tests we want. I applied the branch, resolved conflicts, and ran the full suite: 1284 passed, your 10 new tests all green, ruff and mypy clean. Verdict: this is mergeable after a rebase.

Concretely: (1) Rebase on main — `_extract_functions` now has a Rust `attribute_item` walk and Java Temporal/Kafka `method_extra` logic in the same block; keep the Rust walk appending to `deco_list` and merge your `decorators` key into `method_extra` rather than a second dict. (2) Scope note: the namespace fallback fixes `importers_of`, but `get_impact_radius`/`detect_changes` BFS in graph.py never sees it, so please trim that claim from the description — we'll keep #310 open for the impact-radius half. (3) C# `[HttpGet]` uses `attribute_list` nodes, not `modifiers > annotation`, so it isn't actually captured; drop the claim or add support plus a test.

Rebase and this lands.

azizur100389:
Thanks for the detailed review @tirth8205 — rebased onto current main and addressed all three points:

**1. Rebase / `method_extra` merge.** Done. The function path keeps main's inline `deco_list` (including the Rust `attribute_item` walk) and now merges `decorators` into the existing `method_extra` dict rather than a second dict. `modifiers` is set from the same list. Java's `temporal_role`/`kafka_listener` keys are untouched.

**2. Scope claim trimmed.** You're right — the namespace fallback only fixes `importers_of`; `get_impact_radius`/`detect_changes` BFS in `graph.py` never sees it. I've removed the impact-radius claim from the description; #310 stays open for that half.

**3. C# `[HttpGet]`.** Rather than drop the claim, I added real support: `[HttpGet]`/`[Authorize]`/`[ApiController]` are `attribute_list` nodes (not `modifiers > annotation`), so I added `_csharp_attribute_names()` and wired it into both the function and class paths — with a test (`TestCSharpAttributes`) proving `Get` gets `decorators=['HttpGet','Authorize']`.

New tests: `TestKotlinAnnotations` (4), `TestCSharpAttributes` (3), `TestCSharpNamespaceResolution` (4) — 11 total, all green. Full suite shows only the 3 pre-existing collection errors that also fail on clean main (verified via `comm` diff); zero regressions. `ruff` + `mypy` clean on the changed files.

Single squashed commit on top of `543de44`. Ready for another look.

tirth8205:
Thank you for addressing each review point. #624 has now merged the annotation persistence, real C# `attribute_list` support, namespace-based `importers_of`, and regressions with your authorship; the reconciliation also added nested-namespace and deep-AST fixes. This PR is fully superseded and can close. Issue #310 remains open for the separate impact-radius/`detect_changes` BFS half, which this PR intentionally did not solve.

PR #393, opened 2026-04-27
fix(parser): repair Zig support — grammar nodes were never matched

geeknik:
Zig was registered in EXTENSION_TO_LANGUAGE and the per-language node-type tables, but the mapped types (fn_proto, fn_decl, container_declaration, call_expression, builtin_call_expr) don't exist in the tree-sitter-zig grammar shipped by tree_sitter_language_pack — that grammar emits PascalCase nodes (FnProto, VarDecl, SuffixExpr, FnCallArguments, FieldOrFnCall, BUILTINIDENTIFIER, ContainerDecl, TestDecl). Result: every Zig file produced only its File node, with zero functions, structs, imports, or calls.

Rather than rename and lose information, route Zig through a constructs handler (mirroring the Lua/Elixir pattern) so we can bridge FnProto's sibling Block, distinguish struct/union/enum/opaque ContainerDecl kinds, and split @import out as IMPORTS_FROM rather than CALLS.

- Empty _CLASS_TYPES/_FUNCTION_TYPES/_IMPORT_TYPES/_CALL_TYPES for "zig" with comments explaining why generic dispatch can't be used.
- Add _extract_zig_constructs dispatch in _extract_from_tree.
- Implement _handle_zig_fn_decl, _handle_zig_var_decl, _handle_zig_test_decl, _zig_extract_import_target, and _extract_zig_calls_in_subtree. The call walker emits CALLS for any SuffixExpr/FieldOrFnCall whose direct children include FnCallArguments, resolving the callee to the immediately preceding IDENTIFIER or BUILTINIDENTIFIER. @import is filtered out so it isn't double-counted.
- Add a Zig branch to _resolve_module_to_file: relative @import("./x.zig") resolves to an absolute file path; std and build-graph module names stay unresolved (resolving the latter would require parsing build.zig).
- Add tests/fixtures/sample_zig.zig + sample_zig_util.zig covering top-level fns, struct with methods, enum, tagged union, std + relative @import, @intCast builtin call, and a labelled test block.
- Add TestZigParsing to tests/test_multilang.py (11 tests).

Verified on a 41-file Zig project: before the fix, build produced 0 nodes / 0 edges; after, 393 unique nodes (214 Function, 124 Test, 55 Class) and 2839 edges (2353 CALLS, 393 CONTAINS, 93 IMPORTS_FROM), with hub detection correctly surfacing Reader.next, sanitize helpers, and Response.deinit as the most-called sites.


geeknik:
Those CI failures aren't from my PR.

tirth8205:
I reviewed this during release triage and did not merge it.

Zig parser correctness is relevant, but this branch needs to be brought up to date and validated with focused parser tests before it is safe to include. It should be handled in a parser correctness pass rather than in the v2.3.4 release.

tirth8205:
@geeknik Thanks for this — and for the unusually thorough diagnosis. I verified your claim from scratch: on current main, the language pack's Zig grammar emits PascalCase nodes, so the registered snake_case types never match and a Zig file yields exactly one File node and zero edges. With your patch applied, my scratch fixture produces the full expected structure (Functions, Class with method, Test node, CALLS/IMPORTS_FROM/TESTED_BY), and your 11 fixture tests plus the entire suite (1285 tests), ruff, and mypy all pass.

You were also right about the CI failures — they came from stray duplicate files on the old base, since cleaned up. The constructs-handler approach mirroring Lua/Elixir is exactly the right pattern.

One ask: the branch now conflicts with main (Nix support landed adjacent to your table entries, and new test classes were appended where TestZigParsing sits). The conflicts are mechanical keep-both resolutions — please rebase onto main and re-push, and this is ready to merge. Two optional follow-ups I'll file separately: nested containers inside fn bodies, and line-number-based test node names.

tirth8205:
Thank you for the precise Zig grammar diagnosis and implementation. #624 carried the Zig parser work into current `main` with your authorship, resolved the adjacent Nix conflicts, and added follow-up regressions/fixes for in-source Zig tests. This branch has not changed since the audited revision, so the PR is fully superseded. Closing the duplicate.

PR #394, opened 2026-04-27
fix: make get_minimal_context_tool async via asyncio.to_thread

sdeonvacation:
## Summary

- Offloads blocking git-diff + BFS traversal work in `get_minimal_context_tool` to a thread via `asyncio.to_thread`, preventing event-loop blocking on large repos
- Adds `get_minimal_context_tool` to `HEAVY_TOOLS` coverage in `TestLongRunningToolsAreAsync`

## Motivation

`get_minimal_context_tool` was the last high-traffic tool missing the `asyncio.to_thread` wrapper already applied to other blocking tools (see #136, #46). Without this fix, large-repo calls block the MCP stdio event loop.

## Changes

- `code_review_graph/main.py`: wrap `get_minimal_context_tool` handler body in `asyncio.to_thread`
- `tests/test_main.py`: include `get_minimal_context_tool` in `HEAVY_TOOLS` list


sdeonvacation:
The failing PR checks fail in upstream as well.

tirth8205:
I left this out of v2.3.4.

The current main branch already has async guards around the long-running MCP tools, and the release added further bounds around detect-changes. Please rebase this branch and describe the remaining failure it fixes, if it is still reproducible.

tirth8205:
Thanks @sdeonvacation — I applied this on current main and everything is green locally (98 passed across test_tools/test_main, ruff and mypy clean), and extending the HEAVY_TOOLS regression guard is exactly the right place for this.

One important finding though: since we pinned fastmcp>=3.2.4, FastMCP runs sync tool handlers in an anyio threadpool (call_sync_fn_in_threadpool in tools/function_tool.py), so the sync get_minimal_context_tool no longer blocks the event loop. The original bug you cite from #46/#136 isn't reproducible on main. I still want this as a consistency/defense-in-depth change so we don't depend on FastMCP internals — but framed honestly.

To get it merged:
1. Rebase on main — the import-reorder and show_banner hunks already landed, and the April lint/type-check failures were from the stale base.
2. Reword the new docstring: drop the "don't block the stdio event loop" claim; say it keeps the heavy tools consistent and independent of FastMCP's sync-dispatch behavior.
3. Update "the 5 long-running MCP tools" to 6 in the TestLongRunningToolsAreAsync docstring.

Happy to merge once those land.

tirth8205:
The June requests remain outstanding at head `bacca6a`: rebase on current main, describe `asyncio.to_thread` as consistency/defense-in-depth rather than a currently reproducible FastMCP event-loop fix, and update the test docstring from five heavy tools to six. Lint, type-check, and Python 3.13 are red. Issues #46 and #136 stay closed because the original deadlock is fixed; this PR is additional hardening.

PR #365, opened 2026-04-21
fix: restore 8 silently-reverted parser fixes + repair _apply_tool_filter for fastmcp>=3 (closes #361)

npkriami18:
Closes #361.

## Result

```bash
# Before (upstream/main @ b0f8527)
60 failed, 1015 passed, 1 skipped, 2 xpassed, 6 errors

# After (this branch)
0 failed, 1081 passed, 1 skipped, 2 xpassed, 0 errors
```

A single, self-contained PR that takes the suite to fully green.

### Class 1 — silently-reverted parser PRs (9 cherry-picks, original authorship preserved)

Two `chore: resolve merge conflicts with main` commits (`3cf31ec`, `cb9d25a`) accepted the stale feature-branch side of `code_review_graph/parser.py` for the ReScript PR (#323) and Julia PR (#319). Both feature branches forked from `db2d2df` — before the fixes below landed. Because only `main` had touched those regions, `git merge` auto-resolved silently in the wrong direction (no conflict markers).

| Commit | PR | Author | Tests restored |
|---|---|---|---|
| `fe383c7` | #285 | @michael-denyer | 7 module-scope CALLS |
| `e6e3144` | #276 | @azizur100389 | 10 shebang-detection + `.ksh` |
| `aa627fb` | #274 | @azizur100389 | 1 Databricks CRLF |
| `f092922` + `425810b` + `536fd4b` | #319 | @danvinci | 21 Julia parser |
| `8104eb7` | #278 | @krmahadevan | 3 Java inheritance |
| `9a88f20` | #275 | @krmahadevan | 1 Java method-name extraction |
| `088c281` | #280 | @krmahadevan | 2 Java import resolution |
| `112a442` | #316 | @Mornor | 9 GDScript |
| `d660e02` | #298 | @bakulb-bansal | 1 PHP CALL extraction |

### Class 2 — fastmcp>=3 API drift (1 commit)

`a3a043b` (`feat(serve): add --tools flag and CRG_TOOLS env var for MCP tool filtering`) was authored against the **fastmcp 2.x** internal API. The pinned `fastmcp>=3` removed `mcp._tool_manager._tools` and `mcp.get_tools()`, leaving:

- `_apply_tool_filter` raising `AttributeError` whenever `--tools` or `CRG_TOOLS` was set (6 test errors in `TestApplyToolFilter`)
- a stale `test_stdio_calls_mcp_run_stdio` assertion (PR #290 had added `show_banner=False` to the production call without refreshing the test)

This PR:

[...]

PR #396, opened 2026-04-27
fix: replace _tool_manager with local_provider for FastMCP 3.x tool-filter compat

sdeonvacation:
## Problem

FastMCP 3.x removed `_tool_manager` from the `FastMCP` instance. Code-review-graph used `mcp._tool_manager` to enumerate and filter tools via `CRG_TOOLS`, causing an `AttributeError` startup crash on FastMCP ≥ 3.0.

## Fix

Switch the tool-filter implementation to use `mcp.local_provider`:
- Enumerate registered tools via `mcp.local_provider._components` (keys prefixed `tool:`)
- Remove unwanted tools via `mcp.local_provider.remove_tool(name)`

This is API-stable in FastMCP 3.x and fully backwards-compatible with the existing `CRG_TOOLS` env-var contract.

## Testing

Updated `tests/test_main.py` to patch `local_provider` instead of `_tool_manager`.

## Related

Split from PR #394 (`fix/get-minimal-context-async`) to keep concerns separate.

### DIFF
--- a/code_review_graph/main.py
+++ b/code_review_graph/main.py
@@ -149,9 +149,7 @@ async def run_postprocess_tool(
     """
     return await asyncio.to_thread(
         run_postprocess,
-        flows=flows,
-        communities=communities,
-        fts=fts,
+        flows=flows, communities=communities, fts=fts,
         repo_root=_resolve_repo_root(repo_root),
     )
 
@@ -202,11 +200,8 @@ def get_impact_radius_tool(
         detail_level: "standard" for full output, "minimal" for compact summary. Default: standard.
     """
     return get_impact_radius(
-        changed_files=changed_files,
-        max_depth=max_depth,
-        repo_root=_resolve_repo_root(repo_root),
-        base=base,
-        detail_level=detail_level,
+        changed_files=changed_files, max_depth=max_depth,
+        repo_root=_resolve_repo_root(repo_root), base=base, detail_level=detail_level,
     )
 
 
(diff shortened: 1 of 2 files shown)


tirth8205:
I did not merge this into v2.3.4.

The release now includes FastMCP prompt compatibility fixes and CI has passed against the current FastMCP dependency floor. This PR is older and needs a rebase against main, plus a clear note on what compatibility gap still remains after v2.3.4.

tirth8205:
Thanks @sdeonvacation — and apologies for the silent overlap. You diagnosed the FastMCP 3.x `_tool_manager` breakage correctly and early, and splitting it out of #394 was the right call.

This is now superseded: main commit fd901ca (via #446, applying #365) rewrote `_apply_tool_filter` to enumerate tools with the public `mcp.list_tools()` API (with a running-event-loop fallback) and remove them via `mcp.local_provider.remove_tool()` — the same removal call you chose. The one substantive difference: main avoids reading `local_provider._components` and parsing its `tool:name@version` keys, which are private and may shift between FastMCP releases.

Verified on current main with fastmcp 3.2.4: all 18 tests in `tests/test_main.py` pass, including the six `TestApplyToolFilter` cases, and ruff is clean. Since this branch now conflicts in both files and adds nothing main lacks, I'm closing it as superseded. #394 stays open and will be reviewed on its own merits. Thanks again for the sharp catch — sorry a parallel fix shipped without flagging it here.

PR #398, opened 2026-04-28
feat: add SQL language support to parser

Aabhas-rath:
Adds .sql file parsing via tree-sitter-sql (already in language pack):
- CREATE TABLE/VIEW → Class nodes with extra["sql_kind"]
- CREATE FUNCTION → Function nodes with extra["sql_kind"]="function"
- CREATE PROCEDURE → regex fallback (grammar emits ERROR for this node)
- FROM/JOIN table references → IMPORTS_FROM edges for impact analysis

Includes fixture (tests/fixtures/sample.sql) and 8 TestSQLParsing tests.

### DIFF
--- a/code_review_graph/parser.py
+++ b/code_review_graph/parser.py
@@ -32,6 +32,15 @@ _SQL_TABLE_RE = re.compile(
     re.IGNORECASE,
 )
 
+# SQL keywords that can appear after FROM/JOIN but are NOT table names.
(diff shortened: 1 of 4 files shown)


lotum-david-j:
Was looking through past issues and PRs for this - seems like SQL support is already in the main branch. Just wanted to point out that the documentation is not updated - in case helpful i could do that on the main README.
Thanks for this interesting project !

PR #451, opened 2026-05-08
fix: d3.select("svg") selects legend inline SVG instead of main canvas

### LINKED ISSUE #450: Bug: d3.select("svg") selects legend inline SVG instead of main canvas (#graph-svg)
## Bug Description

In the generated `graph.html`, the `<nav id="legend">` section uses inline `<svg>` elements (16×16 icons) for node type legends. The main canvas `<svg id="graph-svg">` appears later in the DOM. The JavaScript code uses `d3.select("svg")` which selects the **first** `<svg>` element in the DOM — the small 16×16 legend icon — causing the entire graph to be rendered inside that tiny icon.

## Steps to Reproduce

1. Run `code-review-graph visualize` to generate `graph.html`
2. Open `graph.html` in a browser
3. The graph is not visible / rendered inside the legend icon instead of the main canvas

## Root Cause

In `code_review_graph/visualization.py`, there are two occurrences of:

```javascript
var svg = d3.select("svg").attr("viewBox", [0, 0, W, H]);
```

- Line 859
- Line 1759

Since the legend inline SVGs appear before `<svg id="graph-svg">` in the DOM, `d3.select("svg")` matches the wrong element.

## Suggested Fix

Change both occurrences from:

```javascript
var svg = d3.select("svg").attr("viewBox", [0, 0, W, H]);
```

to:

```javascript
var svg = d3.select("#graph-svg").attr("viewBox", [0, 0, W, H]);
```

This selects the main canvas SVG by its existing `id="graph-svg"` attribute, which is already defined at line 741:

```html
<svg id="graph-svg" tabindex="-1" role="img" aria-label="Interactive code knowledge graph visualization. Use search to find nodes, click files to expand."></svg>
```

## Environment

- code-review-graph version: latest (pipx installed)
- Python 3.13
- Browser: any

wangpi26:
## Problem

In the generated `graph.html`, the `<nav id="legend">` section uses inline `<svg>` elements (16×16 icons) for node type legends. The main canvas `<svg id="graph-svg">` appears later in the DOM. The JavaScript code uses `d3.select("svg")` which selects the **first** `<svg>` element in the DOM — the small legend icon — causing the entire graph to be rendered inside that tiny icon.

## Fix

Replace `d3.select("svg")` with `d3.select("#graph-svg")` at both occurrences in `code_review_graph/visualization.py`:
- Line 859
- Line 1759

The main canvas SVG already has `id="graph-svg"` defined, so this is a minimal, safe change.

## Testing

1. Run `code-review-graph visualize` to generate `graph.html`
2. Open `graph.html` in a browser
3. Graph now renders correctly in the main canvas area

Closes #450

### DIFF
--- a/code_review_graph/visualization.py
+++ b/code_review_graph/visualization.py
@@ -856,7 +856,7 @@ function moveTooltip(ev) {
 }
 function hideTooltip() { tooltip.classList.remove("visible"); }
 var W = innerWidth, H = innerHeight;
-var svg = d3.select("svg").attr("viewBox", [0, 0, W, H]);
+var svg = d3.select("#graph-svg").attr("viewBox", [0, 0, W, H]);
 var gRoot = svg.append("g");
 var currentTransform = d3.zoomIdentity;
 var zoomBehavior = d3.zoom()
@@ -1756,7 +1756,7 @@ function hideTooltip() { tooltip.classList.remove("visible"); }
 
 /* --- SVG setup --- */
 var W = innerWidth, H = innerHeight;
-var svg = d3.select("svg").attr("viewBox", [0, 0, W, H]);
+var svg = d3.select("#graph-svg").attr("viewBox", [0, 0, W, H]);
 var gRoot = svg.append("g");
 var currentTransform = d3.zoomIdentity;
 var zoomBehavior = d3.zoom()


tirth8205:
This appears to cover the same visualisation class of bug as the later graph-canvas rendering PR. I did not include it in v2.3.4 because browser-side rendering changes need visual verification before release.

The issue is still worth fixing, but it should come with a small test plan covering the generated graph page and the affected rendering modes.

tirth8205:
Thanks @wangpi26 — you diagnosed #450 first and correctly: the legend's inline 16x16 SVGs precede the canvas, so d3.select("svg") grabs a legend icon. Your change at line 859 (full-mode template) is exactly right.

The second change (line 1759) is in the aggregated template used by community/file modes, and there it regresses: that template's `<svg>` has no id="graph-svg" (its legend uses spans, so the bare selector was already correct). After this patch, d3.select("#graph-svg") returns an empty selection and those views render nothing — I verified by generating graph.html in all three modes.

PR #477 covers the same root cause, adds the missing id to the aggregated template, and fixes the related initial-sizing issue, so I'm going to close this in favor of it. Apologies for the silent overlap — yours was filed first and the diagnosis credit is yours. If you'd like to contribute the missing coverage, a regression test asserting every render mode's HTML contains id="graph-svg" and a matching selector would be very welcome.

PR #453, opened 2026-05-08
feat: add NPM support for global installation and Node.js ecosystem compatibility

reizhafajrian:
### Overview
This PR adds NPM support to the project, allowing users to install `code-review-graph` globally via `npm install -g code-review-graph`. 

While the core logic remains in Python, this addition provides a seamless experience for developers in the Node.js ecosystem, which is common for many AI coding tools (like Claude Code, Cursor, etc.).

### Key Changes
- **Node.js Wrapper:** Added `bin/cli.js` to act as a bridge, automatically locating the Python environment and forwarding commands.
- **Automated Setup:** Included a `scripts/npm-postinstall.js` script that automatically creates an isolated Python virtual environment (`.venv`) and installs all necessary dependencies upon running `npm install`.
- **Packaging:** Configured `package.json` with binary mappings (`code-review-graph` and `crg`) and ensured all required Python assets are bundled.
- **Documentation:** Updated `README.md` and all translations (`zh-CN`, `ja-JP`, `ko-KR`, `hi-IN`) to include the new NPM installation method.

### Why this is beneficial
- **Easier Installation:** Users don't need to manually manage Python virtual environments or paths if they prefer using `npm`.
- **Broader Adoption:** Makes the tool more accessible to frontend and full-stack developers who primarily use Node.js tools.
- **Improved DX:** The post-install script handles the heavy lifting of dependency management automatically.

### How to test
1. Clone this branch/fork.
2. Run `npm install -g .` from the root directory.
3. Verify that `code-review-graph --version` or `crg status` works correctly in any directory.


tirth8205:
Thanks for the packaging work. I did not include this in v2.3.4.

The release was kept away from broad platform and packaging expansion. NPM support may still be worth discussing, but it needs its own release decision and validation path because it changes how users install and run the project.

tirth8205:
Thanks — an npm launcher would make installation easier, but this wrapper cannot merge yet. It assumes a portable Python command/environment that npm does not guarantee, uses shell execution, does not fail installation when the Python package fails, and does not reliably forward signals/exit codes.

Please remove shell execution, implement platform-specific Python discovery or ship a self-contained artifact, fail postinstall correctly, forward signals and exit status, automate npm/Python version synchronization and release checks, remove unrelated .gemini changes, and add Windows/macOS/Linux installation and invocation tests.

tirth8205:
Thanks for the packaging work. I am closing head `d4526df` because it cannot merge into current main and remains 268 commits behind. The exact harness reproduces release-blocking lifecycle failures: without `.venv` the launcher resolves back to the npm shim, failed postinstall exits successfully, Python 3.9 passes the stated 3.10+ check, `shell: true` creates platform quoting risk, and the `child.connected` guard prevents normal SIGTERM forwarding. Only GitGuardian ran on this head.

A future npm distribution is welcome as a new PR from current main once it provides deterministic Python discovery or a self-contained runtime, argument-safe no-shell execution, strict version/error/signal/exit handling, synchronized npm/Python release automation, and Windows/macOS/Linux install, upgrade, offline-failure, and uninstall tests. Closing this implementation does not reject npm support as an idea.

PR #458, opened 2026-05-10
feat: surface graph staleness, omitted counts, and symbol disambiguation

jaissebastian:
Tool responses now carry the information agents need to self-correct without extra round-trips:

graph_meta block — every query_graph, semantic_search_nodes, and get_impact_radius response includes indexed_at (ISO timestamp), indexed_commit / head_commit (8-char SHAs from the graph DB and current HEAD), and is_stale so an agent knows whether to re-index before trusting callers_of results.

Omitted counts — query_graph gains max_results (default 100) and always returns results_omitted so agents don't mistake a cap for "no callers found". Minimal mode reports the same. semantic_search_nodes and get_impact_radius minimal modes add their own omitted counts.

Symbol disambiguation — when a bare name matches multiple nodes, the response now returns a ranked disambiguation list (exact QN > exact name
> partial match) with qualified_name, file_path, and line_start for each
candidate, plus a hint field explaining the fix. Replaces the old unranked candidates key.

29 tests in tests/test_agent_transparency.py covering all three features.

### DIFF
--- a/code_review_graph/tools/_common.py
+++ b/code_review_graph/tools/_common.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import subprocess
 from pathlib import Path
 from typing import Any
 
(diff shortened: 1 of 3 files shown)


tirth8205:
Thanks @jaissebastian — this is exactly the transparency layer the graph needed. The staleness design is right: stored git_head_sha vs live rev-parse HEAD catches the "pulled, never re-indexed" case we'd flagged internally. I merged it locally against main: your 33 new tests pass and the full suite is green (1307 passed); mypy and ruff are clean on the source files.

To get it merged:
1. Rebase on main — _common.py now has _resolve_graph_file_paths where you added graph_meta (keep both), and your resolution cascade needs to nest inside the new `if pattern != "file_summary":` guard; impact-radius minimal mode now uses attach_context_savings.
2. Narrow `except Exception` in graph_meta to `(subprocess.TimeoutExpired, OSError)` and log a warning — house convention.
3. `ruff check --fix tests/test_agent_transparency.py` (unused pytest import, import order).

One note: `update` still stamps git_head_sha = HEAD even when its HEAD~1 base misses earlier commits, so is_stale can under-report after multi-commit pulls — I'll file that as a follow-up; out of scope here. We'll flag the candidates→disambiguation rename in release notes.

tirth8205:
Thanks @jaissebastian — the transparency layer is valuable, and the old-base CI is green. The current head cb08a35 conflicts with main and needs these corrections before merge:

1. Make graph_meta Windows-safe with stdin=DEVNULL, catch only expected subprocess errors, and log failures. Include dirty/staged/untracked state, or rename the field so is_stale=false does not claim a dirty graph is fresh.
2. Validate max_results >= 1, apply bounds before materializing full result sets, keep edges consistent with retained results, and fix omitted-count math when max_results < 5.
3. Keep candidates as a compatibility alias.
4. Do not let a dotted Java-like target silently resolve to an unrelated global bare name; require matching language/class/file evidence or return ambiguity.

Please rebase on current main, preserve _resolve_graph_file_paths, the file_summary guard, and context-savings behavior, add focused tests for these cases, and rerun the full CI matrix.

PR #459, opened 2026-05-10
fix: skip unavailable tree-sitter parsers

letsjo:
## Summary
- lazily load `tree_sitter_language_pack` parsers instead of importing the package at module import time
- probe parser loading in a short child process before loading it in the main graph build
- mark only the failing language unavailable when a parser load hangs/fails, allowing the rest of the repository graph build to continue

## Why
On macOS with `tree-sitter-language-pack==0.13.0`, loading the TSX binding can hang inside `tree_sitter_language_pack.get_parser("tsx")`. Because parser loading previously happened in the main process, one broken language binding could block the entire graph build until an outer timeout killed it.

This change makes that failure language-scoped: the affected language is skipped, while all other languages still contribute graph context.

## Verification
- `uv run python -X faulthandler -m pytest tests/test_parser.py::TestCodeParser::test_parser_probe_timeout_marks_language_unavailable -q -p no:asyncio`
- `uv run python -m py_compile code_review_graph/parser.py tests/test_parser.py`
- `git diff --check`
- local smoke against a mixed repo where the installed CLI hangs on TSX:
  - baseline installed `code-review-graph build --skip-flows`: timed out after 10s
  - patched source via `PYTHONPATH` with `CRG_PARSER_LOAD_TIMEOUT_SECONDS=1`: completed in ~1.9s, produced 640 nodes / 6555 edges while skipping unavailable `tsx`

Note: `ruff check` currently hangs in my local dev environment even with `--no-cache`; no lint output was produced.

### DIFF
--- a/code_review_graph/parser.py
+++ b/code_review_graph/parser.py
@@ -9,12 +9,13 @@ from __future__ import annotations
 import hashlib
 import json
 import logging
+import os
 import re
+import subprocess
+import sys
 from dataclasses import dataclass, field
 from pathlib import Path
 from typing import NamedTuple, Optional
-
-import tree_sitter_language_pack as tslp
 
 from .tsconfig_resolver import TsconfigResolver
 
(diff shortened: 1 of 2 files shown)


tirth8205:
I left this out of v2.3.4.

The goal is reasonable, but parser availability and parser start-up behaviour are risky areas. This needs a smaller failure-mode description and focused tests showing which unavailable grammars are skipped and which parser failures should still surface. It is better suited to a parser reliability pass than to this release.

tirth8205:
Thanks—skipping only the broken grammar is the right goal. One blocking performance issue remains: every fresh `CodeParser` can start another Python subprocess probe, while the parallel path creates parsers per file. Please cache both success and failure per language for the whole process so each language is probed at most once. Add a regression test using several `CodeParser` instances that proves only one probe runs, plus focused tests showing timeout and non-zero probe failures skip only the bad grammar while unexpected parser errors still surface. Then rebase on main.

PR #462, opened 2026-05-10
Feat/webflux entrypoints test linkage config

jaissebastian:
## Summary

This PR adds comprehensive Spring Boot and Spring WebFlux support to the code graph, addressing the core gap where `callers_of` returned 0 for field-injected Spring service calls.

### Spring DI Resolution (two-pass)
- **Two-pass resolver** (`spring_resolver.py`): Pass 1 collects all field/type declarations across Java files; Pass 2 rewrites CALLS edges using the type map — resolves `@Autowired`, `@RequiredArgsConstructor` (Lombok), and explicit constructor injection
- **`@Bean` parameter injection**: `@Bean` factory method parameters now emit `INJECTS` edges so method-reference receivers (`handler::method`) resolve to their declared type
- **Confidence scoring**: edges tagged `EXTRACTED` with confidence 0.95 (constructor), 0.70 (autowired), 0.60 (constant reference)

### HTTP Endpoint Indexing
- **Annotation-based routing**: `@GetMapping`, `@PostMapping`, `@PutMapping`, `@DeleteMapping`, `@PatchMapping`, `@RequestMapping` on `@RestController` methods emit `Endpoint` nodes (`http:GET:/path`) and `HANDLES` edges
- **WebFlux functional routing**: `route().GET("/path", handler::method)` chains now emit the same Endpoint nodes — fixed `_get_call_name` to return the method name for chained Java method invocations (was returning `None`, silently skipping the entire call block)

### Config File Indexing
- **YAML / `.properties` parsing**: `.yml`, `.yaml`, `.properties` files are now parsed into `ConfigProperty` nodes
- **Config key normalization**: kebab-case (`gateway-url`) and camelCase (`gatewayUrl`) and `@Value("${payment.gatewayUrl}")` all normalize to the same key so edges resolve
- **`@Value` and `@ConfigurationProperties`** emit `DEPENDS_ON_CONFIG` edges

### Graph Traversal Fixes (BFS edge bridging)
- **`query_graph consumers_of`**: new pattern — "which classes/methods read this config property?" Works even without a `ConfigProperty` node (only `@Value` annotations parsed)
- **`traverse_graph` config-aware BFS**: `ConfigProperty` nodes now have BFS neighbours; `config:key` edge targets resolve to actual nodes
- **`traverse_graph` Endpoint-aware BFS**: `traverse_graph("POST /bulk-calculation")` now walks `HANDLES` edges to the handler method (depth 1) rather than returning only the Endpoint node. Same bridge pattern as config: `get_edges_by_endpoint_key("http:POST:/path")` maps the virtual key to real node QNs
- **`get_edges_by_config_key` / `get_edges_by_endpoint_key`**: new `GraphStore` methods that bridge virtual edge targets to graph nodes

### Java Parser Improvements
- **`method_reference`** (`handler::process`) added to `_CALL_TYPES` — emits CALLS edges with receiver stored for DI resolution
- **`object_creation_expression`** (`new Foo()`) now correctly extracts the type name from `type_identifier` child (was returning `None`)
- **Chained method invocation** call name extraction — `route().GET(...)` now returns `"GET"` instead of `None`

### Execution Flow Detection (`flows.py`)
- Entry point patterns: `@RestController`, `@KafkaListener`, `@WorkflowMethod`, `@ActivityMethod`, `@Scheduled`
- WebFlux `RouterFunction` return type detected as entry point
- Decorators stored in node `extra` dict so flow detection can read them

### Token Efficiency (from merged branches)
- `detail_level` parameter on all major tools
- Review bundle tool
- Graph staleness indicators, omitted counts, symbol disambiguation
- Java FQN resolution in `query_graph`

## Test plan

- [ ] Run `uv run pytest tests/ --tb=short -q` — 1269 tests pass (6 pre-existing `test_main.py::TestApplyToolFilter` failures unrelated to this PR)
- [ ] `uv tool run ruff check code_review_graph/` — clean
- [ ] Build against a Spring Boot project and verify `callers_of(ServiceA.fieldInjectedMethod)` returns callers with `resolution="spring_di_*"`
- [ ] `traverse_graph("POST /my-endpoint", depth=3)` reaches the handler method
- [ ] `query_graph("consumers_of", "some.config.key")` returns the Java class that reads it
- [ ] New fixture files: `tests/fixtures/SampleJava.java`, `tests/fixtures/SpringDI.java`, `tests/fixtures/application.yml`, `tests/fixtures/app.properties`

### DIFF
--- a/code_review_graph/flows.py
+++ b/code_review_graph/flows.py
@@ -46,9 +46,16 @@ _FRAMEWORK_DECORATOR_PATTERNS: list[re.Pattern[str]] = [
     re.compile(r"(override_settings|modify_settings)", re.IGNORECASE),
     # SQLAlchemy / event systems
     re.compile(r"(event\.)?listens_for", re.IGNORECASE),
-    # Java Spring
+    # Java Spring MVC / WebFlux annotation-based
     re.compile(r"(Get|Post|Put|Delete|Patch|RequestMapping)Mapping", re.IGNORECASE),
     re.compile(r"(Scheduled|EventListener|Bean|Configuration)", re.IGNORECASE),
+    re.compile(r"RestController", re.IGNORECASE),
+    # Spring Kafka / Temporal entry points
+    re.compile(r"KafkaListener", re.IGNORECASE),
+    re.compile(r"(WorkflowMethod|ActivityMethod)", re.IGNORECASE),
+    # Spring WebFlux functional routing (RouterFunction / route().GET())
+    re.compile(r"RouterFunction", re.IGNORECASE),
+    re.compile(r"HandlerFunction", re.IGNORECASE),
     # JS/TS frameworks
     re.compile(r"(Component|Injectable|Controller|Module|Guard|Pipe)", re.IGNORECASE),
     re.compile(r"(Subscribe|Mutation|Query|Resolver)", re.IGNORECASE),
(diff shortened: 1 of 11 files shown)


tirth8205:
Thanks — the WebFlux, endpoint, and configuration ideas are valuable, but this needs a rebase and narrower scope because Spring dependency resolution is already on main.

The current configuration parser stores raw YAML/property values in node metadata, which can expose passwords or tokens. It also treats every YAML file as Spring configuration, including GitHub Actions, Kubernetes, and Ansible files. Endpoint extraction misses controller-level prefixes and important @RequestMapping forms, while @Bean parameter injection can leak across method scope.

Please remove or redact raw values, introduce a shared YAML classifier with #415, correct route composition and method handling, keep injection method-scoped, add security, duplicate-route, non-Spring-YAML, and scope regressions, then run full CI.

PR #477, opened 2026-05-13
fix(visualization): graph renders in legend icon instead of main canvas

vaguemit:
## Problem

When opening the generated graph.html in a browser, the entire D3.js force-directed graph (all nodes, edges, zoom behavior, simulation) was rendering inside a tiny 16x16 pixel SVG icon in the legend sidebar under the 'NODES' heading, rather than on the full-viewport canvas.

### Root Cause 1 — Wrong SVG element selected (primary bug)

Both _HTML_TEMPLATE and _AGGREGATED_HTML_TEMPLATE used:

    var svg = d3.select('svg')         // full mode
    var svgEl = document.querySelector('svg')  // aggregated mode

d3.select('svg') and document.querySelector('svg') select the FIRST <svg> element in the DOM. The legend contains inline SVG icons for each node shape (circle, square, triangle, diamond, cross) that appear before the main <svg id='graph-svg'> canvas in document order. D3 therefore appended all graph content into the first 16x16 legend icon.

### Root Cause 2 — Dimensions captured as zero at parse time

On file:// URLs, the SVG element has not been laid out by the browser when the inline <script> executes. This causes svgEl.clientWidth and svgEl.clientHeight to return 0, so the force simulation centers every node at (0, 0) via:

    .force('center', d3.forceCenter(W/2, H/2))  // -> (0, 0)
    .force('x', d3.forceX(W/2))                  // -> 0
    .force('y', d3.forceY(H/2))                  // -> 0

After the simulation settles all nodes are piled at the origin, and fitGraph() finds a zero-sized bounding box and silently returns.

### 1. Use ID selectors throughout (fixes primary bug)

- Added id='graph-svg' to the <svg> element in _AGGREGATED_HTML_TEMPLATE (it was already present in _HTML_TEMPLATE)
- Changed all selectors from d3.select('svg') / document.querySelector('svg') to d3.select('#graph-svg') / document.getElementById('graph-svg') in both templates

### 2. Re-center nodes when simulation ends with correct dimensions

In simulation.on('end') for both templates:
- Re-read W/H using getW()/getH() (which query svgEl.clientWidth/Height)
- If dimensions changed significantly from initial capture (or were < 100px), translate all node x/y positions by the delta so the cluster moves to the true viewport center
- Reset forceCenter, forceX, forceY with corrected W/H values
- Update the SVG viewBox

### 3. Robust fitGraph() with requestAnimationFrame retry

Replaced the silent early-return on zero BBox with a retry loop:

    function fitGraph(retries) {
      if (retries === undefined) retries = 10;
      var b = gRoot.node().getBBox();
      if (b.width === 0 || b.height === 0) {
        if (retries > 0) requestAnimationFrame(function() { fitGraph(retries - 1); });
        return;
      }
      // ... compute transform and apply ...
    }

Uses requestAnimationFrame instead of setTimeout for accurate post-paint timing on both file:// and http:// origins.

### 4. Live dimension helpers (getW / getH)

Both templates already had getW()/getH() helpers querying svgEl.clientWidth/clientHeight with window.innerWidth/Height fallback. These are now consistently used everywhere dimensions are needed, including the resize event listener.

## Files changed

- code_review_graph/visualization.py: _HTML_TEMPLATE and _AGGREGATED_HTML_TEMPLATE JS sections updated

### DIFF
--- a/code_review_graph/visualization.py
+++ b/code_review_graph/visualization.py
@@ -855,8 +855,11 @@ function moveTooltip(ev) {
   tooltip.style.left = x + "px"; tooltip.style.top = y + "px";
 }
 function hideTooltip() { tooltip.classList.remove("visible"); }
-var W = innerWidth, H = innerHeight;
-var svg = d3.select("svg").attr("viewBox", [0, 0, W, H]);
+var svgEl = document.getElementById("graph-svg");
+function getW() { return svgEl.clientWidth || window.innerWidth || 800; }
+function getH() { return svgEl.clientHeight || window.innerHeight || 600; }
+var W = getW(), H = getH();
+var svg = d3.select("#graph-svg").attr("viewBox", [0, 0, W, H]);
 var gRoot = svg.append("g");
 var currentTransform = d3.zoomIdentity;
 var zoomBehavior = d3.zoom()
@@ -1071,13 +1074,18 @@ if (N > 2000) {
   nodes.forEach(function(n) { if (n.kind === "File") collapsedFiles.add(n.qualified_name); });
 }
 updateNodes();
-function fitGraph() {
+function fitGraph(retries) {
+  if (retries === undefined) retries = 10;
   var b = gRoot.node().getBBox();
-  if (b.width === 0 || b.height === 0) return;
+  if (b.width === 0 || b.height === 0) {
+    if (retries > 0) requestAnimationFrame(function() { fitGraph(retries - 1); });
+    return;
+  }
   var pad = 0.1;
   var fw = b.width * (1 + 2*pad), fh = b.height * (1 + 2*pad);
-  var s = Math.min(W / fw, H / fh, 2.5);
-  var tx = W/2 - (b.x + b.width/2)*s, ty = H/2 - (b.y + b.height/2)*s;
+  var cw = getW(), ch = getH();
+  var s = Math.min(cw / fw, ch / fh, 2.5);
+  var tx = cw/2 - (b.x + b.width/2)*s, ty = ch/2 - (b.y + b.height/2)*s;
   svg.transition().duration(600).call(zoomBehavior.transform, d3.zoomIdentity.translate(tx, ty).scale(s));
 }
 var loadingOverlay = document.getElementById("loading-overlay");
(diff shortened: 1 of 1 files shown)


tirth8205:
I agree this looks like a real visualisation bug, but I did not include it in v2.3.4.

The patch changes embedded browser-side graph rendering behaviour, and I did not want to ship that without browser verification across the generated visualisation modes. The release was centred on MCP and CLI review reliability, so this should be revisited with a small visual test plan rather than folded into the release after the fact.

tirth8205:
Thanks @vaguemit — this is an excellent diagnosis. You caught both root causes (first-`<svg>` selector grabbing the legend icon, and zero clientWidth/Height on file:// at parse time) and fixed both template sites, including the missing id="graph-svg" on the aggregated canvas. I applied the patch, ran the full suite (1274 passed), and confirmed full-mode HTML now mounts via #graph-svg.

One blocker: the resize-listener hunk in _AGGREGATED_HTML_TEMPLATE uses escaped quotes (\"resize\"). Both templates are raw strings (r"""), so those backslashes land verbatim in the generated JS — node --check confirms a SyntaxError, which kills the whole script in community/file/auto modes. That's a worse regression there than the bug itself.

To get this merged:
1. Replace the \" with plain " in that hunk (visualization.py, aggregated template resize listener).
2. Add regression tests in tests/test_visualization.py: for both full and aggregated output, assert d3.select("#graph-svg") is present, d3.select("svg") is absent, and no literal \" appears.

Heads-up: #451 overlaps on the selector fix; we'll reconcile that on our side.

tirth8205:
Updated maintainer disposition: this conflicting branch is no longer waiting on contributor changes. I am porting the safe visualization fix onto current main with Mit Parikh attribution.\n\nThe source branch still cannot merge directly because its aggregated raw JavaScript template emits a literal backslash-quote in the resize listener, which fails JavaScript parsing and disables community/file/auto visualization modes. The current templates have also moved since this branch was cut.\n\nThe replacement will retain:\n- graph-svg ID selection in both full and aggregated modes;\n- live viewport dimensions;\n- post-layout recentering;\n- bounded requestAnimationFrame retry when the initial bounding box is zero.\n\nRequired verification:\n- generated full and aggregated JavaScript parses;\n- neither mode selects the first generic SVG legend icon;\n- unrelated legend SVGs remain untouched;\n- browser smoke tests cover file URL layout, resize, and zero-size retry;\n- full local and protected CI suites pass.\n\nI will post the replacement PR, browser/test evidence, merge commit, and attribution here, then close this superseded source branch.

vaguemit:
Thanks for the thorough review and for catching the raw-string escaping issue, that's a real bug and I appreciate the node --check confirmation.

Given the drift from main since I cut this branch, reimplementing on current main makes sense rather than forcing the merge. I'm fine with that approach and appreciate the attribution.

One ask: before you merge, could you share the diff or draft PR so I can confirm the four preserved behaviors (graph-svg ID selection in both modes, live viewport dimensions, post-layout recentering, bounded rAF retry) match what I intended, particularly around the zero-bounding-box retry logic? I'd also like to see the regression tests you're adding for #graph-svg presence, svg absence, and the literal backslash-quote check.

I'll close out this branch once the replacement PR is up and verified.

tirth8205:
Final maintainer disposition: the safe, non-overlapping part of this contribution has now merged through #642 in commit f2b9fb145504e7c598a3963b18f43b9c08c03d68, with Mit Parikh retained as co-author.\n\nWhat was retained and fixed:\n- graph sizing now reads the live SVG canvas instead of relying only on the initial viewport;\n- resize/end events resynchronize the viewport in both full and aggregated modes;\n- fitGraph uses a bounded requestAnimationFrame retry when getBBox is temporarily zero;\n- fit-button callbacks no longer pass a DOM event as the retry counter;\n- the aggregated-script quote escaping from this patch was corrected so the generated JavaScript parses.\n\nOverlap disposition:\n- the incorrect SVG selector reported here was already fixed independently by #564, so that duplicated portion was not copied or credited twice.\n\nVerification:\n- 24 focused visualization tests passed;\n- generated inline JavaScript is syntax-checked in both rendering modes;\n- all 11 required GitHub checks passed, including Linux on Python 3.10 through 3.13, type-check, lint, security, schema sync, review, and Windows daemon/file-handle coverage.\n\nThis source PR is therefore superseded by the tested replacement and can be closed without losing its valid work.

PR #483, opened 2026-05-15
fix: prompt Message wrapping + add detail_level=minimal to get_architecture_overview

selika:
## Summary

Two independent fixes, both surfaced from the same MCP slash-command failure (`/code-review-graph:architecture_map`):

### 1. fastmcp ≥3.2 prompt Message wrapping (regression fix)

fastmcp 3.2.4 stopped accepting raw `dict` messages in prompt return values:

```
McpError: messages[0] must be Message or str, got dict.
Use Message({'role': 'user', 'content': '...'}) to wrap the value.
```

This breaks **all 5 MCP prompt templates** (`review_changes`, `architecture_map`, `debug_issue`, `onboard_developer`, `pre_merge_check`) since they return `list[dict]`.

Fix: wrap each message via `fastmcp.prompts.prompt.Message` so role is explicit and future multi-turn prompts compose naturally. Added a tiny `_user()` helper to keep the five templates uniform.

### 2. `get_architecture_overview` `detail_level="minimal"` (token-efficiency)

The default output embeds every community's full `members` list *and* every individual cross-community edge. On a 2,878-node repo (`ebm-copilot` — TS/Python/Bash/R, 390 files) this measured **655KB** — over the MCP tool output token cap, forcing fallback-to-file and defeating the whole point of the token-efficient API.

The token-efficiency preamble already tells the LLM to call `get_architecture_overview(detail_level="minimal")` first, but `minimal` wasn't actually a supported value — so this PR finally makes the prompt's own contract work.

`minimal` mode:
- communities → keep only `{id, name, size, cohesion, dominant_language}`, drop `members`
- cross_community_edges → aggregate by `(source_community, target_community)` pair → `{source_community (name), target_community (name), edge_count, top_kinds}`
- warnings → unchanged

Measured on ebm-copilot:

| mode | size | reduction |
|------|------|-----------|
| standard | 655 KB | — |
| minimal | 5.4 KB | **120×** |

Output stays useful — the top pair `v1-route ↔ services-list` (318 `CALLS` edges) immediately surfaces the hottest coupling.

## Test plan

- [x] `uv run pytest tests/test_prompts.py` — 34 passed (updated to assert `msg.role` / `msg.content.text`)
- [x] `uv run pytest tests/test_tools.py` — 3 new cases for minimal mode (drops members, aggregates pairs, summary label)
- [x] `uv run pytest tests/test_integration_v2.py` — step 9 updated for Message attribute access
- [x] Full suite: 1233 passed, 1 skipped, 2 xpassed. The 6 remaining failures are pre-existing pytest-asyncio plugin issues in `tests/test_main.py::TestApplyToolFilter` (the test file uses `@pytest.mark.asyncio` but `pytest-asyncio` is not a dev dep — out of scope here).
- [x] `uvx ruff check` on changed files — clean.
- [x] End-to-end: `architecture_map` prompt now renders through fastmcp's in-process `Client` (verified the exact MCP path the slash command uses).
- [x] Smoke test on real `ebm-copilot` graph confirms the 120× reduction.

### DIFF
--- a/code_review_graph/main.py
+++ b/code_review_graph/main.py
@@ -549,6 +549,7 @@ def get_community_tool(
 @mcp.tool()
 def get_architecture_overview_tool(
     repo_root: Optional[str] = None,
+    detail_level: str = "standard",
 ) -> dict:
     """Generate an architecture overview based on community structure.
 
@@ -558,8 +559,15 @@ def get_architecture_overview_tool(
 
     Args:
         repo_root: Repository root path. Auto-detected if omitted.
+        detail_level: "standard" (default) returns full per-edge detail;
+                      "minimal" drops community member lists and
+                      aggregates cross-community edges to one row per
+                      community pair (typical reduction: 600KB → <5KB).
     """
-    return get_architecture_overview_func(repo_root=_resolve_repo_root(repo_root))
+    return get_architecture_overview_func(
+        repo_root=_resolve_repo_root(repo_root),
+        detail_level=detail_level,
+    )
 
 
 @mcp.tool()
--- a/code_review_graph/prompts.py
+++ b/code_review_graph/prompts.py
@@ -12,6 +12,8 @@ detail_level="minimal" first patterns with get_minimal_context entry point.
 
 from __future__ import annotations
 
+from fastmcp.prompts.prompt import Message
+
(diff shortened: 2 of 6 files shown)


tirth8205:

PR #486, opened 2026-05-15
fix: callers_of returns cross-file callers regardless of same-file caller presence

### LINKED ISSUE #472: query_graph callers_of misses cross-file callers when any same-file caller exists
`query_graph(pattern="callers_of", ...)` returns incomplete results for Objective-C and C codebases. When a function has both same-file callers (edges with fully-qualified `target_qualified`) and cross-file callers (edges with bare `target_qualified`), only the same-file callers are returned. The cross-file callers are silently dropped.

## Root Cause

In `code_review_graph/tools/query.py`, the `callers_of` branch (line ~226):

```python
if pattern == "callers_of":
    for e in store.get_edges_by_target(qn):
        if e.kind == "CALLS":
            caller = store.get_node(e.source_qualified)
            if caller:
                results.append(node_to_dict(caller))
            edges_out.append(edge_to_dict(e))
    # Fallback: search by plain name for unqualified targets
    if not results and node:                    # <--- BUG
        for e in store.search_edges_by_target_name(node.name):
            ...
```

The fallback `search_edges_by_target_name` only executes when `results` is empty. But in ObjC/C, the parser (`_resolve_call_target` in `parser.py`) generates:

- **Fully-qualified targets** (e.g. `file.m::ClassName.methodName`) when the callee is defined in the **same file** (present in `defined_names`)
- **Bare targets** (e.g. `methodName`) when the callee is defined in a **different file** (not in `defined_names`, not in `import_map`)

Since `get_edges_by_target(qn)` matches the qualified edges first, `results` becomes non-empty, and the fallback that would catch the bare-target cross-file edges is skipped.

## Impact

- **87.2% of CALLS edges** in the affected codebase have bare (unqualified) targets
- Any `callers_of` query that has ≥1 same-file caller will miss all cross-file callers
- This is especially severe for ObjC where Category methods are spread across many files — `[self methodDefinedInAnotherCategory]` always produces bare targets

## Reproduction

Given this ObjC codebase structure:

```objc
// ClawGame.m (Category 1 - defines the method)
@implementation KSKTVRoomViewController (ClawGame)
- (void)startLoadClawGameViewInfo { ... }     // line 31: DEFINITION
- (void)handleClawGameMessage:(KSIMMessage *)message {
    [self startLoadClawGameViewInfo];          // line 99: same-file call → qualified target
}
@end

// RoomControl.m (Category 2 - calls the method)
@implementation KSKTVRoomViewController (RoomControl)
- (void)handleDelayAfterEnterRoom {
    [self startLoadClawGameViewInfo];          // line 3009: cross-file call → bare target
}
@end
```

After `code-review-graph build`:

```python
# Query returns only 1 caller, should return 2
result = query_graph(pattern="callers_of", target="startLoadClawGameViewInfo")
assert len(result["results"]) == 1  # FAILS: expected 2
```

The database correctly contains both edges:

| edge_id | target_qualified | file | line |
|---------|-----------------|------|------|
| 109958 | `...ClawGame.m::KSKTVRoomViewController.startLoadClawGameViewInfo` | ClawGame.m | 99 |
| 136509 | `startLoadClawGameViewInfo` | RoomControl.m | 3009 |

But `query_graph` only returns edge 109958.

## Proposed Fix

Change the fallback condition from `if not results` to always execute and deduplicate:

```python
if pattern == "callers_of":
    seen_sources = set()
    for e in store.get_edges_by_target(qn):
        if e.kind == "CALLS":
            caller = store.get_node(e.source_qualified)
            if caller:
                results.append(node_to_dict(caller))
                seen_sources.add(e.source_qualified)
            edges_out.append(edge_to_dict(e))
    # Always search by plain name for unqualified targets, deduplicating
    if node:
        for e in store.search_edges_by_target_name(node.name):
            if e.source_qualified not in seen_sources:
                caller = store.get_node(e.source_qualified)
                if caller:
                    results.append(node_to_dict(caller))
                    seen_sources.add(e.source_qualified)
                edges_out.append(edge_to_dict(e))
```

The same fix should be applied to `callees_of` (which has the same structural issue when the callee's node isn't in the graph).

## Environment

- `code-review-graph` version: 2.3.3
- Language: Objective-C (also affects C based on the same `_resolve_call_target` logic)
- OS: macOS (arm64)
- Python: 3.14

mvanhorn:
## Summary

`query_graph callers_of <function>` silently dropped cross-file callers when any same-file caller existed for the target. The early-return at the same-file scan path didn't continue into the cross-file sweep, so the response was incomplete depending on the call graph shape rather than the user's query intent.

Fix the same-file scan to always feed into the cross-file aggregator (don't short-circuit), and add regression tests at the query layer that cover same-file-only, cross-file-only, and mixed scenarios.

## Why this matters

#472 (filed by @nicobailon) reports the missing callers in real codebases. Once you have multiple callers in the same file, the cross-file ones drop out of the tool result, which defeats the point of asking for the call graph in the first place. The fix is contained to `code_review_graph/tools/query.py`'s callers_of path; the SQL-side aggregation already produced the right result, the bug was in how the Python layer narrowed the iterator.

Tests: `uv run pytest tests/test_tools.py -k "callers_of or test_query"` -> 4 passed.

closes #472

AI was used for assistance.

### DIFF
--- a/code_review_graph/tools/query.py
+++ b/code_review_graph/tools/query.py
@@ -216,29 +216,43 @@ def query_graph(
         qn = node.qualified_name if node else target
 
         if pattern == "callers_of":
+            seen_sources: set[str] = set()
             for e in store.get_edges_by_target(qn):
                 if e.kind == "CALLS":
-                    caller = store.get_node(e.source_qualified)
-                    if caller:
-                        results.append(node_to_dict(caller))
-                    edges_out.append(edge_to_dict(e))
+                    if e.source_qualified not in seen_sources:
+                        seen_sources.add(e.source_qualified)
+                        caller = store.get_node(e.source_qualified)
+                        if caller:
+                            results.append(node_to_dict(caller))
+                        edges_out.append(edge_to_dict(e))
             # Fallback: CALLS edges store unqualified target names
             # (e.g. "generateTestCode") while qn is fully qualified
             # (e.g. "file.ts::generateTestCode"). Search by plain name too.
-            if not results and node:
+            if node:
                 for e in store.search_edges_by_target_name(node.name):
-                    caller = store.get_node(e.source_qualified)
-                    if caller:
-                        results.append(node_to_dict(caller))
-                    edges_out.append(edge_to_dict(e))
+                    if e.source_qualified not in seen_sources:
+                        seen_sources.add(e.source_qualified)
+                        caller = store.get_node(e.source_qualified)
+                        if caller:
+                            results.append(node_to_dict(caller))
+                        edges_out.append(edge_to_dict(e))
 
         elif pattern == "callees_of":
+            seen_targets: set[str] = set()
             for e in store.get_edges_by_source(qn):
                 if e.kind == "CALLS":
-                    callee = store.get_node(e.target_qualified)
-                    if callee:
-                        results.append(node_to_dict(callee))
-                    edges_out.append(edge_to_dict(e))
+                    if e.target_qualified not in seen_targets:
+                        seen_targets.add(e.target_qualified)
+                        callee = store.get_node(e.target_qualified)
+                        if callee:
+                            results.append(node_to_dict(callee))
+                        elif "::" not in e.target_qualified:
+                            results.append({
+                                "kind": "Function",
+                                "name": e.target_qualified,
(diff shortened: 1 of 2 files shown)


tirth8205:

PR #487, opened 2026-05-16
fix(cli): honor skip postprocess flags

iamstuffed:
## Summary

Fix the `build` and `update` CLI commands so their post-processing flags are honored.

Previously, the CLI passed the selected postprocess level into `build_or_update_graph(...)`, but then ran the older full CLI post-processing pipeline afterward. As a result:

- `code-review-graph build --skip-postprocess` still ran full post-processing.
- `code-review-graph update --skip-flows` still ran flow/community post-processing.
- CLI output could report `postprocess=none` or `postprocess=minimal` while doing more work than requested.

This PR removes the extra CLI post-processing pass and relies on `build_or_update_graph(...)` as the source of truth for `none`, `minimal`, and `full` post-processing behavior.

## Changes

- Remove the extra `_cli_post_process(...)` calls after `build_or_update_graph(...)`.
- Remove the now-unused `_cli_post_process` helper.
- Add CLI regression coverage for:
  - `build --skip-postprocess`
  - `update --skip-flows`
- Add `pytest-asyncio` to the `uv` dev dependency group so `uv run python -m pytest` installs the async test plugin used by the existing test suite.

## Validation

```bash
uv run python -m pytest
```

Result:

```text
1238 passed, 1 skipped, 2 xpassed in 16.79s
```

### DIFF
--- a/code_review_graph/cli.py
+++ b/code_review_graph/cli.py
@@ -322,21 +322,6 @@ def _handle_init(args: argparse.Namespace) -> None:
     print("  2. Restart your AI coding tool to pick up the new config")
 
 
-def _cli_post_process(store: GraphStore) -> None:
-    """Run post-build pipeline and print a summary line for each step."""
-    from .postprocessing import run_post_processing
-
-    pp = run_post_processing(store)
-    if pp.get("signatures_computed"):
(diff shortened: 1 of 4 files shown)


tirth8205:

Issue #385, opened 2026-04-24
Windows: semantic_search_nodes_tool hangs on stdio MCP transport (missing asyncio.to_thread + tokenizer deadlock)


gfleg77:
## Summary

On Windows, `semantic_search_nodes_tool` silently hangs indefinitely when sentence-transformers embeddings are available. Three layered issues prevent semantic search from working on Windows.

## Environment

- OS: Windows 10 Enterprise
- Python: 3.11
- code-review-graph: 2.3.2
- sentence-transformers: 3.4.1
- FastMCP: 2.14.7
- Transport: stdio (VS Code extension)

### 1. `semantic_search_nodes_tool` missing `asyncio.to_thread` (regression vs `embed_graph_tool`)

`embed_graph_tool` already uses `asyncio.to_thread` (per the fix in #46 / #136) to avoid blocking the `WindowsSelectorEventLoop` with sentence-transformers inference. `semantic_search_nodes_tool` does not — it is a plain `def` that calls `semantic_search_nodes()` synchronously on the event loop, producing the same deadlock.

**Fix:** convert to `async def` + `asyncio.to_thread`, identical to `embed_graph_tool`:

```python
@mcp.tool()
async def semantic_search_nodes_tool(
    query: str,
    ...
) -> dict:
    """..."""
    return await asyncio.to_thread(
        semantic_search_nodes,
        query=query, kind=kind, limit=limit,
        repo_root=_resolve_repo_root(repo_root),
        model=model, detail_level=detail_level,
    )
```

### 2. Rust tokenizer thread deadlock in subprocess context

Even with `asyncio.to_thread`, the Rust-based fast tokenizers inside sentence-transformers spawn parallel threads on first load. In a Windows MCP server subprocess this causes a deadlock. Required env vars (not currently documented):

```json
"TOKENIZERS_PARALLELISM": "false",
"OMP_NUM_THREADS": "1"
```

### 3. 12-second cold start on first tool call

The `LocalEmbeddingProvider` lazy-loads `SentenceTransformer` on the first inference call (~12s on first call, <1s subsequently). This means the first `semantic_search_nodes_tool` call in a session always blocks the event loop for 12 seconds even after fix #1.

**Fix:** pre-warm synchronously in `main()` before `mcp.run()` starts the event loop:

```python
def _prewarm_embeddings() -> None:
    try:
        from .embeddings import LocalEmbeddingProvider
        import os
[...]

PR #508, opened 2026-05-22
fix(windows): pre-warm sentence-transformers in main thread to prevent semantic_search deadlock (closes #507)

### LINKED ISSUE #507: semantic_search_nodes_tool still deadlocks on Windows at scale — asyncio.to_thread fix from #385 is insufficient on its own
### Symptoms

On Windows 11 + Python 3.13 + `code-review-graph` 2.3.3 + `fastmcp` 2.14.7, calling `semantic_search_nodes_tool` from Claude Code (stdio MCP) deadlocks indefinitely on the **first call** of a fresh MCP server.

- Process stays alive, RAM ≈ 390 MB (sentence-transformers loaded)
- CPU pinned at 2–5 seconds for 120+ seconds wall-clock
- No JSON-RPC response, no progress
- All other CRG tools (`list_graph_stats_tool`, `get_minimal_context_tool`, `query_graph_tool`, `get_architecture_overview_tool`, …) work normally — they return in < 0.2 s

### Empirical repro

Repo with ~8000 embedded nodes (`embeddings_count: 8058`). Cold MCP spawn → init → `semantic_search_nodes_tool({"query":"foo","limit":3})` → no response for 120 s, then killed.

Captured timing via subprocess stdio harness with per-3s progress trace:

```
+   1.55s  ✓ initialize in 0.04s
+   3.06s  >>> calling semantic_search_nodes_tool (COLD)
+   6.07s  trace t+ 3.0s  PID(cpu=2.8s, ram=386MB)   ← idle, model imports done
+   9.08s  trace t+ 6.0s  PID(cpu=2.8s, ram=386MB)
…
+ 120.63s  trace t+117.6s PID(cpu=2.8s, ram=386MB)   ← still idle
+ 123.13s  COLD TIMEOUT after 120s
```

CPU pinned at 2.8 s for the entire 120 s — the tool function never executes.

### Why the existing #385 fix is insufficient

#385 wrapped peer tools in `asyncio.to_thread`. That moves the work off the event loop but the worker thread is still inside fastmcp's executor pool. Torch's Windows DLL init + OpenMP thread-pool registration grab locks the loop also needs.

I locally tried converting `semantic_search_nodes_tool` itself to `async def` + `await asyncio.to_thread(...)` — CPU climbed slightly (2.8 → 5 s) then went idle again; deadlock persisted.

Also tried:
- `TOKENIZERS_PARALLELISM=false` → no effect (deadlock predates tokenizer use)
- `python -m code_review_graph` direct (vs `.exe` wrapper) → no effect, same Python process
- numpy-based `_cosine_similarity` rewrite → legitimate perf improvement at scale but does NOT fix the deadlock (the deadlock occurs before any cosine call)

### Root cause

Lazy-importing `sentence_transformers` + `torch` **inside an asyncio worker thread** on Windows deadlocks on torch's DLL load / OpenMP thread-pool registration.

When the model is pre-imported and pre-loaded **on the main thread before `mcp.run()` starts**, the dangerous init happens on the parent process. Worker threads only access an already-warm module afterwards, and the deadlock cannot manifest.

### Fix verified

Across 4 independent cold-start runs after applying a main-thread pre-warm:

| Run | Model load (main thread) | First cold `semantic_search` | Warm follow-up |
|---|---|---|---|
| 1 | 5.9 s | **0.31 s** | 0.218 s |
| 2 | 6.2 s | **0.32 s** | 0.244 s |
| 3 | 7.0 s | **0.36 s** | 0.396 s |
| 4 | 6.3 s | **0.30 s** | 0.234 s |

Cold tool call: **~0.3 s** (was: 120 s+ timeout).

PR with the fix incoming.

dannymn:
### Problem

`semantic_search_nodes_tool` deadlocks indefinitely on Windows stdio MCP at scale (8 k+ embeddings) even after #385's `asyncio.to_thread` fix on peer tools. The remaining hazard is lazy-importing `sentence_transformers` + `torch` **inside** a FastMCP executor worker thread — torch's Windows DLL init and OpenMP thread-pool registration grab locks the event loop also needs, freezing the call. Other CRG tools (returning in < 0.2 s) are unaffected because they don't load ML libraries on first use.

See #507 for full empirical repro (PID idle at CPU 2.8 s for 120 + s; timeout).

### Fix

A 60-line patch across two files: a module-level model cache + a `prewarm_local_embeddings()` helper that loads the configured local embedding model **on the main thread before `mcp.run()` starts**. Worker threads then only access the already-warm module, and the deadlock cannot manifest.

**`code_review_graph/embeddings.py`** (+ 51 lines):
- `_MODEL_CACHE: dict[str, Any]` — process-wide cache shared across `LocalEmbeddingProvider` instances
- `prewarm_local_embeddings(model_name=None)` — best-effort startup hook; no-op when `sentence-transformers` isn't installed (cloud-provider setups unaffected); idempotent
- `LocalEmbeddingProvider._get_model()` — consults the cache first, populates it on lazy load

**`code_review_graph/main.py`** (+ 9 lines):
- Inside the existing `if sys.platform == "win32":` guard (same place as the `WindowsSelectorEventLoopPolicy` fix already present), call `prewarm_local_embeddings()` right after the event-loop policy is set

### Verified

Windows 11 + Python 3.13 + CRG 2.3.3 + FastMCP 2.14.7, repo with 8 058 embedded nodes, captured via subprocess MCP stdio harness with per-3 s progress trace:

| | Before this patch | After this patch |
|---|---|---|
| Main-thread model load | — | ~ 6 s (one-time at boot) |
| First semantic_search **cold** | deadlock @ CPU 2.8 s for 120 + s → timeout | **0.30 s** |
| Subsequent semantic_search **warm** | — | **0.23 s** |

Reproduced across 4 independent cold MCP spawns. Cold-call latency: 0.30 / 0.31 / 0.32 / 0.36 s. No deadlock observed after patch.

### Alternatives considered

| Approach | Result |
|---|---|
| Wrap `semantic_search_nodes_tool` in `asyncio.to_thread` | CPU climbs (2.8 → 5 s) then idle again; deadlock persists — the worker thread is still in fastmcp's executor pool where torch DLL init still hangs |
| `TOKENIZERS_PARALLELISM=false` env var | No effect — the deadlock occurs before tokenizer use |
| numpy-based `_cosine_similarity` rewrite | Legitimate perf improvement at scale but **does NOT** fix the deadlock (the deadlock occurs before any cosine call). Happy to file as a separate perf PR if welcome. |

The dangerous initialization has to complete on the **main thread before any worker thread is spawned**. This patch ensures that.

### Compatibility

- **Cloud-only setups** (Google / MiniMax / OpenAI): the `ImportError` no-op branch fires; existing behaviour unchanged.
- **Non-Windows users**: the pre-warm sits inside the existing `sys.platform == "win32"` guard alongside the `WindowsSelectorEventLoopPolicy` workaround — zero behaviour change.
- **`embed_graph_tool`** and other already-asyncio.to_thread'd tools: still work the same; they just hit the cache on the second call instead of re-importing.

### Why now

Reproduced on a real project (DreamJob.AI) integrating CRG with Claude Code. After tracing the deadlock empirically (see #507 for the captured trace), the fix was small enough that it felt worth offering back. Same place as PR #425 — Windows + stdio MCP edge cases — should round out the Windows hardening story.

Happy to iterate on review.

### DIFF
--- a/code_review_graph/embeddings.py
+++ b/code_review_graph/embeddings.py
@@ -65,6 +65,50 @@ class EmbeddingProvider(ABC):
 LOCAL_DEFAULT_MODEL = "all-MiniLM-L6-v2"
 
 
+# Process-wide cache of loaded sentence-transformer models, keyed by model name.
+# Populated by ``prewarm_local_embeddings()`` at server startup (see ``main.main``)
+# and by ``LocalEmbeddingProvider._get_model`` on first lazy load. Sharing the
+# loaded model across ``LocalEmbeddingProvider`` instances avoids re-importing
+# ``sentence_transformers`` + ``torch`` from worker threads, which deadlocks
+# ``semantic_search_nodes_tool`` on Windows stdio MCP (#385 fixed the peer
+# tools via ``asyncio.to_thread``; this cache fixes the remaining case where
+# torch DLL / OpenMP init runs inside an executor thread).
+_MODEL_CACHE: dict[str, Any] = {}
+
+
+def prewarm_local_embeddings(model_name: str | None = None) -> None:
+    """Eagerly load the local sentence-transformer model on the calling thread.
+
+    Call this from the **main thread** before entering an asyncio event loop
+    (e.g. before ``mcp.run()``) on Windows to prevent a deadlock where lazy-
+    loading ``sentence_transformers`` + ``torch`` inside a FastMCP executor
+    worker thread blocks indefinitely on DLL init / OpenMP thread-pool
+    registration.
+
+    No-op when ``sentence-transformers`` is not installed (cloud-provider
+    setups remain unaffected) or when the configured model is already cached.
+
+    Args:
+        model_name: Optional override; falls back to the ``CRG_EMBEDDING_MODEL``
+            environment variable and then to ``LOCAL_DEFAULT_MODEL``.
+    """
+    try:
+        from sentence_transformers import SentenceTransformer  # noqa: F401
+    except ImportError:
+        return  # cloud-only setup: nothing to pre-warm
+
+    resolved = model_name or os.environ.get(
+        "CRG_EMBEDDING_MODEL", LOCAL_DEFAULT_MODEL
+    )
+    if resolved in _MODEL_CACHE:
+        return
+
+    try:
+        _MODEL_CACHE[resolved] = LocalEmbeddingProvider(resolved)._get_model()
+    except Exception as exc:  # pragma: no cover — best-effort startup hook
+        logger.warning("prewarm_local_embeddings(%s) skipped: %s", resolved, exc)
+
+
 class LocalEmbeddingProvider(EmbeddingProvider):
     def __init__(self, model_name: str | None = None) -> None:
         self._model_name = model_name or os.environ.get(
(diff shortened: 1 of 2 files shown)


tirth8205:

PR #517, opened 2026-05-28
Add VB.NET Language Support

liteAdamAnnosantini:
This PR adds VB.NET language support to code-review-graph.  Since there is no language pack for VB, I had codex create a fallback parser, similar to how ReScript is handled.

A sample VB file was included, as were tests similar in design to those present for C#.

### DIFF
--- a/code_review_graph/parser.py
+++ b/code_review_graph/parser.py
@@ -141,6 +141,9 @@ EXTENSION_TO_LANGUAGE: dict[str, str] = {
     ".svh": "verilog",
     ".v": "verilog",
     ".vh": "verilog",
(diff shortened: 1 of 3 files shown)


tirth8205:
Thanks — VB.NET support is valuable, but this head needs a correctness pass before merge.

Please rebase on current main, model namespaces in qualified names, resolve Inherits and Implements to real graph nodes, and make symbol lookup case-insensitive. Cover property bodies and common VB call/signature forms. Add graph-store traversal tests—not only name-presence tests—for namespace collisions, mixed-case calls, inheritance/interfaces, Me.Foo, properties, overloads, generics, and multiline signatures, then run full CI.

PR #522, opened 2026-05-30
feat: Enhance sv support

chuanseng-ng:
Changes:
- Further enhance system-verilog support with ports, nets/wires, parameters support
- Add packages, typedef, modports support
- Add port connections and verification construct support

Tests: PASS
Linting: PASS


tirth8205:
Thanks — this has useful SystemVerilog coverage and earlier CI was green. One graph correctness issue remains: data_declaration nodes inside functions and tasks are emitted as module-level signals. The same local name in two functions therefore collides and can create false signal/reference relationships.

Please keep function-local declarations in their real scope or exclude them from module signal indexing, add tests for duplicate local names and hierarchical/member port references, rebase on current main, and run the complete CI suite on the resulting merge.

PR #527, opened 2026-06-02
fix(graph): correct TESTED_BY edge direction in tests_for queries

### LINKED ISSUE #515: Bug: tests_for query uses wrong edge direction for TESTED_BY, always falls back to naming convention
## Summary

The `tests_for` query pattern (and `get_transitive_tests` in `graph.py`) looks up `TESTED_BY` edges using `target_qualified = <production_function>`, but the parser stores these edges with `source = production_function` and `target = test_function`. Because the direction is inverted between build and query, `tests_for` never matches any `TESTED_BY` edges from the graph and silently falls back to naming-convention heuristics (`test_{name}` / `Test{name}`).

This means `query_graph(pattern=\"tests_for\", target=\"...::add\")` returns results only when a test happens to be named `test_add` or `TestAdd`, not when the parser actually discovered a `TESTED_BY` relationship.

## Build side (parser.py)

In every parser path that handles test files, `TESTED_BY` edges are created by reversing a `CALLS` edge:

```python
# code_review_graph/parser.py:977-993
if test_file:
    test_qnames = set()
    for n in nodes:
        if n.is_test:
            qn = self._qualify(n.name, n.file_path, n.parent_name)
            test_qnames.add(qn)
    for edge in list(edges):
        if edge.kind == "CALLS" and edge.source in test_qnames:
            edges.append(EdgeInfo(
                kind="TESTED_BY",
                source=edge.target,        # production function
                target=edge.source,        # test function
                file_path=edge.file_path,
                line=edge.line,
            ))
```

So the stored edge is:
- `source_qualified` = `src/calc.py::add`  (production)
- `target_qualified` = `src/test_calc.py::test_add`  (test)

## Query side (tools/query.py)

```python
# code_review_graph/tools/query.py:295-308
elif pattern == "tests_for":
    for e in store.get_edges_by_target(qn):
        if e.kind == "TESTED_BY":
            test = store.get_node(e.source_qualified)
```

`qn` is the production function. `get_edges_by_target(qn)` searches `WHERE target_qualified = <production_function>`. But the parser stored the production function in `source_qualified`, not `target_qualified`. Therefore this loop never yields any edges.

The same inverted assumption exists in `graph.py:get_transitive_tests`:

```python
# code_review_graph/graph.py:418-430
for row in conn.execute(
    \"SELECT source_qualified FROM edges \"
    \"WHERE target_qualified = ? AND kind = 'TESTED_BY'\",
    (qn,)
).fetchall():
```

## Test data is also inconsistent

- `tests/test_tools.py:1165` creates `TESTED_BY` with `source=\"auth.py::login\"` (production) and `target=\"tests/test_auth.py::test_login\"` (test) — matching the parser.
- `tests/test_graph.py:382` creates `TESTED_BY` with `source=test_qn` (test) and `target=callee_qn` (production) — the opposite direction.

## Expected behavior

`tests_for` should look up `TESTED_BY` edges by **source** (`get_edges_by_source(qn)`) because the parser stores the production function in `source_qualified`:

```python
for e in store.get_edges_by_source(qn):
    if e.kind == "TESTED_BY":
        test = store.get_node(e.target_qualified)   # target = test function
```

Alternatively, the parser could be changed to store `source = test_function, target = production_function` to match the current query convention. Either way, build and query must agree on the same direction.

## Environment

- code-review-graph v2.3.5
- Python 3.10+

Devilthelegend:
## Summary

`query_graph(pattern="tests_for")` traversed `TESTED_BY` edges in the wrong direction. The parser stores these edges as **source = production node, target = test node**, but three sites in the code looked them up as if the direction were reversed:

1. `code_review_graph/tools/query.py` — the MCP-facing `tests_for` branch.
2. `code_review_graph/graph.py::get_transitive_tests` — three SQL queries (direct match, bare-name fallback, transitive via `CALLS`).
3. `code_review_graph/graph.py::load_flow_adjacency` — the `has_tested_by` set was populated with test nodes instead of production nodes, corrupting criticality scoring's coverage-gap detection.

Because `query_graph` falls back to a naming-convention search (`test_<name>` / `Test<name>`) when no edges match, the bug was silently masked for tests following standard naming. It surfaced for users whose tests had unconventional names or whose test discovery relied on the underlying edges.

## Fix

Flipped the traversal direction in all three places so that for a production node `qn`:

- Edges are looked up as `source_qualified = qn`.
- The matching test node is read from `target_qualified`.

For `load_flow_adjacency`, `has_tested_by.add(src)` so the set represents production nodes that have at least one test covering them — which is what `flows.compute_criticality` expects when checking for coverage gaps.

## Tests

Added three regression tests that deliberately use **unconventional test names** (`verify_addition`, `verify_compute`, `verify_combine_behaviour`) so the naming-convention fallback cannot mask future regressions:

- `tests/test_graph.py::TestGraphStore::test_get_transitive_tests_follows_direct_tested_by_edge`
- `tests/test_graph.py::TestGraphStore::test_get_transitive_tests_follows_calls_then_tested_by`
- `tests/test_tools.py::TestQueryGraphTestsFor::test_query_graph_tests_for_finds_direct_edge`

All three pass locally. The broader `test_graph.py`, `test_tools.py`, and `test_flows.py` suites show no real-test regressions (only pre-existing Windows SQLite teardown `PermissionError`s, which affect every test in those classes).

## Blast radius

| Component | Before | After |
|---|---|---|
| `query_graph(pattern="tests_for")` | Wrong results / empty + naming-convention fallback | Correct direct lookup |
| `get_transitive_tests` (used by review context, hints, criticality) | Wrong direction at every hop | Correct |
| `compute_criticality` coverage-gap detection | All flows scored as under-tested | Reflects real coverage |
| Naming-convention fallback | Kept as belt-and-braces | Kept |

Fixes #515


Devilthelegend:
All 9 actual checks are green. The pending "expected" status looks like a stale required-context entry in branch protection — could you take a look when you get a chance?

tirth8205:
Thanks @Devilthelegend — thorough work. I applied and tested locally: all five inverted consumer sites are fixed (including enrich.py and load_flow_adjacency, which my own audit missed), the already-correct readers in analysis.py/build.py/review.py were rightly left alone, your three regression tests fail on main and pass with the fix, and the full suite is green (1277 passed). Since only the parser writes TESTED_BY edges, existing DBs need no migration — good that this is read-side only.

One ask before merge: the producer side of the contract is still unpinned. test_parser.py::test_tested_by_edges_generated only asserts edges exist, not direction — a future parser flip would pass your new tests (they hand-insert edges) and silently reintroduce #515. Either assert source=production/target=test there, or add one end-to-end parse -> store -> get_transitive_tests test.

Non-blocking: refactor.py:495-507 still reads TESTED_BY by target — inert today (shadowed by companion CALLS edges) but worth a follow-up issue. The stale "expected" status check was a branch-protection misconfiguration on my side (a required context named "CI" that no job reports) — being fixed.

tirth8205:
Update: I've integrated this into the v2.4.0 release PR (#559), with the producer-side regression test we discussed added on top of your consumer-side fix — so a future parser flip can't silently re-break #515. Your commit is preserved with `Co-Authored-By` credit, and the refactor.py site #527 didn't cover is fixed in the same change.

Verified end to end: all five inverted consumer sites read the canonical direction, `tests_for` / `get_transitive_tests` / test-gap detection / flow criticality / enrichment / dead-code are correct, and the GitHub Action's "Tested" column now works. Full suite green (1474 passed), no DB migration needed.

Thank you — this was the keystone correctness fix for the whole review feature set. Crediting you in the 2.4.0 changelog.

tirth8205:
This contribution is already fully incorporated on main. Commit ddc8544 carries the complete TESTED_BY direction fix across graph.py, tools/query.py, changes.py, enrich.py, fixtures, and regression tests, with @Devilthelegend credited as co-author. Commit 580205d adds an end-to-end parser → store → get_transitive_tests contract test.

Current main also includes the later max_frontier performance guard, so merging this stale head would reintroduce conflicts without adding behavior. Closing as incorporated; issue #515 is fixed.

PR #529, opened 2026-06-03
fix(graph): guard store_file_batch against open transactions (#489)

### LINKED ISSUE #489: BEGIN IMMEDIATE in store_file_nodes_edges raises OperationalError when conn already in an implicit transaction
## Summary

`GraphStore.store_file_nodes_edges` in `code_review_graph/graph.py` issues `BEGIN IMMEDIATE` without first checking whether the connection is already inside an implicit transaction. Python's `sqlite3` module under its default deferred isolation auto-starts a transaction on the first DML statement, so any prior write on the same connection makes `BEGIN IMMEDIATE` raise:

```
```

This aborts any CLI path that reaches `store_file_nodes_edges` after such a write — I hit it in v2.2.1 during an automated `update` invocation chained off an editor hook.

## Stack trace

```
  File "/Users/.../bin/code-review-graph", line 10, in <module>
    sys.exit(main())
  File ".../code_review_graph/cli.py", line 501, in main
    result = build_or_update_graph(...)
  File ".../code_review_graph/tools/build.py", line 311, in build_or_update_graph
    result = incremental_update(root, store, base=base)
  File ".../code_review_graph/incremental.py", line 515, in incremental_update
    store.store_file_nodes_edges(...)
  File ".../code_review_graph/graph.py", line 239, in store_file_nodes_edges
    self._conn.execute("BEGIN IMMEDIATE")
```

## Deterministic reproduction

```python
import tempfile, pathlib
from code_review_graph.graph import GraphStore, NodeInfo

with tempfile.TemporaryDirectory() as d:
    store = GraphStore(pathlib.Path(d) / "graph.db")
    # Force the connection into an implicit transaction (any DML works).
    store._conn.execute(
        "INSERT OR REPLACE INTO metadata (key, value) VALUES (?, ?)",
        ("repro", "1"),
    )
    assert store._conn.in_transaction
    node = NodeInfo(kind="File", name="x.py", file_path="/x.py",
                    line_start=1, line_end=1)
    store.store_file_nodes_edges("/x.py", [node], [], "deadbeef")
    # ^ raises sqlite3.OperationalError
```

## Root cause

`sqlite3.connect(...)` in `GraphStore.__init__` uses the default `isolation_level="" (deferred)`, where Python opens an implicit transaction on the first INSERT/UPDATE/DELETE and only closes it on `commit()` / `rollback()`. `BEGIN IMMEDIATE` issued via `execute()` while one of those is open hits SQLite's "cannot start a transaction within a transaction" rule.

## Proposed minimal fix

```diff
--- a/code_review_graph/graph.py
+++ b/code_review_graph/graph.py
@@ -235,6 +235,14 @@ class GraphStore:
     def store_file_nodes_edges(
         self, file_path: str, nodes: list[NodeInfo], edges: list[EdgeInfo], fhash: str = ""
     ) -> None:
         """Atomically replace all data for a file."""
+        # Python's sqlite3 module auto-starts a deferred transaction on the
+        # first DML statement under default isolation. If the connection is
+        # already inside such an implicit transaction when we reach here,
+        # `BEGIN IMMEDIATE` raises "cannot start a transaction within a
+        # transaction". Flush any pending implicit transaction first so the
+        # IMMEDIATE write-lock acquisition succeeds cleanly.
+        if self._conn.in_transaction:
+            self._conn.commit()
         self._conn.execute("BEGIN IMMEDIATE")
```

## Audit

Grepping the package, `BEGIN IMMEDIATE` only appears once (`graph.py:239`), so this single guard fully closes the regression. Other `commit()` / `rollback()` call sites in `graph.py` look correct.

## Optional broader hardening

A more durable fix would be to switch the connection to autocommit mode (`sqlite3.connect(..., isolation_level=None)`) and manage every transaction explicitly. That removes a whole class of "stray DML left an implicit transaction open" bugs, but it changes the semantics of `upsert_node` / `upsert_edge` / similar (they currently rely on the caller's `commit()` to flush a batch), so the diff is bigger and probably wants its own PR.

## Regression test

```python
import tempfile, unittest
from pathlib import Path
from code_review_graph.graph import GraphStore, NodeInfo

class StoreFileNodesEdgesTransactionTest(unittest.TestCase):
    def setUp(self):
        self._tmp = tempfile.TemporaryDirectory()
        self.store = GraphStore(Path(self._tmp.name) / "g.db")

    def tearDown(self):
        self.store.close(); self._tmp.cleanup()

    def test_succeeds_when_implicit_transaction_open(self):
        self.store._conn.execute(
            "INSERT OR REPLACE INTO metadata (key, value) VALUES (?, ?)",
            ("repro", "1"),
        )
        self.assertTrue(self.store._conn.in_transaction)
        node = NodeInfo(kind="File", name="x.py", file_path="/x.py",
                        line_start=1, line_end=1)
        self.store.store_file_nodes_edges("/x.py", [node], [], "deadbeef")
        rows = self.store._conn.execute(
            "SELECT COUNT(*) AS n FROM nodes WHERE file_path = ?", ("/x.py",),
        ).fetchone()
        self.assertEqual(rows["n"], 1)
```

Without the fix: `OperationalError: cannot start a transaction within a transaction`.
With the fix: passes.

## Environment

- `code-review-graph` 2.2.1 (installed via `uv tool install`)
- Python 3.10.19
- macOS (darwin)

Devilthelegend:
store_file_nodes_edges already rolled back any uncommitted transaction before issuing BEGIN IMMEDIATE, but store_file_batch did not. When a caller had an explicit BEGIN open, store_file_batch raised sqlite3.OperationalError: cannot start a transaction within a transaction.

Extract a small _begin_immediate() helper and use it from both call sites so the guard stays consistent. Adds a regression test that exercises both methods with an explicit BEGIN already active.

Fixes #489

### DIFF
--- a/code_review_graph/graph.py
+++ b/code_review_graph/graph.py
@@ -260,14 +260,20 @@ class GraphStore:
         self._conn.execute("DELETE FROM edges WHERE file_path = ?", (file_path,))
         self._invalidate_cache()
 
-    def store_file_nodes_edges(
-        self, file_path: str, nodes: list[NodeInfo], edges: list[EdgeInfo], fhash: str = ""
-    ) -> None:
-        """Atomically replace all data for a file."""
+    def _begin_immediate(self) -> None:
+        """Start an IMMEDIATE transaction, rolling back any prior uncommitted
+        transaction first (regression guard for #135 / #489).
+        """
         if self._conn.in_transaction:
             logger.warning("Rolling back uncommitted transaction before BEGIN IMMEDIATE")
             self._conn.rollback()
         self._conn.execute("BEGIN IMMEDIATE")
+
+    def store_file_nodes_edges(
+        self, file_path: str, nodes: list[NodeInfo], edges: list[EdgeInfo], fhash: str = ""
+    ) -> None:
+        """Atomically replace all data for a file."""
+        self._begin_immediate()
         try:
             self.remove_file_data(file_path)
             for node in nodes:
@@ -284,7 +290,7 @@ class GraphStore:
         self, batch: list[tuple[str, list[NodeInfo], list[EdgeInfo], str]]
     ) -> None:
         """Atomically replace data for a batch of files in one transaction."""
-        self._conn.execute("BEGIN IMMEDIATE")
+        self._begin_immediate()
         try:
             for file_path, nodes, edges, fhash in batch:
                 self.remove_file_data(file_path)
--- a/tests/test_graph.py
+++ b/tests/test_graph.py
@@ -154,6 +154,28 @@ class TestGraphStore:
         assert self.store.get_node(new_path) is not None
         assert self.store.get_node("/test/file_0.py") is None
 
+    def test_store_with_open_transaction_no_error(self):
+        """Regression test for #489: store_file_nodes_edges and
+        store_file_batch must not raise 'cannot start a transaction
+        within a transaction' when the caller has an explicit BEGIN open.
+        """
+        node_a = self._make_file_node("/test/a.py")
+        node_b = self._make_file_node("/test/b.py")
+
+        # Force an open transaction on the shared connection.
+        self.store._conn.execute("BEGIN")
+        assert self.store._conn.in_transaction
+
+        # Must not raise sqlite3.OperationalError.
+        self.store.store_file_nodes_edges("/test/a.py", [node_a], [])
+        assert self.store.get_node("/test/a.py") is not None
+
+        # Re-open the transaction and verify the batch path is guarded too.
+        self.store._conn.execute("BEGIN")
+        assert self.store._conn.in_transaction
+        self.store.store_file_batch([("/test/b.py", [node_b], [], "")])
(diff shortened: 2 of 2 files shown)


tirth8205:
Thanks @Devilthelegend — this is a clean, well-scoped fix and I've merged it. I applied it locally on top of main: test_graph.py, test_transactions.py, test_incremental.py, and test_tools.py all pass (165 tests), your regression test fails without the fix and passes with it, and ruff/mypy are clean. Extracting _begin_immediate() instead of duplicating the guard was the right call — it gives us one place to change when we harden this further.

Two notes for the record, neither blocking: store_file_batch currently has no in-package callers, so this is API-consistency hardening rather than a hot-path fix; and the guard's rollback semantics mean a leaked transaction's writes are discarded rather than committed — that matches the existing convention in test_transactions.py, but I'm filing a follow-up to serialize writers with a lock so a concurrent thread's in-flight transaction can never be rolled back from under it. Appreciate the regression test covering both call sites.

PR #539, opened 2026-06-08
feat(parser): Markdown and YAML knowledge extractors

otalrapha:
## What
Adds two regex-based extractors so the graph can index knowledge files that have no bundled tree-sitter grammar:

- **Markdown** (`.md/.mdx/.qmd`): headings → `Section` nodes, nesting → `CONTAINS` edges, links (backtick filenames, `[[wikilinks]]`, `[text](x.md)`) → `REFERENCES` edges. Fenced code blocks are skipped.
- **YAML** (`.yaml/.yml`): top-level keys → `Section` nodes, registry-style `id:`/`name:` list entries → `Type` nodes, with `CONTAINS` edges. Pure regex, **no PyYAML dependency**.

## Why
Both reuse the existing `NodeInfo`/`EdgeInfo` schema and the shared `CONTAINS`/`REFERENCES` edge kinds, so documentation and config land in the **same graph as code**: a `spec.md` can `REFERENCES` the source file it describes, and a `CRON-123` registry entry becomes a queryable node. Helpful for repos that mix code with specs, docs and config registries.

## Approach
Follows the existing pattern for grammar-less inputs (`_parse_rescript`, `_parse_sql`): extension mapping in `EXTENSION_TO_LANGUAGE`, dispatch in `parse_bytes`, dedicated `_parse_markdown` / `_parse_yaml` methods.

## Tests
`tests/test_knowledge_extractors.py` — 4 cases (headings/nesting, fenced-code skip, references, yaml entries). `ruff check` passes clean, `pytest` green locally.

### DIFF
--- a/code_review_graph/parser.py
+++ b/code_review_graph/parser.py
@@ -142,6 +142,14 @@ EXTENSION_TO_LANGUAGE: dict[str, str] = {
     ".v": "verilog",
     ".vh": "verilog",
     ".sql": "sql",
+    # Knowledge formats (.md/.yaml): no tree-sitter grammar bundled, so they use
+    # regex-based extraction (same approach as ReScript). This lets documentation and
+    # config live in the same graph as code, linkable via CONTAINS/REFERENCES edges.
+    ".md": "markdown",
+    ".mdx": "markdown",
+    ".qmd": "markdown",
+    ".yaml": "yaml",
+    ".yml": "yaml",
 }
 
 # Shebang interpreter → language mapping for extension-less Unix scripts.
@@ -939,6 +947,12 @@ class CodeParser:
         if language == "sql":
             return self._parse_sql(path, source)
 
+        # Knowledge formats: regex-based extraction (no tree-sitter grammar bundled).
+        if language == "markdown":
+            return self._parse_markdown(path, source)
+        if language == "yaml":
+            return self._parse_yaml(path, source)
+
         parser = self._get_parser(language)
         if not parser:
             return [], []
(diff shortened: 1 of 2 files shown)


tirth8205:
Thanks for the contribution. The parser-level tests pass, but the persisted graph is not connected correctly: File nodes use full paths while the new CONTAINS/REFERENCES edges use path.name and custom bare section IDs, so those edges do not point to stored nodes.

There are significant false positives. Ordinary inline code such as python or GraphStore becomes a reference, and YAML creates separate Types for both id: CRON-1 and name: alpha. Link directories are discarded, section spans stop on the heading line, and fence handling covers only simple triple-backtick toggling.

Please use canonical endpoints, add a full_build/query integration test, narrow Markdown reference and YAML registry detection, preserve link paths, add negative/error cases, and provide graph-size/build-time evidence. This would automatically index 61 files in this repository. Coordinate YAML design with #577, rebase, and run full CI.

tirth8205:
Thanks for the parser work. I am closing head `b4cd950` because it cannot merge into current main and remains 246 commits behind. Its parser-level tests miss the persisted-graph failure: the exact full-build harness generated 1,619 `CONTAINS`/`REFERENCES` edges and all 1,619 were dangling because File nodes use full paths while these edges use `path.name` and bare section IDs. It also generated 1,038 references, treating ordinary inline code such as `python` and `GraphStore` as targets; link directories are discarded; and YAML emits separate Types for an entry's `id` and `name`. Only GitGuardian ran on this head.

A future contribution should use canonical qualified endpoints, include full-build/query integration and dangling-edge assertions, preserve link paths and section spans, add conservative Markdown/YAML detection with negative and malformed-input cases, measure graph-size/build-time impact, and align with current YAML/Ansible classification before running full CI.

marcus-sa:
Is markdown support on the path?

PR #554, opened 2026-06-12
fix(daemon): write PID and install signal handlers in foreground mode

Maple-Aikon:
## Linked issue

Closes a daemon lifecycle bug found while dogfooding `crg-daemon start --foreground`. No public issue number — happy to open one if maintainer prefers a tracked issue.

## What & why

`crg-daemon start --foreground` leaks the PID file and ignores SIGTERM/SIGHUP. Cause: `_handle_start` short-circuits to `daemonize()` when `is_daemon_running()` is False. The non-daemonize (foreground) branch returns early without writing the PID file or installing signal handlers, so:

- `crg-daemon stop` fails with "PID file missing"
- The process cannot be terminated by signal
- `crg watch` / `crg add` workflows that need `crg-daemon` running in foreground become un-killable

The Windows branch of `daemonize()` already does `write_pid() + self._setup_signal_handlers()` — this PR applies the same pattern to the Linux/macOS foreground branch.

### Diff

```diff
 def _handle_start(args: argparse.Namespace) -> None:
     """Start the daemon process."""
-    from .daemon import WatchDaemon, is_daemon_running, load_config
+    from .daemon import WatchDaemon, is_daemon_running, load_config, write_pid

     if is_daemon_running():
         print("Error: Daemon is already running.")

     if not args.foreground:
         daemon.daemonize()
+    else:
+        write_pid()
+        daemon._setup_signal_handlers()

     daemon.run_forever()
```

`uv.lock` was auto-updated by the cherry-pick (one transitive bump) — happy to drop it from the PR if maintainer prefers source-only diffs.

## How it was tested

```bash
# Cherry-picked onto v2.3.6 base (b72413c)
$ uv run pytest tests/test_daemon.py -v --tb=short -q
============================== 51 passed in 0.29s ==============================

$ uvx ruff check code_review_graph/daemon_cli.py
All checks passed!

$ uvx mypy code_review_graph/daemon_cli.py --ignore-missing-imports --no-strict-optional
Success: no issues found in 1 source file
```

**Manual functional test** (with isolated `HOME`):

```bash
# Start foreground daemon
$ HOME=/tmp/crg-test/home uv run crg-daemon start --foreground &
# Process stays alive, PID file written:
$ cat /tmp/crg-test/home/.code-review-graph/daemon.pid
14469

# Send SIGTERM
$ kill -TERM 14466
# Daemon log shows graceful shutdown:
INFO: Received signal 15 — shutting down
INFO: Daemon stopped
# Process exits 0 within 5s
```

Before the fix: SIGTERM was ignored and the process had to be killed with SIGKILL.

## Checklist

- [x] Tests added for new functionality — N/A, this is a bug fix that brings the foreground branch in line with the Windows branch already covered by existing tests
- [x] All tests pass: `uv run pytest tests/test_daemon.py --tb=short -q` → 51/51
- [x] Linting passes: `uvx ruff check code_review_graph/daemon_cli.py` → clean
- [x] Type checking passes: `uvx mypy code_review_graph/daemon_cli.py --ignore-missing-imports --no-strict-optional` → clean
- [ ] Docs updated where behavior changed — happy to add a note to docs/CLI.md about foreground signal handling if maintainer wants

### DIFF
--- a/code_review_graph/daemon_cli.py
+++ b/code_review_graph/daemon_cli.py
@@ -30,7 +30,7 @@ logger = logging.getLogger(__name__)
 
 def _handle_start(args: argparse.Namespace) -> None:
     """Start the daemon process."""
-    from .daemon import WatchDaemon, is_daemon_running, load_config
+    from .daemon import WatchDaemon, is_daemon_running, load_config, write_pid
 
     if is_daemon_running():
         print("Error: Daemon is already running.")
@@ -42,6 +42,9 @@ def _handle_start(args: argparse.Namespace) -> None:
 
     if not args.foreground:
         daemon.daemonize()
+    else:
+        write_pid()
+        daemon._setup_signal_handlers()
 
     daemon.run_forever()
 


tirth8205:
Thanks — the foreground bug is confirmed: current main still skips PID creation and signal setup for --foreground. I am not merging this head unchanged because it has no regression test, and lifecycle setup still happens only after WatchDaemon.start(), leaving a startup window.

We are carrying the focused fix onto current main with explicit startup ordering plus PID creation, SIGTERM shutdown, and PID cleanup tests, preserving your attribution. No further action is required unless you would like to contribute those tests.

tirth8205:
Merged the foreground lifecycle fix through maintainer PR #630 (merge commit 9c26c62), with @Maple-Aikon credited as co-author in commit 52bde2c.

The port strengthens the original change: PID creation and signal handlers happen before foreground watcher threads start, POSIX daemonization happens before any background threads, and cleanup is guaranteed after startup failure or shutdown. New tests first reproduced all three failures, then 52 daemon unit tests and the complete Linux Python 3.10–3.13 CI passed. Closing this source PR as incorporated.

PR #560, opened 2026-06-14
feat: julia - broaden parser construct coverage

danvinci:
**New constructs captured:**
- Function stubs (`function foo end`)
- Qualified defs (`function Base.show`) — node carries `julia_module_qualifier`
- Qualified calls (`Mod.f(...)`) — edge carries `julia_call_module`
- Multi-segment paths (`LinearAlgebra.BLAS.gemv`) — full dotted qualifier
- `const` type aliases (`const V = Vector{Float64}`) → `Type` nodes
- Aliased imports (`import X as Y`)

**Bug fixes:**
- Operator defs (`Base.:+`) produced a bogus node named after the module; now named by the operator
- Short-form one-liner bodies (`f(x) = g(x)`) lost their call edges; now captured

**Validation:** 
- 11 new tests, full suite 1394 passed. 
- Parsed several installed packages (Makie, DataFrames, Lux, Distributions, …) on `main` vs this branch: 1607 .jl files, 0 parse errors

### DIFF
--- a/code_review_graph/parser.py
+++ b/code_review_graph/parser.py
@@ -3093,11 +3093,11 @@ class CodeParser:
         for child in call_expr.children:
             if child.type == "identifier":
                 return child.text.decode("utf-8", errors="replace")
+            if child.type == "operator":
+                # bare operator def: ``+(a, b) = ...``
+                return child.text.decode("utf-8", errors="replace")
             if child.type == "field_expression":
-                for ident in reversed(child.children):
-                    if ident.type == "identifier":
-                        return ident.text.decode("utf-8", errors="replace")
-                return None
+                return self._julia_field_method_name(child)
             if child.type == "parametrized_type_expression":
                 for ident in child.children:
                     if ident.type == "identifier":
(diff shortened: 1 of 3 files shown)


tirth8205:
Thanks — this contains valuable Julia improvements, and the substantive CI matrix is green; the PR Review failure is only the fork workflow's 403 while posting its comment.

The head is now stale/conflicting, and current main independently covers some qualified definitions and short-form calls, so a direct merge is unsafe. We are porting the non-overlapping stubs, operators, aliased imports, multi-segment qualifiers, and parameterized const-alias cases with your attribution.

During the port, module qualifiers will participate in canonical identity and call resolution. Storing julia_module_qualifier only in extra does not stop Base.show and local show from colliding, and julia_call_module is ignored downstream. We will add collision and end-to-end resolution tests. No further action is required unless you prefer to rebase it yourself.

tirth8205:
Thank you, Dan. I reconciled this contribution against current main through PR #647, merged as 218d2b5728399ec5d3e3c1df1b17ede521537d20. Your authorship is preserved in the merged commit.\n\nThe adopted work covers Julia function stubs; short- and long-form bare, quoted, Base-qualified, and multi-segment operators; parameterized const type aliases without misclassifying value constants; direct and selected import aliases; one-line body calls; nested modules/functions/testsets; and persisted qualified graph identities.\n\nThe source branch could not merge directly because its qualified-name representation collapsed local and foreign methods, produced a bogus Base identity for quoted operators, lost multi-segment selected-import paths, treated wrapped definitions as self-calls, and mixed explicit module qualifiers into lexical lookup. The reconciled parser keeps collision-free stored identities while resolving bare calls through the true lexical module, preserves exact nested symbols under ordinary and qualified methods, keeps explicit qualified calls distinct, and corrects enum variant parents.\n\nFinal evidence on the exact merged head: 51 focused Julia cases passed; the full suite passed with 1,674 passed, 5 skipped, and 2 expected xpasses at 76.29% coverage; Ruff, mypy across 63 files, Bandit, schema, diff, and graph rebuild passed; all 11 GitHub checks passed including Windows. Closing this source PR as superseded by the adopted and hardened implementation, not as a rejection of the contribution.

PR #564, opened 2026-06-18
fix(visualization): use #graph-svg selector instead of bare svg

### LINKED ISSUE #523: graph.html renders into legend icon, not #graph-svg — d3.select("svg") selects wrong element (tiny graph, top-left)
## Summary

The generated `graph.html` (from `code-review-graph visualize` / `build`) renders the **entire force graph into the legend's tiny "File" icon** instead of the main `#graph-svg` canvas. You get a microscopic scatter crammed into the top-left legend swatch while the full-screen canvas stays empty.

## Environment
- code-review-graph **2.3.5**
- graph.html opened via `file://` (Chrome)
- Repo graph: ~233 nodes / ~312 edges (happens on any non-trivial repo)

## Root cause

In `code_review_graph/visualization.py` the embedded HTML template does:

```js
var svg = d3.select("svg").attr("viewBox", [0, 0, W, H]);
```

`d3.select("svg")` returns the **first `<svg>` in document order**. The legend `<nav>` is emitted *before* `#graph-svg`, and its first child is the File-kind icon:

```html
<svg width="16" height="16" viewBox="-8 -8 16 16" aria-hidden="true"><circle r="6" fill="#58a6ff"/></svg>
```

So `gRoot` + all nodes/links get appended into that 16px icon (viewBox `-8 -8 16 16`), producing a tiny graph inside the legend and an empty `#graph-svg`. The hosted demo works because it isn't competing with inline legend SVGs.

There are **two** affected template copies — both `var svg = d3.select("svg")` occurrences.

## Fix

Target the dedicated element by id (it already has `id="graph-svg"`):

```js
var svg = d3.select("#graph-svg").attr("viewBox", [0, 0, W, H]);
```

Everything downstream (`gRoot`, zoom, `fitGraph` on `simulation.on("end")`) chains off this one selector, so the single change fixes the whole render.

## Repro
1. Build a graph for any repo with a few hundred nodes.
2. `code-review-graph visualize` → open `.code-review-graph/graph.html`.
3. Graph appears tiny inside the top-left "File" legend item; main canvas empty.

itxaiohanglover:
## Summary

- `d3.select("svg")` returns the first `<svg>` in document order, which is a 16px legend icon — not the main `#graph-svg` canvas. This caused the entire force graph to render inside the tiny legend swatch while the full-screen canvas stayed empty.
- Fix: target `#graph-svg` by id in both template copies (full mode + aggregated/community mode).
- Added `id="graph-svg"` to the aggregated-view template's `<svg>` element (it was missing).

## Root cause

In `visualization.py`, the legend `<nav>` is emitted *before* `#graph-svg` and its first child is an inline `<svg>` icon:

```html
<svg width="16" height="16" viewBox="-8 -8 16 16" aria-hidden="true">...</svg>
```

`d3.select("svg")` matches this 16px icon, so `gRoot` + all nodes/links get appended into it (viewBox `-8 -8 16 16`), producing a microscopic graph in the legend and an empty canvas.

## Fix

```js
// Before
var svg = d3.select("svg").attr("viewBox", [0, 0, W, H]);

// After
var svg = d3.select("#graph-svg").attr("viewBox", [0, 0, W, H]);
```

Applied to both template copies. The aggregated-view template's `<svg>` also gets `id="graph-svg"` added (it was missing).

## Test plan

- [x] `test_generate_html_uses_id_selector_for_svg` — verifies full-mode HTML uses `d3.select("#graph-svg")` and contains no bare `d3.select("svg")`
- [x] `test_community_mode_uses_id_selector_for_svg` — verifies community-mode HTML uses `d3.select("#graph-svg")`, has no bare `d3.select("svg")`, and the `<svg>` has `id="graph-svg"`
- [x] All 22 existing visualization tests still pass
- [x] `ruff check` passes

Closes #523

### DIFF
--- a/code_review_graph/visualization.py
+++ b/code_review_graph/visualization.py
@@ -856,7 +856,7 @@ function moveTooltip(ev) {
 }
 function hideTooltip() { tooltip.classList.remove("visible"); }
 var W = innerWidth, H = innerHeight;
-var svg = d3.select("svg").attr("viewBox", [0, 0, W, H]);
+var svg = d3.select("#graph-svg").attr("viewBox", [0, 0, W, H]);
 var gRoot = svg.append("g");
 var currentTransform = d3.zoomIdentity;
 var zoomBehavior = d3.zoom()
@@ -1605,7 +1605,7 @@ _AGGREGATED_HTML_TEMPLATE = r"""<!DOCTYPE html>
 <div id="stats-bar" role="status" aria-label="Graph statistics"></div>
 <div id="tooltip"></div>
 <button id="btn-back" aria-label="Back to overview">&larr; Back to Overview</button>
-<svg role="img" aria-label="Interactive code knowledge graph visualization (aggregated view)."></svg>
+<svg id="graph-svg" role="img" aria-label="Interactive code knowledge graph visualization (aggregated view)."></svg>
 <script>
 "use strict";
 var graphData = __GRAPH_DATA__;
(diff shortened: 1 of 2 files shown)


tirth8205:
Thank you — the `#graph-svg` selector fix and both focused regressions were carried into #624 with your authorship and are now merged. This branch has not changed since the audited revision, so the PR is fully superseded. Closing it to avoid a duplicate merge.

Issue #558, opened 2026-06-14
[Bug]: claude-code hooks in settings.json shall not contain absolute paths to the repo


zeroisnan:
### code-review-graph version

Built on branch: main Built at commit: 06161b4f7f55

### Operating system

Linux

### Python version

3.12.3

### AI platform

claude-code

### Output of `code-review-graph status`

```shell
Nodes: 420
Edges: 2936
Files: 91
Languages: cpp, python
Last updated: 2026-06-14T02:42:17
Built on branch: main
Built at commit: 06161b4f7f55
```

### Steps to reproduce

code-review-graph install --platform claude-code

### Expected vs actual behavior

Expected:
.claude/settings.json is a repository level file impacying all collaborators on the repository, as such it should not contain user level aspects, such as paths to the repo

Actual: PostToolUse and SessionStart both contains a hard coded absolute path to the repo, making it not possible to share settings.json with other users/setups

Example:
```json
     {
        "matcher": "Edit|Write|Bash",
        "hooks": [
          {
            "type": "command",
            "command": "cat >/dev/null || true; git rev-parse --git-dir >/dev/null 2>&1 && code-review-graph update --skip-flows --repo \"/home/username/work/some/path/to/depot\" || true",
            "timeout": 30
          }
        ]
      }
```
Further to that, the install command will not replace existing hooks installed by previous versions of the tool, but add to the list instead

### Additional context

Here is a possible rewrite of the PostToolUse. The path to the depot is captured on the fly in a variable and used for the tool invocation

```json
            "command": "repo_root=$(git rev-parse --show-toplevel 2>/dev/null) && [ -n \"$repo_root\" ] && code-review-graph update --skip-flows --repo \"$repo_root\" || true",
````

itxaiohanglover:
Good point — `settings.json` is a repo-level file, embedding absolute checkout paths makes it useless for anyone else on the team.

Fixed in #565 (same PR as #549 since they touch the same `generate_hooks_config`). The hooks now resolve the repo root at runtime with `git rev-parse --show-toplevel`, so the file is shareable. The `repo_root` param is kept for backward compat but not embedded anymore.

clezcoding:
> [!NOTE]
[...]

PR #565, opened 2026-06-18
fix(skills): add PATH guard and remove absolute paths from hooks

### LINKED ISSUE #549: Generated Claude Code hooks fail noisily when the binary is not on PATH, and fire on every Bash command
## Summary

`code-review-graph install` (v2.3.2, `skills.py` → `generate_hooks_config()`) writes Claude Code hooks into `.claude/settings.json` that have two issues:

### 1. No install/PATH guard

The generated hooks call the bare binary:

```json
{ "type": "command", "command": "code-review-graph update --skip-flows", "timeout": 30 }
```

When the package is installed in a project virtualenv (the common case), the binary is not on the PATH that Claude Code's `/bin/sh` sees, so **every Edit/Write/Bash tool call** surfaces:

```
PostToolUse:Edit hook error
Failed with non-blocking status code: /bin/sh: 1: code-review-graph: not found
```

Notably, the git pre-commit hook installed by the very same `install` command already does this correctly:

```sh
if command -v code-review-graph >/dev/null 2>&1; then
    code-review-graph detect-changes --brief || true
fi
```

The Claude Code hooks just lack the same guard. Suggested fix in `generate_hooks_config()` — resolve the project venv binary first and exit silently when absent:

```sh
CRG="$CLAUDE_PROJECT_DIR/.venv/bin/code-review-graph"; command -v "$CRG" >/dev/null 2>&1 || CRG=code-review-graph; command -v "$CRG" >/dev/null 2>&1 || exit 0; "$CRG" update --skip-flows
```

### 2. Matcher is too broad

`"matcher": "Edit |Write|Bash"` means the graph update runs after **every shell command** — `git status`, `ls`, test runs, everything — not just file changes. `Edit|Write` covers the cases where the graph can actually become stale from within Claude Code; if Bash-driven file mutations are a concern, a scoped `if` filter (e.g. on git mutation commands) would be far cheaper than firing on all of Bash.

### Environment

- code-review-graph 2.3.2 (PyPI), installed in a project venv
- Claude Code on Linux (WSL2), hooks executed via /bin/sh

Happy to provide more detail if useful.

itxaiohanglover:
## Summary

Fixes two related bugs in `generate_hooks_config()`:

- **#549**: Hooks called bare `code-review-graph` without checking PATH. When installed in a project venv, every Edit/Write/Bash tool call surfaced `code-review-graph: not found`. Also, the `Bash` matcher fired on every shell command (`git status`, `ls`, test runs), not just file mutations.
- **#558**: Hooks embedded `json.dumps(repo_root.resolve().as_posix())` — an absolute path — making `.claude/settings.json` non-shareable across collaborators with different checkout paths.

### PATH guard (#549)
```sh
# Before
"command": "cat >/dev/null || true; git rev-parse --git-dir >/dev/null 2>&1 && code-review-graph update ..."

# After
"command": "cat >/dev/null || true; command -v code-review-graph >/dev/null 2>&1 || exit 0; git rev-parse --git-dir >/dev/null 2>&1 && code-review-graph update ..."
```

This mirrors the pattern already used by the git pre-commit hook installed by the same `install` command.

### Narrowed matcher (#549)
```python
# Before
"matcher": "Edit|Write|Bash"

# After
"matcher": "Edit|Write"
```

`Edit|Write` covers the cases where the graph can actually become stale from within Claude Code. Bash-driven file mutations are rare and the cost of firing on every shell command is high.

### Dynamic repo root (#558)
```sh
# Before
--repo "/home/username/work/some/path/to/depot"

# After
--repo "$(git rev-parse --show-toplevel 2>/dev/null)"
```

The `repo_root` parameter is retained for backward compatibility but no longer embedded in commands.

## Test plan

- [x] `test_hooks_have_path_guard` — verifies both hooks include `command -v code-review-graph`
- [x] `test_hooks_use_dynamic_repo_root` — verifies both hooks use `git rev-parse --show-toplevel`
- [x] `test_hooks_no_absolute_path_embedded` — verifies no hardcoded path appears in commands
- [x] `test_post_tool_use_matcher_excludes_bash` — verifies matcher is `Edit|Write`, not `Edit|Write|Bash`
- [x] All 144 existing skills tests still pass
- [x] `ruff check` passes (no new lint errors introduced)

Closes #549, #558

### DIFF
--- a/code_review_graph/skills.py
+++ b/code_review_graph/skills.py
@@ -550,21 +550,28 @@ def generate_hooks_config(repo_root: Path) -> dict[str, Any]:
     Hooks use the v1.x+ schema: each entry needs a ``matcher`` and a nested
     ``hooks`` array. Timeouts are in seconds. ``PreCommit`` is not a valid
     Claude Code event — pre-commit checks are handled by ``install_git_hook``.
+
+    The ``repo_root`` parameter is retained for backward compatibility but is
+    not embedded in hook commands. Instead, the repo root is resolved at
+    runtime via ``git rev-parse --show-toplevel`` so that ``settings.json``
+    is shareable across collaborators with different checkout paths.
+    A PATH guard ensures the hook exits silently when the binary is not on
+    ``$PATH`` (e.g. installed in a project venv).
     """
-    repo_arg = json.dumps(repo_root.resolve().as_posix())
     return {
         "hooks": {
             "PostToolUse": [
                 {
-                    "matcher": "Edit|Write|Bash",
+                    "matcher": "Edit|Write",
                     "hooks": [
                         {
                             "type": "command",
                             "command": (
                                 "cat >/dev/null || true; "
+                                "command -v code-review-graph >/dev/null 2>&1 || exit 0; "
                                 "git rev-parse --git-dir >/dev/null 2>&1"
-                                f" && code-review-graph update --skip-flows"
-                                f" --repo {repo_arg}"
+                                " && code-review-graph update --skip-flows"
+                                " --repo \"$(git rev-parse --show-toplevel 2>/dev/null)\""
                                 " || true"
                             ),
                             "timeout": 30,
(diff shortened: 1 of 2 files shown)


itxaiohanglover:
This PR addresses issue #558 — the hooks in `settings.json` should not contain absolute paths. The PATH guard ensures the hooks work regardless of the user's installation path.

tirth8205:
Thank you — the portable hook commands, PATH guard, narrowed matcher, and regressions were carried into #624 with your authorship and are now merged. This branch has not changed since the audited revision, so the PR is fully superseded. Closing it to avoid a duplicate merge.

PR #566, opened 2026-06-18
fix: add errors="replace" to all subprocess.run calls in incremental.py

lininn:
Prevents UnicodeDecodeError when git output contains non-UTF-8 or truncated multi-byte characters (e.g., branch names with CJK characters that get truncated by git rev-parse --abbrev-ref HEAD).

The _git_branch_info function was crashing at full_build because it lacked errors='replace' on subprocess.run() calls, while other functions in the same file (e.g., _get_svn_changed_files, _read_ignore_file) already used errors='replace'.

Also adds UnicodeDecodeError to except clauses so that any remaining decode failures are handled gracefully rather than crashing the build.

Closes #302

# Pull Request

## Linked issue

Closes #

## How it was tested

```bash
```

## Checklist

- [ ] Tests added for new functionality
- [ ] All tests pass: `uv run pytest tests/ --tb=short -q`
- [ ] Linting passes: `uv run ruff check code_review_graph/`
- [ ] Type checking passes: `uv run mypy code_review_graph/ --ignore-missing-imports --no-strict-optional`
- [ ] Lines are at most 100 characters

### DIFF
--- a/code_review_graph/incremental.py
+++ b/code_review_graph/incremental.py
@@ -416,25 +416,27 @@ def _git_branch_info(repo_root: Path) -> tuple[str, str]:
         result = subprocess.run(
             ["git", "rev-parse", "--abbrev-ref", "HEAD"],
             capture_output=True,
-            text=True, encoding='utf-8',            cwd=str(repo_root),
+            text=True, encoding='utf-8', errors='replace',
+            cwd=str(repo_root),
             timeout=_GIT_TIMEOUT,
             stdin=subprocess.DEVNULL,
         )
         if result.returncode == 0:
             branch = result.stdout.strip()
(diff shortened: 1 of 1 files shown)


itxaiohanglover (approved):
Good fix — `errors='replace'` is the right approach for subprocess output that may contain non-UTF-8 bytes (e.g. git output on Windows with non-ASCII paths). Adding `UnicodeDecodeError` to the except clause is correct defense-in-depth. The formatting cleanup (splitting the long line) is a nice bonus.

itxaiohanglover:
Good defensive fix. Adding `errors='replace'` to all subprocess calls prevents crashes on repos with non-UTF-8 git output. Catching `UnicodeDecodeError` as a fallback is belt-and-suspenders but harmless.

PR #568, opened 2026-06-18
fix: resolve scoped/static (`Class::method`) call targets for PHP and Rust

### LINKED ISSUE #567: Static/scoped calls (`Class::method`) aren't tracked as callers in PHP and Rust
## What's wrong

If you call a method with the scoped/static syntax (`Mailer::send($x)` in PHP, `Mailer::send(x)` in Rust), the call edge gets stored but `callers_of`, `get_impact_radius`, and `tests_for` never see it. They report zero callers for a method that's obviously being called.

The data isn't lost, it's just filed under a key nothing looks up. The parser records the CALLS edge with the target `Mailer::send`, but every node in the graph is keyed by its dotted qualified name, e.g. `src/Mailer.php::Mailer.send`. Those two strings never compare equal, and nothing in the resolution layer bridges them, so the edge dangles.

The underlying reason: the resolution layer assumes a call target is one of two shapes, a fully-qualified node name (`<file>::Class.method`) or a bare name (`method`). The `Class::method` form is neither, so it falls through every path.

This is on 2.3.6. I'd call it high severity. It's a common call style in general, and on Rust it's the *normal* way to call associated functions and constructors (`Foo::new()`, `Self::method()`), so a big chunk of Rust call edges quietly go missing, which makes blast-radius and review scope under-report.

## Which languages this actually affects

I built a small repo for each `::`-using language and looked at the `edges` table directly. Only PHP and Rust actually hit this:

- **PHP** `Mailer::send($x)` — stores `Mailer::send`, dangles against `…::Mailer.send`. Affected.
- **Rust** `Mailer::send(x)` — same thing. Affected.
- **C++** `Mailer::send(x)` — no CALLS edge gets created at all, so there's nothing to dangle. That's a different, extraction-side problem (looks adjacent to #463).
- **Ruby** — same as C++, no CALLS edge produced for the scoped call.
- **R** `pkg::fn(x)` — does produce `pkg::fn`, but `pkg` is an external package with no node in the graph, so there's nothing to resolve it to. Intra-file calls already resolve fine.

So this report is scoped to PHP and Rust, where the edge exists but points at the wrong key. C++/Ruby/R are out of scope here for the reasons above.

## Reproduction

Two PHP files:

`src/Mailer.php`
```php
<?php
class Mailer {
    public static function send($to) {
        return true;
    }
}
```

`src/SignupController.php`
```php
<?php
class SignupController {
    public function register($email) {
        return Mailer::send($email);  // scoped/static call — this is what breaks
    }
}
```

Build it and query:

```bash
```

```
query_graph(pattern="callers_of", target="Mailer.send")  ->  {"status": "ok", "results": []}
get_impact_radius(changed_files=["src/Mailer.php"])       ->  no method callers
```

You'd expect `SignupController.register` to show up as a caller of `Mailer.send`. Instead you get nothing.

The edge is in the DB, just keyed wrong:

```sql
sqlite> SELECT source_qualified, target_qualified FROM edges
        WHERE kind='CALLS' AND target_qualified LIKE '%send%';
src/SignupController.php::SignupController.register | Mailer::send   -- target is "Class::method"

sqlite> SELECT qualified_name FROM nodes WHERE name='send';
src/Mailer.php::Mailer.send                                          -- node key is "<file>::Class.method"
```

The same two-file pattern in Rust (`struct Mailer; impl Mailer { fn send(..) }`, called via `Mailer::send(..)`) reproduces identically.

If you poke at the store directly, it's clearly the resolution layer and not extraction:

```
store.resolve_bare_call_targets()                 -> 0   # batch resolver never touches it (it skips '::' targets)
store.search_edges_by_target_name("send")         -> 0   # the callers_of bare-name fallback misses it
store.search_edges_by_target_name("Mailer::send") -> 1   # only findable by the Class::method key, which nothing queries
store.get_edges_by_target("src/Mailer.php::Mailer.send") -> 1 CONTAINS edge only (no CALLS)
```

## Where it goes wrong

Three spots all make the same assumption (line numbers from `main` @ 2.3.6, they'll drift):

1. `code_review_graph/parser.py:2189` (`_resolve_call_targets`) only qualifies bare targets:
   ```python
   if edge.kind in ("CALLS", "REFERENCES") and "::" not in edge.target:
   ```
   `Class::method` has `::`, so it's skipped.

2. `code_review_graph/graph.py:499` (`resolve_bare_call_targets`) explicitly excludes anything with `::`:
   ```python
   "FROM edges WHERE kind = 'CALLS' AND target_qualified NOT LIKE '%::%'"
   ```

3. `code_review_graph/graph.py:370` (`search_edges_by_target_name`, the cross-file `callers_of` fallback) matches on the bare name:
   ```python
   "SELECT * FROM edges WHERE target_qualified = ? AND kind = ?", (name, kind)
   ```
   The stored target is `Mailer::send`, the lookup is `send`, no match.

The target itself is produced in the PHP scoped-call handler at `parser.py:6462`, which returns `f"{parts[0]}::{parts[-1]}"`. Worth noting: Rust's `scoped_identifier` (`parser.py:6567`) and R's `namespace_operator` (`parser.py:6571`) return the *full* scoped text, so a Rust target can carry more than one `::` (e.g. `crate::module::func`).

## Environment

- code-review-graph 2.3.6
- Reproduced on PHP and Rust
- Checked and ruled out: C++ and Ruby (no CALLS edge produced, separate extraction issue) and R (`::` is external package access, no in-graph target)

NaeemHaque:
## Summary

Scoped/static calls (`Mailer::send()` in PHP, `Mailer::send(x)` / `Self::method()` in Rust) were stored as CALLS edges with the intermediate `Class::method` target, which matches neither the qualified-name lookup (`<file>::Class.method`) nor the bare-name lookup. The edges dangled, so `callers_of`, `get_impact_radius`, and `tests_for` silently missed these callers — especially impactful in Rust, where `Type::method()` / `Self::method()` is the dominant call form.

Fixes #567.

## What this does

Adds a post-build resolver (`code_review_graph/scoped_resolver.py`) in the same style as the existing Spring/Temporal/ReScript resolvers. It rewrites resolvable `Class::method` CALLS targets to the canonical node qualified name, wired into both full build and incremental update (gated to `.php`/`.rs`).

It is deliberately conservative so it never fabricates an edge: 

- Only genuine two-part `Class::method` targets are resolved. Multi-segment Rust module paths (`a::b::c`) are left alone — resolving them by their last two segments would point at unrelated types.
- PHP namespaces are stripped before lookup (`App\Mailer` → `Mailer`).
- `self` / `Self` resolve to the enclosing class.
- Resolves on a single matching `(class, method)` node, or disambiguates via the caller file's `IMPORTS_FROM` edges. Ambiguous or unknown targets (external types like `Vec::new`) are left untouched.
- Rewritten edges are tagged `confidence_tier = INFERRED` to distinguish them from directly-extracted ones.

Known limitation: same-name collisions across files with no import statement have nothing to disambiguate on and stay unresolved. Strictly better than today, not 100%.

Scope is PHP and Rust — the two languages that actually produce a dangling `Class::method` edge. C++/Ruby produce no CALLS edge for scoped calls (a separate extraction gap), and R's `::` is external-package access; both are out of scope here.

## Tests

`tests/test_scoped_resolver.py` (14 tests): PHP cross-file and namespaced calls, Rust cross-file and `Self::`, the multi-segment false-edge guard, external-target-left-untouched, import disambiguation (resolved + ambiguous), incremental re-resolution, idempotency, and `INFERRED` tagging.

- `uv run pytest` — all pass (full suite green)
- `uv run ruff check code_review_graph/` — clean
- mypy clean; coverage gate (65%) passes

### DIFF
--- a/code_review_graph/incremental.py
+++ b/code_review_graph/incremental.py
@@ -93,6 +93,27 @@ def _run_temporal_resolver(store: GraphStore) -> Optional[dict]:
         logger.warning("Temporal resolver failed: %s", exc)
         return None
 
+
+# File extensions whose grammars emit resolvable ``Class::method`` scoped-call
+# targets.  Only PHP and Rust produce a dangling Class::method CALLS edge that
+# this pass can rewrite: C++/Ruby emit no CALLS edge for scoped calls (a
+# separate extraction gap), and R's ``pkg::fn`` is external package access with
+# no in-graph target.  Adding those extensions would only trigger wasted
+# full-graph scans, so they are intentionally excluded.
+_SCOPED_CALL_EXTS = (".php", ".rs")
+
+
+def _run_scoped_resolver(store: GraphStore) -> Optional[dict]:
+    """Run the scoped/static call (``Class::method``) resolver, swallowing any
+    failure so build never fails because of it. Returns stats or None on error.
+    """
+    try:
+        from .scoped_resolver import resolve_scoped_calls
+        return resolve_scoped_calls(store)
+    except Exception as exc:  # noqa: BLE001 - best-effort post-pass
+        logger.warning("Scoped call resolver failed: %s", exc)
+        return None
+
 # Default ignore patterns (in addition to .gitignore).
 #
 # `<dir>/**` patterns are matched at any depth by _should_ignore, so
@@ -904,6 +925,7 @@ def full_build(
     rescript_stats = _run_rescript_resolver(store)
     spring_stats = _run_spring_resolver(store)
     temporal_stats = _run_temporal_resolver(store)
+    scoped_stats = _run_scoped_resolver(store)
 
     return {
         "files_parsed": len(files),
@@ -913,6 +935,7 @@ def full_build(
         "rescript_resolution": rescript_stats,
         "spring_resolution": spring_stats,
         "temporal_resolution": temporal_stats,
+        "scoped_resolution": scoped_stats,
     }
 
 
@@ -1043,6 +1066,9 @@ def incremental_update(
     spring_stats = _run_spring_resolver(store) if spring_changed else None
     temporal_stats = _run_temporal_resolver(store) if spring_changed else None
 
+    scoped_changed = any(rp.endswith(_SCOPED_CALL_EXTS) for rp in all_files)
+    scoped_stats = _run_scoped_resolver(store) if scoped_changed else None
+
     return {
         "files_updated": len(all_files),
         "total_nodes": total_nodes,
@@ -1053,6 +1079,7 @@ def incremental_update(
         "rescript_resolution": rescript_stats,
         "spring_resolution": spring_stats,
         "temporal_resolution": temporal_stats,
+        "scoped_resolution": scoped_stats,
     }
 
 
--- a/code_review_graph/scoped_resolver.py
+++ b/code_review_graph/scoped_resolver.py
@@ -1,0 +1,170 @@
+"""Post-build scoped/static call resolver (``Class::method`` targets).
+
+Grammars that emit a ``::`` scope separator in call expressions store a CALLS
+edge whose target is the intermediate ``Class::method`` form rather than a
+node qualified name:
+
+    PHP   ``Mailer::send($x)``      -> target ``Mailer::send``
+    Rust  ``Notifier::notify(x)``   -> target ``Notifier::notify``
+
+Every node is keyed by its dotted qualified name (``<file>::Class.method``),
+so these ``Class::method`` targets match neither the qualified-name nor the
+bare-name lookup paths and dangle forever — ``callers_of`` /
+``get_impact_radius`` / ``tests_for`` silently under-report the caller.
+
+This module runs as a post-build pass (like the Spring/Temporal/ReScript
+resolvers) and rewrites resolvable ``Class::method`` targets to the canonical
+node qualified name.  Resolution is conservative: an edge is only rewritten
+when the ``(class, method)`` pair maps to exactly one node, or to exactly one
+node whose file the source file imports.  Unresolvable targets (external
+types such as ``Vec::new``) are left untouched so no false edge is created.
+
+The intra-class receivers ``self`` / ``static`` / ``Self`` / ``this`` resolve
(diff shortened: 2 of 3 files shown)


itxaiohanglover (approved):
Clean implementation — follows the existing resolver pattern (`_run_temporal_resolver` etc.) consistently. The `_SCOPED_CALL_EXTS` comment explaining why only PHP and Rust are included is helpful. 

One minor question: in `incremental_update`, the condition checks `spring_changed` for temporal stats but `scoped_changed` for scoped stats — is that intentional? If spring and scoped are independent, looks correct. Just confirming the temporal → spring coupling is existing behavior.

NaeemHaque:
Hey @itxaiohanglover  — thanks for taking the time to review this!
 
Yes, that's intentional. Temporal shares Spring's check on purpose, and scoped has its own. The temporal one is existing behavior — I just left it as-is.

itxaiohanglover:
Good work on this! I had commented on #567 about `Class::method` calls not being tracked. Restricting to PHP and Rust makes sense — the comment about C++/Ruby not emitting CALLS edges and R being external package access shows good understanding of the grammar differences.

The incremental update guard with `_SCOPED_CALL_EXTS` is efficient. Have you thought about adding a test fixture with a scoped call to verify the resolver actually rewrites the target correctly?

NaeemHaque:
That's already covered. `tests/test_scoped_resolver.py` has tests that do exactly this: they build a scoped call and assert the resolver rewrites the target to the real node (e.g. `Mailer::send → Mailer.php::Mailer.send`, and the raw `Mailer::send` is gone), plus the Rust equivalent (`Notifier::notify`). There are 14 tests in there covering self::, namespaced calls, the multi-segment guard, idempotency, etc.

tirth8205:
Thanks—the existing tests show that scoped calls are rewritten, but two blockers remain.

A globally unique `Class::method` can still be connected to an unrelated file without same-file or import evidence. Please leave that call unresolved unless there is trustworthy evidence, and add a regression for that false-positive case.

The resolver also performs full-graph work during incremental updates. Please use an indexed or otherwise bounded lookup and include an incremental performance regression or benchmark.

PR #571, opened 2026-06-22
fix(skills): string-aware trailing comma stripping for JSONC

### LINKED ISSUE #553: fix(install): trailing-comma regex breaks on commas inside string values
## Problem

`code_review_graph/skills.py:417` uses a regex to strip trailing commas before `}` or `]` when reading existing config files for merge:

```python
stripped = re.sub(r',(\s*[}\]])', r'\1', stripped)
```

This regex does **not** track JSON string boundaries. If a string value contains a comma followed by `}` or `]` on the next line, the regex will incorrectly remove the comma **inside the string**, producing invalid JSON that the subsequent `json.loads` silently misparses (or rejects outright).

## Concrete breakage

A user config like:

```jsonc
{
  "mcp": {
    "my-server": {
      "command": ["x"],
      "description": "foo, bar",
    }
  }
}
```

becomes (after the regex pass):

```json
{
  "mcp": {
    "my-server": {
      "command": ["x"],
      "description": "foo bar",
    }
  }
}
```

The comma inside `"foo, bar"` is deleted, silently corrupting the user's string value. Worse: a real `]` or `}` inside a string (e.g. `"see [docs]"`) will also be touched.

Same bug class as the `_strip_jsonc_comments` issue fixed in #530 (commit `47a43dc`) — that one was string-aware and resolved cleanly; this one is a leftover raw regex that needs the same treatment.

## Reproduction

```python
import re
raw = '{"a": "foo, bar", "b": 1}'
print(re.sub(r',(\s*[}\]])', r'\1', raw))
# -> {"a": "foo bar", "b": 1}   ← comma inside string deleted
```

## Suggested fix

Fold the trailing-comma strip into the existing stateful `_strip_jsonc_comments` scanner (introduced in #530) so both passes happen in one string-aware walk. The scanner already tracks string boundaries; extending it to also drop a `,` when the next non-whitespace token is `}` or `]` is a small change and removes the need for the post-pass regex entirely.

Sketch:

```python
def _strip_jsonc_and_trailing_commas(text: str) -> str:
    # same state machine as _strip_jsonc_comments, plus:
    #   in_state == "code" and c == "," and the next non-ws
    #   char is "}" or "]" -> drop the comma
```

Rename `_strip_jsonc_comments` -> `_strip_jsonc_and_trailing_commas` and delete the `re.sub` line at 417.

## Out of scope for #530

Owner (@tirth8205) confirmed in https://github.com/tirth8205/code-review-graph/pull/530#issuecomment that this is out of scope for the spec-compliance PR and asked for a separate follow-up issue. Filing per that request.

## Acceptance

- Add a test case with commas inside strings followed by `}` / `]` to confirm the strip is string-aware
- Remove the raw `re.sub` at line 417
- Full suite green (was 1282 at #530 merge time)

itxaiohanglover:
Fixes #553

## Problem

The regex `r',(\s*[}\]])'` in `skills.py:351` strips trailing commas before `}` or `]`, but does not track JSON string boundaries. If a string value contains a comma followed by `}` or `]` on the next line, the regex incorrectly removes the comma **inside the string**, corrupting the JSON.

Example breakage:
```jsonc
{
  "mcp": {
    "my-server": {
      "description": "foo, bar",
    }
  }
}
```
The comma in `"foo, bar"` gets removed → `"foo bar"` — silently corrupting user config.

## What changed

Replaced the regex with `_strip_trailing_commas_jsonc()` — a minimal stateful parser that:
1. Tracks `in_string` state (toggles on `"` outside of escape sequences)
2. Only strips commas that are followed by `}` or `]` **outside** of strings
3. Handles escape sequences (`\"`, `\\`) correctly

No new dependencies. No existing comments modified.

## Validation

- The existing `json.loads(stripped)` call on line 353 works unchanged
- Strings containing `,}` or `,]` patterns are preserved
- Trailing commas outside strings are still stripped correctly

### DIFF
--- a/code_review_graph/skills.py
+++ b/code_review_graph/skills.py
@@ -15,6 +15,7 @@ import platform
 import re
 import shutil
 import stat
+import subprocess
 import sys
 from pathlib import Path
 from typing import Any
@@ -289,6 +290,41 @@ def _merge_toml_mcp_server(
     return True
 
 
+def _strip_trailing_commas_jsonc(text: str) -> str:
+    """Strip trailing commas before } or ], respecting string boundaries."""
+    result = []
+    in_string = False
+    escape = False
+    i = 0
+    while i < len(text):
+        ch = text[i]
+        if escape:
+            result.append(ch)
+            escape = False
+            i += 1
+            continue
+        if ch == '\\':
+            result.append(ch)
+            escape = True
+            i += 1
+            continue
+        if ch == '"':
+            in_string = not in_string
+            result.append(ch)
+            i += 1
+            continue
+        if not in_string and ch == ',':
+            j = i + 1
+            while j < len(text) and text[j] in ' \t\n\r':
+                j += 1
+            if j < len(text) and text[j] in '}]':
+                i += 1
+                continue
+        result.append(ch)
+        i += 1
+    return ''.join(result)
+
+
 def install_platform_configs(
     repo_root: Path,
     target: str = "all",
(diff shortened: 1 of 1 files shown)


itxaiohanglover:
Closing this in favor of #578, which covers both JSONC comment stripping AND trailing comma removal in a single string-aware pass. That's more comprehensive than my approach here which still used regex for comment stripping. Nice work @dlwjdghkszkf-png!

PR #573, opened 2026-06-23
fix: resolve PHP `use` imports to file paths

### LINKED ISSUE #574: PHP `use` imports stored as raw statement text; importers_of and impact-radius return 0 for PHP classes
### code-review-graph version

2.3.6 (run as the MCP server). Also reproduced on the 2.3.2 CLI; the root cause
is present in both.

### Operating system

macOS 15 (Darwin 24.5.0), Apple Silicon.

### Python version

Python 3.14.4

### AI platform

Claude Code (MCP server over stdio).

From the minimal reproducer below:

```
Nodes: 6
Edges: 5
Files: 3
Languages: php
```

### Steps to reproduce

1. Create a minimal PSR-4 PHP project and build the graph:

```bash
mkdir -p repro/src/App/Domain/Entity repro/src/App/Service repro/src/App/Api
cd repro && git init -q
printf '<?php\nnamespace App\\Domain\\Entity;\nclass Job {}\n'                                   > src/App/Domain/Entity/Job.php
printf '<?php\nnamespace App\\Service;\nuse App\\Domain\\Entity\\Job;\nclass MatchService {}\n'  > src/App/Service/MatchService.php
printf '<?php\nnamespace App\\Api;\nuse App\\Domain\\Entity\\Job;\nclass JobController {}\n'      > src/App/Api/JobController.php
git add -A
code-review-graph build
```

2. Ask for the importers of the `Job` class (via the MCP `query_graph` tool, or
the package API):

```python
from code_review_graph.tools.query import query_graph
query_graph("importers_of", "<abs>/src/App/Domain/Entity/Job.php", repo_root="<abs>")
```

**Expected:** `importers_of(Job.php)` returns the two files that
`use App\Domain\Entity\Job;` (`MatchService.php` and `JobController.php`).
Equivalently, the `IMPORTS_FROM` edges should target the resolved path of
`Job.php`.

**Actual:** `importers_of` returns 0 results. Inspecting the graph shows the
`IMPORTS_FROM` edge targets are the raw `use` statement text instead of a
resolved file path or FQN:

```
src/App/Api/JobController.php      ->  use App\Domain\Entity\Job;
src/App/Service/MatchService.php   ->  use App\Domain\Entity\Job;

importers_of(Job.php)  ->  Found 0 result(s)
```

The same defect makes `tests_for` and `inheritors_of` return nothing for PHP
classes, hides PHP importers from the upstream side of `get_impact_radius`, and
(because `resolve_bare_call_targets` uses `IMPORTS_FROM` targets to disambiguate
cross-file `CALLS`) degrades PHP call resolution too. On a real Symfony/PHP
codebase this meant a core entity imported by 452 files showed 0 importers.

### Additional context

Root cause: `_extract_import` in `parser.py` has a per-language branch for
Python, JS/TS, Go, Rust, Java, C#, Scala, Dart, Julia and others, but none for
PHP, so PHP `use` statements fall through to the raw-text fallback
(`imports.append(text)`) and store the entire statement as the edge target.
There is also no PHP case in `_do_resolve_module` (Java resolves `import a.b.C;`
to the class file; PHP needs the analogous mapping).

A fix is proposed in #573: a PHP branch in `_extract_import` (recording the FQN,
handling `as` aliases, grouped `use A\{B, C}`, and `use function` / `use const`)
plus a PHP branch in `_do_resolve_module` that resolves the FQN to an absolute
`.php` path by walking up from the importing file, mirroring the Java resolver.

stefanws0:
## Linked issue

Closes #574.

Found while using the graph on a Symfony/PHP codebase: `importers_of` (and the
upstream side of impact-radius) returned **0** for PHP classes that are imported
by hundreds of files.

## What & why

PHP `use` statements had no branch in the parser's `_extract_import`, so they
fell through to the raw-text fallback (`imports.append(text)`) and stored the
**entire statement** as the `IMPORTS_FROM` edge target. For example,
`"use App\Domain\Entity\Job;"` was stored instead of the FQN
`App\Domain\Entity\Job`.

Consequences:

- `importers_of`, `tests_for`, and `inheritors_of` returned nothing for PHP
  classes, because the query matches edge targets by resolved file path, which
  the raw `use ...;` text never matches.
- The upstream side of `get_impact_radius` saw no PHP importers.
- `resolve_bare_call_targets` uses `IMPORTS_FROM` targets to disambiguate
  cross-file `CALLS`, so the unparsed targets silently degraded PHP call
  resolution too.

Every other FQN-style language already has an import branch, and Java
additionally resolves `import a.b.C;` to the class's `.java` file in
`_do_resolve_module`. This PR brings PHP to parity:

- **`_extract_import`**: new `php` branch recording the fully-qualified name of
  each imported symbol, handling `as` aliases (records the FQN, not the alias),
  grouped `use A\B\{C, D as E}` (prefix-joined), `use function` / `use const`,
  and a leading `\`.
- **`_do_resolve_module`**: new `php` branch mirroring Java. It converts `\` to
  `/`, appends `.php`, and walks up from the importing file to the source root.
  Vendor/global classes with no local file (such as `\Exception`) stay as the
  bare FQN, exactly like JDK imports.

No new dependencies, no schema changes, and no post-build resolver. The fix
lives entirely in the existing per-language parser paths, consistent with Java.

### Impact

In the codebase where this surfaced, `importers_of` for a core entity returned
**0** despite **452** files doing `use App\Domain\Entity\Job;` (confirmed via
grep). With the fix those `use` statements resolve to the entity's file. The
behavior is covered by a self-contained test (`TestPHPImportResolution`) on a
mini PSR-4 project.

## How it was tested

```bash
uv run pytest tests/test_multilang.py::TestPHPImportResolution -q   # 4 passed (new)
uv run pytest tests/ --tb=short -q                                  # 1388 passed, 1 skipped, 2 xpassed
uv run ruff check code_review_graph/                                # All checks passed!
uv run mypy code_review_graph/parser.py --ignore-missing-imports --no-strict-optional  # Success
```

Also verified end to end by building a graph for a mini PSR-4 project with the
patched code and confirming `query_graph("importers_of", <Job.php>)` returns the
two importing files and excludes a non-importing file. It returned nothing
before the change.

## Checklist

- [x] Docs updated where behavior changed (CHANGELOG)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,21 @@
 
 ## [Unreleased]
 
+### Fixed
+
+- **PHP `use` imports now resolve to files** (`importers_of`, impact radius,
+  call disambiguation): PHP `use` statements had no branch in the parser's
+  import extraction and fell through to the raw-text fallback, storing the whole
+  `use ...;` statement as the `IMPORTS_FROM` edge target (e.g.
+  `"use App\Domain\Entity\Job;"`). As a result `importers_of` / `tests_for` /
+  `inheritors_of` and the upstream side of `get_impact_radius` returned nothing
+  for PHP classes, and the unresolved targets also degraded cross-file `CALLS`
+  disambiguation in `resolve_bare_call_targets`. PHP imports are now recorded as
+  fully-qualified names (handling `as` aliases, grouped `use A\{B, C}`, and
+  `use function` / `use const`) and resolved to absolute `.php` paths by walking
+  up from the importing file, mirroring the existing Java resolver. Vendor/global
+  classes with no local file stay as the bare FQN.
+
 ## [2.3.6] - 2026-06-10
 
 **Community-response release.** Built from a full audit of every open PR,
--- a/code_review_graph/parser.py
+++ b/code_review_graph/parser.py
@@ -5531,6 +5531,24 @@ class CodeParser:
                         break
                     current = current.parent
 
+        elif language == "php":
+            # ``use App\Domain\Entity\Job;`` — convert namespace separators to
+            # a relative path and walk up from the caller's directory to find
+            # the file, mirroring the Java resolver. PSR-4 layouts where a
+            # namespace segment maps to a real directory (e.g. ``App\Foo`` ->
+            # ``.../App/Foo``) resolve; vendor/global classes (``\Exception``)
+            # and ``use function`` / ``use const`` targets with no matching
+            # file stay unresolved and keep the bare FQN, like JDK imports.
+            rel_path = module.replace("\\", "/").lstrip("/") + ".php"
+            current = caller_dir
+            while True:
+                target = current / rel_path
+                if target.is_file():
+                    return str(target.resolve())
+                if current == current.parent:
+                    break
+                current = current.parent
+
         return None
 
     def _find_dart_pubspec_root(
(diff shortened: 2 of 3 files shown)


itxaiohanglover:
Solid work! The PHP use import resolver is comprehensive — handling namespace separators, as aliases, grouped use, and use function/const. The walk-up directory traversal mirroring the Java resolver is the right approach. The CHANGELOG entry is very well-documented too.

stefanws0:
> Solid work! The PHP use import resolver is comprehensive — handling namespace separators, as aliases, grouped use, and use function/const. The walk-up directory traversal mirroring the Java resolver is the right approach. The CHANGELOG entry is very well-documented too.

Is there something I still need to do here, or is it now in a process where I just need to wait for it to be further reviewed?

tirth8205:
Thank you — and to answer your earlier question, nothing more is needed from you. The complete PHP `use` resolver (including grouped imports, aliases, functions/constants, and local-file resolution) and its tests were carried into #624 with your authorship and are now merged. This unchanged PR is fully superseded, so I’m closing the duplicate.

PR #578, opened 2026-06-24
fix(install): make JSONC sanitizer string-aware (#553)

### LINKED ISSUE #553: fix(install): trailing-comma regex breaks on commas inside string values

dlwjdghkszkf-png:
Fixes #553.

## Problem

`install_platform_configs` reduces an existing JSONC config to strict JSON before merging, using two regexes that don't track string boundaries:

```python
re.sub(r'//.*?$', '', raw, flags=re.MULTILINE)
re.sub(r',(\s*[}\]])', r'\1', stripped)
```

Both corrupt string *values*:

- the `//` stripper truncates a `"https://..."` URL at the `//`
- the trailing-comma stripper deletes a comma inside `"foo, bar"` when the next non-space char is `}` or `]`

The damaged text then fails `json.loads`, or — worse — parses to silently wrong data, corrupting the user's config on write-back.

## Fix

Replace both regexes with `_strip_jsonc()`: a single character-walking pass that tracks in-string state (honoring `\` escapes) and only removes `//` + `/* */` comments and trailing commas in **structural** position. Content inside string values is preserved verbatim. The now-unused `import re` is dropped.

## Tests

New `TestStripJsonc` (10 cases): the #553 repro (comma inside a string before `}`), `https://` URL preservation, escaped quotes, in-string comment markers (`"x // y"`, `"p /* q */ r"`), and real comment / trailing-comma removal.

Validation:

- `pytest tests/test_skills.py` → **146 passed** (10 new)
- `mypy code_review_graph/skills.py` → clean
- `ruff check code_review_graph/skills.py` → clean

### DIFF
--- a/code_review_graph/skills.py
+++ b/code_review_graph/skills.py
@@ -12,7 +12,6 @@ import json
 import logging
 import os
 import platform
-import re
 import shutil
 import stat
 import sys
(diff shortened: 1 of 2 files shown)


itxaiohanglover:
This is solid — the character-by-character walk with string state tracking is the right approach. I had opened #571 with a simpler version that only handled trailing commas, but this covers both comments and commas in one pass. Just closed mine in favor of this.

The `_skip_comment` helper is clean. One edge case: unterminated `/*` block comments will run off the end — the `idx + 2` could overshoot. Probably fine since it's best-effort, but worth a note.

tirth8205:
Thank you — the string-aware JSONC scanner and its regressions were rebased into #624 with your authorship and are now merged. We retained that stronger scanner through the later config conflict resolution and added preservation coverage for invalid nested config shapes. This unchanged PR is fully superseded, so I’m closing the duplicate.

PR #580, opened 2026-06-25
parser: tag CALLS edges inside dead guards with reachable flag

### LINKED ISSUE #576: CALLS edges are emitted for calls inside statically-dead code (`if False:`, `if TYPE_CHECKING:`, `#if 0`)
## Summary

The graph indexes function calls that occur inside statically-unreachable code as ordinary live CALLS edges. Because the parse is structural (Tree-sitter, no reachability analysis and no C preprocessor evaluation), downstream consumers cannot distinguish a call that can execute from one that never will.

## Reproduction (Python)

```python
# consumer.py
from api import process_request

def live_caller():
    process_request()          # line 5

def dead_caller():
    if False:                  # statically dead
        process_request()      # line 10 -- never executes
```

```bash
code-review-graph build --skip-flows
```

Resulting edges (both present):

```
CALLS  consumer.py::dead_caller -> process_request   line 10
CALLS  consumer.py::live_caller -> process_request   line 5
```

The `dead_caller` edge should not be presented as a live call.

## Reproduction (C/C++)

A call inside a `#if 0 ... #endif` block is likewise emitted as a live CALLS edge (observed on a real C++ project: a call inside a `#if 0` block was indexed identically to the active code path).

## Impact

Any consumer doing impact analysis / caller discovery over CALLS edges gets false positives from dead code: "symbol X is called at `<dead site>`", when that site can never run.

## Suggestion

Either (a) skip emitting CALLS edges whose call expression is inside an obviously-dead construct, or (b) tag such edges (e.g. `reachable=0` / a "dead-code" `confidence_tier`) so consumers can filter. The common cheap-to-detect idioms cover most of the value:

- **Python:** `if False:` / `if 0:` / `if TYPE_CHECKING:` ; code after an unconditional `return`/`raise`/`break`/`continue`
- **C/C++:** `#if 0 ... #endif` (and `#ifdef` of an undefined macro)

Full reachability is undecidable and not expected; handling these common static-dead idioms would remove the bulk of the false positives.

HouMinXi:
## Linked issue

Closes #576

## What & why

Calls inside `if False:` and `if TYPE_CHECKING:` blocks are statically
unreachable, but the parser emits CALLS edges for them anyway.  This
causes false positives in dead-code detection (`refactor_tool` mode
`dead_code`) and inflates impact-radius results.

This PR adds a lightweight ancestor-walk helper
`_is_in_static_dead_guard()` that detects these two common Python
idioms and sets `extra["reachable"] = False` on the affected CALLS
edges.  Live edges omit the key entirely, so:

- Existing `graph.db` files remain forward-compatible (absent key =
  live).
- Downstream consumers can filter with:
  `json_extract(extra, '$.reachable') IS NULL OR
   json_extract(extra, '$.reachable') != 0`

### Scope

- **Covered (v1):** All Tree-sitter languages routed through
  `_extract_calls` (Python, JS/TS, Go, Rust, C, C++, Java, etc.).
- **Not covered:** Bespoke handlers (ReScript, Dart regex path,
  Elixir), C preprocessor guards (`#if 0`).  These are left for
  follow-up.

### Design choice: ancestor walk vs threaded flag

Chose ancestor walk (approach b from #576 discussion) over threading
a flag through `_extract_from_tree` because it requires zero signature
changes, keeps the diff minimal, and is O(scope-depth) per call node
-- negligible overhead.

## How it was tested

```bash
uv run pytest tests/test_parser.py -v          # 104 passed
uv run pytest tests/ --tb=short -q             # 1385 passed, 0 failed
uv run ruff check code_review_graph/parser.py  # All checks passed
```

**Bug-inject verification:** neutralized the guard check (`if False
and ...`) -- test failed on `dead_false_call` edge missing the flag.
Restored -- test passed.

**Real-path smoke test:** parsed a Python file with `if False:` and
`if TYPE_CHECKING:` calls outside the test suite; confirmed live calls
show `reachable=ABSENT` and dead calls show `reachable=False`.

## Checklist

- [ ] Type checking passes (mypy not run locally; CI will verify)
- [ ] Docs updated where behavior changed (no user-facing API change)

### DIFF
--- a/code_review_graph/parser.py
+++ b/code_review_graph/parser.py
@@ -4579,11 +4579,13 @@ class CodeParser:
         """Walk ancestors to detect statically-dead guards.
 
         Returns True when ``node`` sits inside an ``if_statement`` whose
-        condition is ``False`` (the literal) or ``TYPE_CHECKING`` (the
-        ``typing`` sentinel).  This covers the two most common Python
-        idioms for compile-time-only code:
+        condition is ``False`` (the literal), the integer ``0``, or
+        ``TYPE_CHECKING`` (the ``typing`` sentinel).  This covers the
+        most common Python idioms for compile-time-only code:
 
             if False:
+                ...
+            if 0:
                 ...
             if TYPE_CHECKING:
                 ...
@@ -4605,6 +4607,12 @@ class CodeParser:
                 if cond is not None:
                     # ``if False:``
                     if cond.type == "false":
+                        return True
+                    # ``if 0:``
+                    if (
+                        cond.type == "integer"
+                        and cond.text == b"0"
+                    ):
                         return True
                     # ``if TYPE_CHECKING:``
                     if (
--- a/code_review_graph/parser.py
+++ b/code_review_graph/parser.py
@@ -4574,6 +4574,47 @@ class CodeParser:
                 line=child.start_point[0] + 1,
             ))
 
+    @staticmethod
+    def _is_in_static_dead_guard(node) -> bool:
+        """Walk ancestors to detect statically-dead guards.
+
+        Returns True when ``node`` sits inside an ``if_statement`` whose
+        condition is ``False`` (the literal) or ``TYPE_CHECKING`` (the
+        ``typing`` sentinel).  This covers the two most common Python
+        idioms for compile-time-only code:
+
+            if False:
+                ...
+            if TYPE_CHECKING:
+                ...
+
+        The walk stops at function/class/module boundaries so a dead
+        guard in an outer scope does not leak into nested definitions.
+        """
+        cursor = node.parent
+        while cursor is not None:
+            ntype = cursor.type
+            # Stop at scope boundaries -- a dead guard in an outer
+            # function does not make an inner function's calls dead.
+            if ntype in (
+                "function_definition", "class_definition", "module",
+            ):
+                break
+            if ntype == "if_statement":
+                cond = cursor.child_by_field_name("condition")
+                if cond is not None:
+                    # ``if False:``
+                    if cond.type == "false":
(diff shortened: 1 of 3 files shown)


itxaiohanglover:
Nice implementation! I reported #576 and this is exactly what I had in mind. The scope boundary check (stopping at function/class/module) is the right call — a dead guard in an outer scope shouldn't propagate to nested definitions. Tagging with `reachable: False` rather than removing edges is smart too, keeps the graph complete for downstream consumers.

One small thing: might be worth handling `if 0:` as well? It's less common but I've seen it in C-extension bindings. Not a blocker though.

HouMinXi:
Good catch -- added `if 0:` support in ded4913. Tree-sitter parses `0` as an `integer` node with `text == b"0"`, so the check mirrors the existing `false` node pattern. Test fixture and assertions updated accordingly.

HouMinXi:
Hello Folks
Any suggestion or comments for this PR?

tirth8205:
Thanks, including for the if 0 case. This does not fix #576 end-to-end because every caller, impact, flow, and dead-code consumer still traverses CALLS edges without checking extra.reachable.

Branch semantics are also wrong: a call in the live else branch of if False is marked unreachable, while a nested definition declared under a dead guard is marked live. The code misses typing.TYPE_CHECKING and aliases but false-positives on an unrelated local TYPE_CHECKING name.

Please either omit statically dead CALLS edges or introduce one central live-edge predicate used by all downstream consumers. Add end-to-end callers/impact/dead-code tests for else/elif, nested definitions, qualified/aliased typing sentinels, and false-positive controls. Issue #576 stays open until those product paths are correct.

PR #581, opened 2026-06-25
Fixed a bug

woshigelaji:
Fixed a bug where daemon wrote an invalid TOML (Invalid hex value) when saving configurations on Windows due to unescaped backslash paths, causing configuration reloading.

### DIFF
--- a/code_review_graph/daemon.py
+++ b/code_review_graph/daemon.py
@@ -164,22 +164,57 @@ def load_config(path: Path | None = None) -> DaemonConfig:
 # ---------------------------------------------------------------------------
 
 
+def _toml_escape_string(value: str) -> str:
+    """Escape *value* so it can be embedded in a TOML basic (double-quoted) string.
+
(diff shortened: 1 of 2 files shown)


tirth8205:
Closing as superseded by the independently implemented Windows daemon fix merged in #632 (merge commit 94da74e1dc9158840db6617b399e04579dec386a).\n\nExact overlap:\n- PR #581 escapes backslashes, double quotes, and U+0000–U+001F control characters before writing daemon TOML.\n- Current main now applies the same required TOML escaping to session_name, log_dir, repo path, and alias.\n- The merged implementation is slightly stricter: it also escapes U+007F, which TOML forbids unescaped.\n- Current regression tests parse Windows backslash paths and quoted values through tomllib, and the dedicated Windows CI job passed after merge.\n\nBecause the production fix was implemented independently through #632, no code from #581 was copied and I am not claiming contributor attribution for the merged implementation. The reported bug itself is resolved on main, so there is no remaining change to merge from this branch.

PR #582, opened 2026-06-25
fix(csharp): emit INHERITS edges from base_list clause

Saeed-Nourizadeh:
## Problem

C# classes produce **zero `INHERITS` edges**. `_get_bases()` handles C# in the shared `java`/`kotlin`/`csharp` branch, which only looks for `superclass` / `super_interfaces` / `delegation_specifier` nodes. The **tree-sitter-csharp** grammar emits none of those — it wraps a class's `: Base, IFace` clause in a single **`base_list`** node. So the branch never matches and C# inheritance is lost: `inheritors_of` / `get_impact_radius` return empty for `.cs` files.

Reproduced on a real 1,527-file / 440K-LoC C# codebase (SplendidCRM): **0** INHERITS edges before, **1,373** after this fix (top bases that were previously invisible: framework base classes like `SplendidControl` with 440 subclasses, page bases, `System.Web.Services.WebService`, etc.).

## Fix

Give C# its own branch in `_get_bases()` that drills into `base_list` for the named base-type children — `identifier`, `qualified_name`, `generic_name` — and stores the bare type name (skipping the leading `:` and commas). Kotlin keeps its existing handling; Java is unaffected. This mirrors the approach in #278 (Java) and the C++ `base_class_clause` branch.

```
public class CachedRepo : InMemoryRepo, IRepository   ->  INHERITS: InMemoryRepo, IRepository
public class DisposableService : System.IDisposable   ->  INHERITS: System.IDisposable  (qualified_name preserved)
public class UserList : List<User>                     ->  INHERITS: List<User>          (generic_name preserved)
```

## Tests

- Extends `tests/fixtures/Sample.cs` with extends + implements, a qualified base (`System.IDisposable`) and a generic base (`List<User>`).
- Adds `test_finds_inheritance` and `test_inheritance_target_is_bare_name` to `TestCSharpParsing` (asserts targets are bare type names — no leading `:`, no comma-joined clauses).
- `pytest tests/test_multilang.py -k "CSharp or Java or Kotlin"` → 23 passed (no regression). `ruff check` clean.

### DIFF
--- a/code_review_graph/parser.py
+++ b/code_review_graph/parser.py
@@ -6101,11 +6101,20 @@ class CodeParser:
             # superclass/super_interfaces nodes the C# grammar never emits,
             # so C# types produced zero INHERITS edges and inheritors_of /
             # get_impact_radius returned empty for .cs files.
+            #
+            # Two things the base_list contains that are NOT base types:
+            #  * `enum E : byte` — the base_list holds the enum's *underlying
+            #    type*, not a base. Enums never inherit, so emit nothing.
+            #  * `class C(int x) : Base(x)` — the primary-constructor arguments
+            #    `(x)` appear as an `argument_list` sibling of the base type
+            #    (and inside `primary_constructor_base_type` for records).
+            if node.type == "enum_declaration":
+                return bases
             for child in node.children:
                 if child.type != "base_list":
                     continue
                 for sub in child.children:
-                    if not sub.is_named:
+                    if not sub.is_named or sub.type == "argument_list":
                         continue
                     if sub.type == "primary_constructor_base_type":
                         # positional record: `record R(...) : Base(args)` —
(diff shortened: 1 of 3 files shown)


tirth8205:
Merged the corrected C# inheritance behavior through maintainer PR #628 (merge commit 28ea6a2), with @Saeed-Nourizadeh credited as co-author in commit 8c06514.

The port was rebuilt on current main, reproduced the zero-INHERITS failure first, then covered classes, interfaces, records, structs, qualified/generic bases, generic constraints, primary-constructor arguments, and enum storage types. C#/Java/Kotlin tests passed 40/40; the non-daemon suite passed 1,427 tests with 2 expected xpasses; Ruff and the complete GitHub Python 3.10–3.13 matrix are green. Closing this draft source PR as incorporated.

PR #586, opened 2026-06-27
fix: prevent silent node loss from qualified_name collisions

### LINKED ISSUE #585: Silent node loss: same-file qualified_name collisions are dropped on insert
## Summary

Two same-named symbols in one file collapse into a single node on insert, so the graph silently drops the earlier definitions. A build reports more nodes than the database actually stores, and caller/callee analysis on those names resolves to a single merged node.

## Root cause

`qualified_name` is the stable identity key, built by `GraphStore._make_qualified` as `file::name` or `file::parent.name`. That key is not unique per symbol: methods of different object literals, or repeated generated helpers, produce the same key. `upsert_node` inserts with `ON CONFLICT(qualified_name) DO UPDATE`, so colliding rows collapse last writer wins and earlier definitions are lost.

The discrepancy is also visible in the counts: `full_build` reports `total_nodes` summed in memory from the parsed node lists, while `status` reads `SELECT COUNT(*)`. The gap is exactly the number of dropped nodes.

## Reproduction

```ts
// db.ts
export const personas = { getById(id) { return id } };
export const users = { getById(id) { return id } };
export const cart = { getById(id) { return id } };
```

```
$ code-review-graph build
Full build: 1 files, 4 nodes, 3 edges (postprocess=full)
$ code-review-graph status
Nodes: 2
```

Build claims 4 nodes (one File plus three `getById`); the database holds 2 (the three `getById` collapse into one). The same pattern shows up in real object modules and in generated files with repeated helper names at module scope.

## Proposed fix

Deduplicate per file in the batch store functions (`store_file_nodes_edges`, `store_file_batch`), which already receive the whole file's node list. Keep the first occurrence at its bare key so existing edges still resolve to it, and suffix later collisions with `:L<line_start>`, which is always unique because two definitions cannot share a start line. Unique nodes keep byte identical keys, so embeddings and incremental updates are unaffected.

I also propose surfacing the disambiguation so collisions stop being invisible: a `find_disambiguated_nodes()` helper, a `disambiguated_nodes` field in the build result, and a one line CLI summary.

### Convention question

Does the `:L<line_start>` suffix work for you, or would you prefer a different marker for the second and later collisions? One honest caveat: a collided node's identity shifts if its own start line moves. That affects only ambiguous duplicates, never the unique majority.

A PR implementing this (two commits, tests, suite green) is ready to open against `main`.

CodeBlackwell:
Fixes #585.

## Problem

`qualified_name` is the graph's stable identity key, built as `file::name` or `file::parent.name`. It is not unique per symbol: same-named symbols in one file (methods of different object literals, repeated generated helpers) produce the same key. `upsert_node` inserts with `ON CONFLICT(qualified_name) DO UPDATE`, so colliding rows collapse last writer wins and earlier definitions are silently dropped. A build then reports more nodes than the database stores, and caller/callee analysis on those names points at a single merged node.

## Changes

**Commit 1 — disambiguate colliding qualified_names.** Deduplicate per file in `store_file_nodes_edges` and `store_file_batch`, which already hold the whole file's node list. The first occurrence keeps its bare key so existing edges still resolve to it; later collisions are suffixed with `:L<line_start>`, which is always unique because two definitions cannot share a start line. Unique nodes keep byte identical keys, so embeddings and incremental updates are unaffected.

**Commit 2 — report disambiguated names.** Adds `GraphStore.find_disambiguated_nodes()`, a `disambiguated_nodes` field in the `full_build` and `incremental_update` results, a log line, and a one line CLI summary after build/update. Counts are not rewritten: now that collisions are kept rather than dropped, the reported node total and the stored `COUNT(*)` already agree.

## Verification

```
# before: build claims 4 nodes, status reports 2
# after:
$ code-review-graph build
Full build: 1 files, 4 nodes, 3 edges (postprocess=full)
Disambiguated 2 duplicate symbol name(s): .../db.ts::getById:L2, .../db.ts::getById:L3
$ code-review-graph status
Nodes: 4
```

New `tests/test_collisions.py` covers: both same-named methods persist, three module scope duplicates persist, an edge to the bare key still resolves to the first definition, the disambiguation report, and that unique names keep bare keys. Full suite green (1389 passed), ruff and mypy clean.

## Caveat

A collided node's identity shifts if its own `line_start` moves. This affects only ambiguous duplicates, never the unique majority. Open to a different marker convention if you prefer (see #585).

### DIFF
--- a/code_review_graph/cli.py
+++ b/code_review_graph/cli.py
@@ -88,6 +88,16 @@ def _supports_color() -> bool:
     return sys.stdout.isatty()
 
 
+def _print_disambiguated(result: dict, limit: int = 10) -> None:
+    """Report duplicate symbol names that were disambiguated during a build."""
+    dups = result.get("disambiguated_nodes") or []
+    if not dups:
+        return
+    shown = ", ".join(dups[:limit])
+    more = f" (+{len(dups) - limit} more)" if len(dups) > limit else ""
+    print(f"Disambiguated {len(dups)} duplicate symbol name(s): {shown}{more}")
+
+
 def _print_banner() -> None:
     """Print the startup banner with graph art and available commands."""
     color = _supports_color()
@@ -994,6 +1004,7 @@ def main() -> None:
             nodes = result.get("total_nodes", 0)
             edges = result.get("total_edges", 0)
             print(f"Full build: {parsed} files, {nodes} nodes, {edges} edges (postprocess={pp})")
+            _print_disambiguated(result)
             if result.get("errors"):
                 print(f"Errors: {len(result['errors'])}")
 
@@ -1019,6 +1030,7 @@ def main() -> None:
                 f"{nodes} nodes, {edges} edges"
                 f" (postprocess={pp})"
             )
+            _print_disambiguated(result)
 
             # --brief: append a one-line change-impact summary with the same
             # estimated context-savings approximation that detect-changes uses.
--- a/code_review_graph/graph.py
+++ b/code_review_graph/graph.py
@@ -186,10 +186,19 @@ class GraphStore:
 
     # --- Write operations ---
 
-    def upsert_node(self, node: NodeInfo, file_hash: str = "") -> int:
-        """Insert or update a node. Returns the node ID."""
+    def upsert_node(
+        self, node: NodeInfo, file_hash: str = "", qualified: str | None = None
+    ) -> int:
+        """Insert or update a node. Returns the node ID.
+
+        Pass ``qualified`` to override the computed key — used by the batch store
+        functions to disambiguate same-named symbols in one file (see
+        ``_qualified_names_for_file``). Without it, collisions would collapse
+        under ``ON CONFLICT(qualified_name) DO UPDATE`` and silently drop nodes.
+        """
         now = time.time()
-        qualified = self._make_qualified(node)
+        if qualified is None:
+            qualified = self._make_qualified(node)
         extra = json.dumps(node.extra) if node.extra else "{}"
 
         self._conn.execute(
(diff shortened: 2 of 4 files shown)


CodeBlackwell:
Quick status update: this fix is still current and independent of #614 (the CSS/SCSS feature PR) — the two don't overlap, so this one stays open on its own merits. The branch sits directly on today's main tip and GitHub reports it mergeable with no conflicts. Happy to adjust anything if you'd like changes.

tirth8205:
Thanks—this prevents colliding rows from being lost, but ambiguous calls are still attached to the first definition. That can produce a confident but incorrect edge. The `:L<line>` identity also changes when lines move.

Please define a stable identity or migration/rebuild strategy, then make calls resolve to the correct duplicate when evidence exists—or remain explicitly ambiguous when it does not. Please add end-to-end tests for callers of each overload/duplicate and an incremental edit that shifts line numbers.

tirth8205:
Thanks for the contribution. I rechecked exact head `99b0a41` against current `main` and am closing this PR.

Adding `:L<line>` suffixes preserves colliding rows but not graph correctness. Parser-produced edges still target the bare qualified name, so every ambiguous call resolves to the first definition and later duplicates cannot receive callers; one submitted test explicitly asserts that behavior. Harmless line shifts also change identity, and direct `upsert_node()` calls retain the original collapsing behavior unless callers opt into the override. There is no end-to-end coverage for correct or explicitly ambiguous caller resolution, or for identity migration after an incremental line shift. GitHub reports the branch non-mergeable, and both workflows are `action_required` with zero jobs.

Issue #585 remains open for a stable parser/resolver-level identity design.

PR #589, opened 2026-06-30
fix(java): correct Lombok @RequiredArgsConstructor/@AllArgsConstructor DI semantics

YutenC:
## Problem

v2.3.3 introduced Spring DI resolution via `_emit_spring_field_injection()`, but the Lombok constructor annotation handling has two semantic bugs:

**1. `@RequiredArgsConstructor` misses `@NonNull` fields**

Lombok's `@RequiredArgsConstructor` generates a constructor for `final` fields **and** `@NonNull`-annotated non-final fields. The current implementation only checks `has_final`, so `@NonNull` fields silently produce no `INJECTS` edge.

**2. `@AllArgsConstructor` behaves like `@RequiredArgsConstructor`**

`@AllArgsConstructor` generates a constructor for *all* non-static fields regardless of `final` or `@NonNull`. The current code shares the same `has_lombok_constructor and has_final` guard, so non-final fields in an `@AllArgsConstructor` class are never resolved.

## Fix

Split `has_lombok_constructor: bool` into two flags and apply the correct Lombok field-selection rule for each annotation:

| Annotation | Fields included in generated constructor |
|---|---|
| `@RequiredArgsConstructor` | `final` **or** `@NonNull` (non-static) |
| `@AllArgsConstructor` | All non-static fields |

A new `injection_type` value `"constructor_lombok_all"` is introduced for `@AllArgsConstructor` edges. The existing `"constructor_lombok"` value is preserved for `@RequiredArgsConstructor` — no breaking change to existing graphs.

## Changes

- `code_review_graph/parser.py` — `_emit_spring_injections()` and `_emit_spring_field_injection()`
- `tests/fixtures/SpringDI.java` — add `PaymentService` (`@RequiredArgsConstructor` + `@NonNull` non-final field) and `ReportService` (`@AllArgsConstructor` with non-final fields)
- `tests/test_multilang.py` — three new tests, updated edge count assertion

### DIFF
--- a/code_review_graph/parser.py
+++ b/code_review_graph/parser.py
@@ -3885,17 +3885,17 @@ class CodeParser:
     ) -> None:
         """Emit INJECTS edges for Spring DI injection points in a Java class.
 
-        Handles three patterns:
+        Handles four patterns:
         - @Autowired / @Inject / @Resource field injection
         - @Autowired constructor injection
-        - Lombok @RequiredArgsConstructor / @AllArgsConstructor with final fields
+        - Lombok @RequiredArgsConstructor: final + @NonNull non-static fields
+        - Lombok @AllArgsConstructor: all non-static fields
         """
         if language != "java":
             return
 
-        has_lombok_constructor = any(
-            a in _LOMBOK_CONSTRUCTOR_ANNOTATIONS for a in class_annotations
-        )
+        has_required_args = "RequiredArgsConstructor" in class_annotations
+        has_all_args = "AllArgsConstructor" in class_annotations
         qualified_source = self._qualify(class_name, file_path, None)
 
         # Find the class body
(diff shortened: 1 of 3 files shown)


tirth8205:
Thanks — the underlying Lombok bug is real, but this needs one correctness pass before merge.

@RequiredArgsConstructor covers uninitialized final fields and uninitialized @NonNull fields. The current final || NonNull test includes initialized fields incorrectly, and only one name is retained from a multi-declarator field such as Foo a, b.

Please emit per variable_declarator, track each initializer, and add tests for initialized final, initialized @NonNull, multiple declarators, and static fields. Rebase first; the current workflows stopped at action_required without executing jobs.

YutenC:
I noticed you merged #589–#591 with the review comments still open. Happy to follow up with fixes for the Lombok initialized-field and multi-declarator issues — or are you planning to address them on your end?

PR #590, opened 2026-06-30
feat(java): detect @Scheduled methods and emit TRIGGERS edges

YutenC:
## Problem

Spring `@Scheduled` methods are scheduler entry points with no static caller in the codebase. CRG previously had no way to discover them via graph queries — users had to fall back to `rg "@Scheduled"`.

## Solution

Emit `TRIGGERS` edges from virtual scheduler source nodes to annotated methods at parse time. No post-build resolver needed — schedule parameters are statically present in the annotation.

| `@Scheduled` param | Source node |
|---|---|
| `cron = "..."` | `SCHEDULER::cron` |
| `fixedRate = N` | `SCHEDULER::fixedRate` |
| `fixedDelay = N` | `SCHEDULER::fixedDelay` |
| (none / unrecognised) | `SCHEDULER::scheduled` |

Schedule metadata (cron expression, rate value, etc.) is stored in edge `extra` for traceability.

## Usage after this change

```
# Find all methods the scheduler triggers
query_graph(callers_of="SCHEDULER::cron")
query_graph(callers_of="SCHEDULER::fixedRate")
```

## Changes

- `code_review_graph/parser.py` — `_SPRING_SCHEDULED_ANNOTATIONS` constant + `_emit_scheduled_edges_from_method()` + hook in `_extract_functions`
- `tests/fixtures/ScheduledTasks.java` — fixture with cron / fixedRate / fixedDelay / plain methods
- `tests/test_multilang.py` — `TestSpringScheduledParsing` (8 assertions)

### DIFF
--- a/code_review_graph/parser.py
+++ b/code_review_graph/parser.py
@@ -518,6 +518,9 @@ _KAFKA_PRODUCER_TYPES = frozenset({
     "KafkaSender",
 })
 
+# Spring scheduling annotation
+_SPRING_SCHEDULED_ANNOTATIONS = frozenset({"Scheduled"})
+
 
 # ---------------------------------------------------------------------------
 # ReScript regex patterns and helpers (no tree-sitter grammar bundled)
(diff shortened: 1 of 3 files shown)


tirth8205:
Thanks — scheduler edges are useful, but the advertised query path does not work yet.

query_graph first resolves a stored node and callers_of/callees_of traverse only CALLS. This patch emits TRIGGERS edges from string-only SCHEDULER::* endpoints, creates no scheduler nodes, and adds no trigger-aware query. The example therefore returns not_found even though parser-only tests pass.

Please add stored virtual nodes plus triggers_of/triggered_by (or equivalent flow integration), test through GraphStore and query_graph, cover repeatable @Scheduled annotations and enum-valued timeUnit metadata, then rebase and run current CI.

PR #591, opened 2026-06-30
feat(java): detect Spring @EventListener/publishEvent() and emit HANDLES/PUBLISHES edges

YutenC:
## Problem

Spring Application Events create implicit caller-callee relationships invisible to static analysis: a publisher calls `publishEvent(new XxxEvent())` and Spring routes it to all `@EventListener` methods for that event type. CRG had no way to trace this path.

## Solution

Two-phase resolution following the existing Spring DI / Temporal patterns:

**Parse phase** (parser.py):
- `@EventListener` method → `HANDLES` edge to virtual `event:XxxEvent` node
- `publishEvent(new XxxEvent())` call → `PUBLISHES` edge to virtual `event:XxxEvent` node

Event type is inferred from (in priority order):
1. `@EventListener(OrderEvent.class)` — explicit annotation arg
2. `@EventListener(classes = {A.class, B.class})` — multi-type named arg
3. First method parameter type — implicit / no-arg form

**Post-build resolver** (event_resolver.py):
- Joins `PUBLISHES` and `HANDLES` edges by `event_type`
- Emits `CALLS` edges: publisher method → listener method

## Usage after this change

```
# Find all listeners for OrderPlacedEvent
query_graph(callers_of="event:OrderPlacedEvent")

# Find what events a publisher emits
query_graph(callees_of="OrderService.placeOrder")
```

## Changes

- `code_review_graph/parser.py` — 2 constants + 3 new methods + 2 hooks in `_extract_functions`
- `code_review_graph/event_resolver.py` — new post-build resolver
- `code_review_graph/incremental.py` — `_run_event_resolver` wired into build pipeline
- `tests/fixtures/SpringEvents.java` — fixture with publisher + 2 listener patterns
- `tests/test_multilang.py` — `TestSpringEventListenerParsing` (6 assertions)

### DIFF
--- a/code_review_graph/event_resolver.py
+++ b/code_review_graph/event_resolver.py
@@ -1,0 +1,125 @@
+"""Post-build Spring Application Event resolver.
+
+Connects publishEvent() call sites to @EventListener methods by matching
+on the event class name.
+
+Resolution chain:
+    publisher_method  →(PUBLISHES)→  event:XxxEvent
+    listener_method   →(HANDLES)→   event:XxxEvent
+    ⟹ emit CALLS edge: publisher_method → listener_method
+"""
+
+from __future__ import annotations
+
+import json
+import logging
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from .graph import GraphStore
+
+logger = logging.getLogger(__name__)
+
+
+def resolve_spring_events(store: GraphStore) -> dict:
+    """Emit CALLS edges from event publishers to matching @EventListener methods.
+
+    Safe to call multiple times — already-resolved edges are skipped via
+    extra.event_resolved flag.
+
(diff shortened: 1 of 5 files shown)


tirth8205:
Thanks — Spring application events would be valuable, but the post-build resolver is not functional.

The edges table requires updated_at, while the new INSERT OR IGNORE omits it; SQLite ignores every derived CALLS row even though calls_emitted increments. Current tests inspect parser edges only and never verify a stored publisher-to-listener call.

Please write through the graph-store API or include every required column, then add full-build and incremental tests. Listener add/remove/type changes must rebuild or delete derived calls rather than permanently skipping event_resolved publishers. Match events with import/package-qualified identity so same-named events do not cross-link. Rebase and run approved current CI.

PR #595, opened 2026-07-03
fix: escape backslashes, quotes, and control chars in daemon TOML serialization

SHudici:
## Problem

The daemon config writer serializes TOML by hand. Repo paths are embedded verbatim, so any value containing backslashes or quotes produces invalid TOML. On Windows every path has backslashes (`C:\Users\...`), which makes `tomllib.load` fail on the next read — `crg daemon` cannot round-trip its own config file.

## Fix

Escape backslashes and double quotes when writing string values, matching what `tomllib` expects to read back. Control characters (U+0000–U+001F, U+007F) are also escaped — named short escapes (`\b \t \n \f \r`) where TOML defines them, `\uXXXX` otherwise — since TOML basic strings may not contain them raw and a stray tab or newline in a path/name would corrupt the file the same way. No behavior change for values that never needed escaping.

## Testing

- New/updated tests in `tests/test_daemon.py` cover round-tripping Windows-style paths, quoted values, and control characters.
- Full suite passes on Windows 11 (Python 3.12) and is unaffected on POSIX (paths without backslashes never triggered the bug there, which is why CI never saw it).

One of a small series of PRs making the project fully usable on Windows contributor machines.

### DIFF
--- a/code_review_graph/daemon.py
+++ b/code_review_graph/daemon.py
@@ -164,6 +164,40 @@ def load_config(path: Path | None = None) -> DaemonConfig:
 # ---------------------------------------------------------------------------
 
 
+# TOML basic strings have named escapes for these; everything else in
+# the control range must use the \uXXXX form.
+_TOML_SHORT_ESCAPES = {
+    "\\": "\\\\",
+    '"': '\\"',
+    "\b": "\\b",
+    "\t": "\\t",
+    "\n": "\\n",
+    "\f": "\\f",
+    "\r": "\\r",
+}
+
+
+def _toml_str(value: object) -> str:
+    """Render *value* as a TOML basic string.
+
+    Backslashes and double quotes are escape characters in TOML basic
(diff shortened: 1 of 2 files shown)


tirth8205:
Thanks — the TOML bug and escaping implementation are correct, and the round-trip tests cover quotes, backslashes, and forbidden control characters. No repository workflow ran for this head, so it cannot merge without independent validation.

We are carrying it into a current-main Windows daemon bundle with #597 and #596, then running the focused serializer tests and a real windows-latest job. Your attribution will be preserved.

tirth8205:
Completed through replacement PR #632, merged as 94da74e1dc9158840db6617b399e04579dec386a. Your TOML-safe daemon serialization was ported onto current main with SHudici attribution. The replacement escapes Windows backslashes, quotes, named TOML escapes, and all forbidden control characters; serialize-to-parse regressions pass, and the new Windows daemon CI job passed on windows-latest. This stale source was closed only because the credited replacement is merged; no further contributor action is needed.

PR #596, opened 2026-07-03
test: make the suite pass on Windows

SHudici:
## Problem

On Windows the test suite reports over 200 teardown errors and several failures that have nothing to do with the code under test:

- `tempfile.NamedTemporaryFile(delete=False)` keeps the OS handle open; Windows then refuses `Path.unlink()` in teardown while `GraphStore` (or the fixture itself) still holds the file open. POSIX allows unlinking open files, so CI never sees this.
- `tests/test_skills.py` wrote through the real `Path.home()` and asserted POSIX exec bits, which fails on Windows runners.
- Two tests exercise POSIX-only branches (`os.kill` probing, executable-bit checks) that cannot pass on Windows by design.

## Fix

Test-only changes across 11 files:

- Close the `NamedTemporaryFile` handle immediately after creation (the file persists thanks to `delete=False`; `GraphStore` reopens it by path). 23 sites.
- Patch `Path.home()` in `test_skills.py` instead of touching the real home directory, and restore `store.close` via `patch.object` in `test_changes.py` so teardown always runs against the real method.
- `@pytest.mark.skipif(os.name != "posix")` on the two POSIX-only tests, with reasons.

## Testing

- Windows 11 (Python 3.12): suite goes from ~228 teardown errors to 0. Combined with the two Windows product fixes in this series (TOML escaping, PID liveness), the full suite is 1412 passed / 0 failed / 0 errors.
- POSIX behavior unchanged: closing the handle early is a no-op for the assertions, and the skipped tests still run on Linux/macOS CI.

### DIFF
--- a/tests/test_changes.py
+++ b/tests/test_changes.py
@@ -19,6 +19,7 @@ from code_review_graph.parser import EdgeInfo, NodeInfo
 class TestChanges:
     def setup_method(self):
         self.tmp = tempfile.NamedTemporaryFile(suffix=".db", delete=False)
+        self.tmp.close()  # close the OS handle so Windows can unlink it in teardown
         self.store = GraphStore(self.tmp.name)
 
     def teardown_method(self):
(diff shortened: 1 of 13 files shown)


tirth8205:
Thanks — closing temporary-file handles, patching Path.home(), and skipping genuinely POSIX-only assertions are the right fixes. The live test tree has gained additional temporary database fixtures since this branch, so the old 23-site sweep is incomplete, and no Windows job ran.

We are redoing the sweep on current main, retaining only justified platform skips, and validating the complete focused suite on windows-latest. Your attribution will be preserved.

tirth8205:
Completed through replacement PR #632, merged as 94da74e1dc9158840db6617b399e04579dec386a. Your Windows file-handle fixes were ported onto current main with SHudici attribution, expanded to all 24 current temporary GraphStore fixtures, and protected by an AST regression guard. The dedicated windows-latest job ran the daemon plus all touched resource-handle modules and passed. I am closing this stale source branch only because the credited replacement is merged; no further contributor action is needed.

Issue #511, opened 2026-05-24
[Windows] uvx code-review-graph daemon status raises WinError 87 even when watcher PIDs are alive


Mr-B-1:
## Summary

On Windows, `uvx code-review-graph@latest daemon status` consistently raises `OSError: [WinError 87] The parameter is incorrect` even when the daemon's watcher PIDs are alive and the daemon is otherwise healthy. The status command is therefore unreliable as a health check on Windows.

## Repro

1. Start the daemon: `uvx code-review-graph@latest daemon start` (succeeds; watchers spawn)
2. Verify watchers are alive: PowerShell `Get-Process -Id <pid>` returns the process for each PID in `~/.code-review-graph/daemon-state.json`
3. Query status: `uvx code-review-graph@latest daemon status`
4. Observed: `OSError: [WinError 87] The parameter is incorrect`
5. Expected: status report of running daemon

## Root cause hypothesis

This appears to be a known CPython asyncio Proactor handle-close issue family:
- [cpython#149388 — `_ProactorBasePipeTransport._call_connection_lost` leaks `OSError` from `PipeHandle.close()` on Windows](https://github.com/python/cpython/issues/149388)
- [python.org Async-SIG: Proactor WinError 87 error](https://discuss.python.org/t/proactor-winerror-87-error/3574)

`PipeHandle.close()` is not idempotent at the Win32 level — it calls `_winapi.CloseHandle` unconditionally, and if the underlying handle has already been closed between `read_handle` and `CloseHandle`, it raises. The Proactor event loop's `_call_connection_lost` callback then propagates the exception.

## Suggested fixes (any of)

1. **Wrap with `try/except OSError`** at the daemon-status RPC's connection-close site
2. **Pin `WindowsSelectorEventLoopPolicy`** at process start on Windows:
   ```python
   if sys.platform == "win32":
       asyncio.set_event_loop_policy(asyncio.WindowsSelectorEventLoopPolicy())
   ```
3. **Guard the close path** with an explicit `is_closed` check before calling `_winapi.CloseHandle`

(1) is the lowest-risk surgical fix. (2) is broader but may have other implications for any subprocess code. (3) is the most correct but requires deeper plumbing.

## Current workaround

[...]

PR #597, opened 2026-07-03
fix: request SYNCHRONIZE access in the Windows PID liveness check

SHudici:
## Problem

`_pid_alive_windows` opens the process with `PROCESS_QUERY_LIMITED_INFORMATION` only, then calls `WaitForSingleObject` on the handle. That access right does not include `SYNCHRONIZE`, so `WaitForSingleObject` always returns `WAIT_FAILED` (last error 5, access denied) — for live and dead processes alike. The code treated any non-`WAIT_OBJECT_0` result as "alive", so dead PIDs read as alive: stale daemons were never detected, and `crg daemon start` refused to start after a crash left a PID file behind.

Verified with a direct Win32 probe: waiting on a handle opened without `SYNCHRONIZE` fails with error 5; adding `SYNCHRONIZE` makes the wait report dead processes correctly.

## Fix

- Open with `PROCESS_QUERY_LIMITED_INFORMATION | SYNCHRONIZE`.
- Treat `WAIT_FAILED` explicitly: log and presume alive (fail-safe — never kill or replace a daemon we cannot inspect), instead of silently conflating it with "still running".
- Declare `argtypes`/`restype` (via `ctypes.wintypes`) for `OpenProcess`, `WaitForSingleObject`, and `CloseHandle`. Without them ctypes defaults every return to C `int`: 64-bit `HANDLE`s can truncate, and `WaitForSingleObject`'s `WAIT_FAILED` (`0xFFFFFFFF`) comes back as `-1`, which never compares equal to the unsigned constant — so the explicit `WAIT_FAILED` branch above could never fire against the real API.
- Name the Win32 constants instead of inlining magic numbers.

## Testing

- `tests/test_daemon.py`: the fake kernel32 now asserts the exact access mask, and a new test covers the `WAIT_FAILED` → presumed-alive path with handle cleanup. Prototypes are declared only where the real DLL is constructed, so the injected-fake seam the tests use is unchanged.
- On Windows 11 the previously-failing `test_pid_alive_for_dead_pid` / `test_is_daemon_running_alive` pass against the real API. POSIX path untouched.

One of a small series of PRs making the project fully usable on Windows contributor machines.

### DIFF
--- a/code_review_graph/daemon.py
+++ b/code_review_graph/daemon.py
@@ -316,8 +316,10 @@ def clear_pid(path: Path | None = None) -> None:
 
 # Win32 constants for the OpenProcess-based liveness check (#511).
 _PROCESS_QUERY_LIMITED_INFORMATION = 0x1000
+_SYNCHRONIZE = 0x00100000
 _ERROR_ACCESS_DENIED = 5
 _WAIT_OBJECT_0 = 0x0
+_WAIT_FAILED = 0xFFFFFFFF
 
 
 def _pid_alive_windows(
@@ -327,6 +329,11 @@ def _pid_alive_windows(
 ) -> bool:
     """Win32 PID liveness check via OpenProcess/WaitForSingleObject.
 
+    The access mask must include SYNCHRONIZE: a handle opened with only
+    PROCESS_QUERY_LIMITED_INFORMATION cannot be waited on, so
+    WaitForSingleObject returns WAIT_FAILED (ERROR_ACCESS_DENIED) and
+    every exited process reads as alive.
+
     The kernel32 interface is injected so tests can drive handle/wait
     outcomes on non-Windows platforms. *get_last_error* defaults to
     ``kernel32.GetLastError``; the real caller passes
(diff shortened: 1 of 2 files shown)


tirth8205:
Thanks — the root cause is confirmed: the process handle passed to WaitForSingleObject needs SYNCHRONIZE, and explicit ctypes signatures are required to avoid 64-bit handle and WAIT_FAILED conversion bugs. This corrects the incomplete Windows fix from #541/#511.

We are porting it with #595 and #596 and will run the live/dead-process tests against the real Win32 API on windows-latest before merge, preserving your attribution. Issue #511 is being reopened until that real Windows validation passes.

tirth8205:
Completed through replacement PR #632, merged as 94da74e1dc9158840db6617b399e04579dec386a. Your Win32 PID-liveness correction was ported onto current main with SHudici attribution: OpenProcess now requests SYNCHRONIZE, ctypes uses pointer-safe HANDLE/DWORD prototypes, WAIT_FAILED is handled explicitly, and handles always close. Unit regressions and the real windows-latest daemon job passed. I am closing this stale source branch only because the credited replacement is merged; no further contributor action is needed.

PR #598, opened 2026-07-03
fix: read TESTED_BY edges in the parser's direction everywhere

SHudici:
## Problem

The parser emits `TESTED_BY` edges as **production → test** (`source` = the code under test, `target` = the test function). Several consumers read the edge in the opposite direction, so they silently return nothing:

- `query_graph(pattern="tests_for")` looked for incoming edges on the production node and returned the wrong endpoint.
- `graph.get_transitive_tests` queried `target_qualified = ?` for the production node in all three lookup blocks (direct, bare-name fallback, transitive).
- `changes.py` counted "tested" nodes from incoming edges → `tests_for` context in change reviews was empty.
- `enrich.py`'s Tests block and `refactor.py`'s test-reference augmentation read the reverse direction as well.

Some call sites already read the correct direction (`build.py`, `analysis.py`, `review.py`, `communities.py`), which is how the inconsistency stayed invisible: half the features worked, half quietly returned empty results.

## Fix

Read `TESTED_BY` by `source_qualified` (production) and take `target_qualified` (test) everywhere, matching the parser. Test seeds that encoded the inverted direction were flipped to the parser's real output, and two end-to-end regression tests pin the direction:

- `test_tested_by_edges_point_from_production_to_test` (parser level)
- `TestQueryGraphTestsFor` (tool level, with a test name that deliberately defeats the name-matching fallback so only the edge lookup can find it)

## Notes

`resolve_bare_call_targets` upgrades bare `CALLS` targets to qualified names but does not touch `TESTED_BY` sources that inherited a bare name from an unresolved cross-file call. `refactor.py` keeps its plausibility-checked bare-name fallback for exactly that case. Two natural follow-up PRs if you want them: extending bare-name resolution to `TESTED_BY`, and a canonical `GraphStore.get_tests_of()` accessor so the direction knowledge lives in one place instead of each consumer's SQL.

## Testing

Full suite: 1412 passed / 0 failed on Windows 11; the new regression tests fail on main and pass here.

### DIFF
--- a/code_review_graph/changes.py
+++ b/code_review_graph/changes.py
@@ -352,7 +352,7 @@ def analyze_changes(
     for node in changed_funcs:
         if node.is_test:
             continue
-        tested = store.get_edges_by_target(node.qualified_name)
+        tested = store.get_edges_by_source(node.qualified_name)
         if not any(e.kind == "TESTED_BY" for e in tested):
             test_gaps.append({
                 "name": _sanitize_name(node.name),
--- a/code_review_graph/communities.py
+++ b/code_review_graph/communities.py
@@ -830,8 +830,8 @@ def get_architecture_overview(store: GraphStore) -> dict[str, Any]:
     cross_counts: Counter[tuple[int, int]] = Counter()
 
     for e in all_edges:
-        # TESTED_BY edges are expected cross-community coupling (test → code),
-        # not an architectural smell.
+        # TESTED_BY edges are expected cross-community coupling (code and
+        # its tests), not an architectural smell.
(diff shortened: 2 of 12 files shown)


tirth8205:
Thank you for independently tracing the TESTED_BY direction bug across the consumers. #624 has now merged those behavior corrections plus parser→store→query and dead-code regressions. It used the earlier credited TESTED_BY series rather than cherry-picking this commit, but the current head has no remaining unique behavior and is fully superseded. Closing it to avoid a duplicate merge.

PR #600, opened 2026-07-03
feat: keep test noise out of community names and attach tests to their subjects

SHudici:
## Problem

On test-heavy repositories, community output is dominated by test artifacts in two ways:

1. **Clustering:** shared fixtures and helpers give test files dense internal `CALLS` edges, while `TESTED_BY` links to subjects are sparse and weakly weighted (0.4). Leiden therefore clusters tests with each other instead of with the code they cover, producing test-only blobs that are useless for review context.
2. **Naming:** BDD-style test names (`test_should_return_x_when_y`) flood the keyword counter with grammar words, and `tests/` directories win the file-prefix vote — mixed communities come out named like `tests-should`.

## Fix

- **`_reassign_test_nodes`** (new post-Leiden pass): each `Test` node moves to the community holding the majority of its `TESTED_BY` partners; its own community counts as a vote, so tests already placed with their subjects stay put. Ties also prefer the current cluster and otherwise resolve to the lowest cluster index, so the outcome never depends on edge order. The test endpoint of an edge is identified by **node kind, not edge direction**, so the pass is independent of which direction the parser emits (and composes with the TESTED_BY direction fix, but does not require it). `TESTED_BY` endpoints that inherited a bare name from an unresolved cross-file call are resolved by node name when the name is unambiguous, so their tests still vote.
- **Naming:** mixed communities are named from their production members only, and BDD grammar words (`should`, `when`, `given`, `returns`, ...) joined `_COMMON_WORDS`.

Deliberate scoping: the file-based fallback keeps grouping strictly by directory — its contract is "group by file", and moving tests out of their file group would contradict it. Behavior notes: tests with no `TESTED_BY` partner keep their original cluster; a test-only cluster that loses most members to reassignment can fall below `min_size` and drop out of the community list.

## Testing

- Unit tests for `_reassign_test_nodes` (move, direction-agnostic, majority vote, own-cluster votes, no-op without TESTED_BY, test-test edges ignored, bare-name endpoints resolved when unique / ignored when ambiguous, tie-breaks: stay home on a tie, deterministic lowest-cluster otherwise) — pure-function tests that run without igraph.
- Naming tests: mixed community named from the production side, pure-test community still named, BDD words never become keywords.
- End-to-end Leiden test (skipped when igraph is absent): two dense clusters linked only by TESTED_BY end up with every test co-located with its subject.
- Full suite: 1412 passed / 0 failed (igraph 1.0.0 and without igraph).

### DIFF
--- a/code_review_graph/communities.py
+++ b/code_review_graph/communities.py
@@ -57,12 +57,22 @@ _COMMON_WORDS = frozenset({
     "add", "remove", "make", "build", "from", "to", "for", "with",
     "the", "and", "test", "main", "run", "do", "is", "has", "on",
     "of", "in", "at", "by", "my", "this", "that", "all", "none",
+    # BDD test-name grammar ("test_should_return_x_when_y") — frequent
+    # in test-heavy communities but carries no domain meaning.
+    "should", "when", "then", "given", "return", "returns", "raise",
+    "raises", "expect", "expected", "assert", "tests", "be", "it", "if",
+    "not",
 })
 
 
 # ---------------------------------------------------------------------------
 # Community naming
 # ---------------------------------------------------------------------------
+
+
+def _is_test_node(node: GraphNode) -> bool:
+    """True when *node* is a test (by kind or the parser's is_test flag)."""
+    return node.kind == "Test" or node.is_test
 
 
 def _generate_community_name(members: list[GraphNode]) -> str:
@@ -76,6 +86,14 @@ def _generate_community_name(members: list[GraphNode]) -> str:
     """
     if not members:
         return "empty"
+
+    # Mixed communities are named from their production members only:
+    # tests parrot the subject's vocabulary through BDD grammar and live
+    # under tests/ directories, which pollutes both the keyword and the
+    # file-prefix signals.
+    production = [m for m in members if not _is_test_node(m)]
+    if production and len(production) < len(members):
+        members = production
 
     # 1. Find common file prefix
     file_paths = [m.file_path for m in members]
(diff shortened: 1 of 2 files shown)


tirth8205:
Maintainer disposition: the production-only naming and TESTED_BY reassignment are viable, but this branch will not be merged directly. It overlaps #603 and #605 in the same community pipeline, so I am preparing one current-main replacement with Stefan Hudici attribution.\n\nCombined review found three corrections required around this behavior:\n1. Reassigning tests only before oversized-community splitting is not stable: the later split can separate tests from their subjects again. The final reassignment must run after splitting.\n2. Multiple TESTED_BY edges from repeated call sites currently count as multiple votes for the same subject. Voting must use unique subject partners or repeated calls can manufacture a false majority.\n3. Mutating community membership while iterating makes the move pass order-dependent and quadratic. The replacement computes moves from an immutable snapshot and applies them as a separate phase.\n\nThe replacement retains direction-agnostic TESTED_BY handling, unambiguous bare-name resolution, deterministic tie-breaking, production-only mixed-community names, and BDD stop words. It will be merged only after combined regressions, full-suite, graph, Linux, and Windows checks pass.

tirth8205:

Retained from this contribution: production-oriented community naming and post-detection test placement. Stefan Hudici is recognized as a co-author on the replacement commit.

This source branch was not merged directly because it is based on old, overlapping community code and did not safely handle duplicate TESTED_BY call sites or globally ambiguous bare endpoint names. The replacement counts unique tested subjects, skips ambiguous bare names, preserves ties deterministically, and rebuilds assignments without order-dependent in-place moves.

PR #601, opened 2026-07-03
fix: resolve bare-name TESTED_BY sources and run resolution in postprocess

SHudici:
## Problem

TESTED_BY edges are derived from test→production CALLS edges at parse time, so their production endpoint inherits the bare name of any call the parser couldn't resolve cross-file. That is the *dominant* case in practice — tests almost always import the code they test. On a mid-size production repo we measured **79% of TESTED_BY edges (6.8k of 8.5k) carrying bare sources**, so every qualified-name consumer (`tests_for`, review context, transitive test lookup) returned 0 results for most functions.

Compounding this: `GraphStore.resolve_bare_call_targets` — the existing post-hoc resolver for bare CALLS targets — is defined but **never called from any pipeline**, so bare endpoints survived every build untouched.

## Fix

- **`GraphStore.resolve_bare_tested_by_sources()`**: resolves bare TESTED_BY sources with the same strategy as the CALLS resolver (unique node name → resolve; multiple candidates → prefer the one whose file the test file imports; otherwise stay bare rather than guess). Both methods now share one private helper; CALLS behavior is unchanged and pinned by a regression test.
- **Wire resolution into all three postprocess paths** (`_run_postprocess` for the MCP build tool, `run_postprocess` for the re-run tool, shared `run_post_processing` for watch mode), ahead of the derived steps so flows and communities see qualified edges too. It runs at the `minimal` level as well — it's a cheap correctness pass over an indexed subset, not an enrichment. Failures degrade to build warnings per the existing step-isolation contract.

## Measured effect (same production repo, 4.7k nodes / 39k edges)

- 4,117 bare CALLS targets + 2,588 bare TESTED_BY sources resolved in one postprocess pass (the rest are genuinely ambiguous and deliberately left bare)
- qualified TESTED_BY edges: 1,756 → 4,344
- `tests_for` on a well-tested function: **0 results → 15 results**
- communities consolidated 228 → 103 as cross-file edges finally connected at clustering time

## Testing

New tests: resolver units (unique name, ambiguous-stays-bare, import disambiguation via the test file, unknown name untouched, CALLS regression), shared-pipeline wiring (resolves + reports `bare_edges_resolved`, failure isolated to a warning), and build-tool level gating (`minimal` resolves, `none` skips). Full suite passes.

Composes with #598 (TESTED_BY read direction) but does not require it; this PR is cut independently from main.

### DIFF
--- a/code_review_graph/graph.py
+++ b/code_review_graph/graph.py
@@ -492,11 +492,42 @@ class GraphStore:
 
         Returns the number of resolved edges.
         """
+        return self._resolve_bare_endpoints("CALLS", "target_qualified")
+
+    def resolve_bare_tested_by_sources(self) -> int:
+        """Batch-resolve bare-name TESTED_BY sources using the global node table.
+
+        TESTED_BY edges are derived from test-to-production CALLS edges at
+        parse time, so their production endpoint (the source) inherits the
+        bare name of any call the parser couldn't resolve cross-file — the
+        common case, since tests almost always import the code they test.
+        Qualified-name lookups (tests_for, review context) miss those edges
+        entirely.  Same in-place resolution and disambiguation strategy as
+        :meth:`resolve_bare_call_targets`.
+
+        Returns the number of resolved edges.
+        """
+        return self._resolve_bare_endpoints("TESTED_BY", "source_qualified")
+
+    def _resolve_bare_endpoints(self, kind: str, endpoint: str) -> int:
+        """Resolve bare-name edge endpoints of *kind* in column *endpoint*.
+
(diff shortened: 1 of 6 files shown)


tirth8205:
Thanks—this addresses a real TESTED_BY gap. The TESTED_BY direction correction is already included in PR #624, but this resolver still needs more safety before merging.

A globally unique name can currently be linked from an unrelated file even when there is no import evidence. Please require same-file, import, or equally strong evidence before resolving it. Please also add false-positive tests for unrelated files, positive import tests, and representative precision and runtime measurements. The measurements should explain the large community change, not only count newly resolved edges.

PR #602, opened 2026-07-03
feat: extract docstrings into node metadata and embed them for semantic search

SHudici:
## Problem

Semantic search embeds only identifier-derived text — name, parent, params, return type, directory (`_node_to_text`). The one thing written in the same natural language as user queries — the docstring — is never captured anywhere in the graph. Queries that describe *behavior* ("parse uploaded rate sheets", "retry on transient errors") only match nodes whose names happen to share words with the query.

## Fix

**Parser**: extract a documentation summary for every function/class into `extra["docstring"]` — no schema change, `nodes.extra` already exists and round-trips through the store:

- **Python**: the real docstring, matching CPython semantics — plain/r/u string first statements including parenthesized and implicitly concatenated literals; bytes and f-strings are **not** docstrings (`__doc__` stays None, so no key is stored). Handles both grammar shapes (bare `string` in the block, and `expression_statement`-wrapped).
- **JSDoc / Javadoc / Doxygen**: the `/** ... */` (or `/*! ... */`) block directly above the definition; for exported JS/TS declarations the block above the `export` statement is found too.
- **C#-style doc lines**: `///` and `//!`; attributes/decorators sitting between the comment and the definition are skipped.
- **Rust**: `///` only — inner doc comments (`//!`, `/*! ... */`) document the enclosing module/crate, never the following item, so they are dropped (leading `//!` lines above a `///` block are trimmed).
- **Go**: contiguous plain `//` block, matching godoc.

Deliberately conservative: plain `//` comments in non-Go languages are ignored as noise, and a blank line detaches a comment from the definition. Stored value is the **first paragraph** (the summary by every doc convention — PEP 257, JSDoc, godoc, rustdoc), whitespace-collapsed, capped at 400 chars so multi-page docstrings don't bloat the DB or drown the name/signature terms.

**Embeddings**: `_node_to_text` appends the docstring summary. Existing graphs pick this up automatically — the embedding text hash changes, so the next embed/refresh re-embeds documented nodes.

## Measured effect (mid-size production repo)

Parsing a 352-file Python service (4,134 non-File nodes): **812 nodes (19.6%) gain a docstring summary** that was previously invisible to search — 32.3% of Functions (641/1,983), 31.6% of Classes (125/396), 2.6% of Tests. Example of what enters the embedding text that no identifier carries:

> `require_auth` → *"FastAPI dependency that validates the Bearer token against atlas_api_key."*

A query like "where is the bearer token validated" previously had zero lexical overlap with this node.

## Testing

New tests (17): Python function/class/method docstrings, first-paragraph selection, 400-char cap, no-docstring leaves no key, bytes/f-string rejection, parenthesized implicit concatenation; JSDoc block; exported JS declarations; plain `//` ignored on JS; godoc plain-`//` block accepted on Go; Rust `///` across `#[inline]` plus `//!`/`/*!` inner-doc rejection; Javadoc first paragraph (param tags excluded); blank-line detachment. Plus `_node_to_text` includes the docstring and is byte-identical when absent. Full suite passes.

Follow-ups deliberately out of scope: docstrings for the JS `const f = () => {}` extraction site and language-specific creation sites (Elixir/Julia/Nix/R); adding docstring to the FTS index (needs an FTS5 table rebuild migration).

### DIFF
--- a/code_review_graph/embeddings.py
+++ b/code_review_graph/embeddings.py
@@ -802,8 +802,9 @@ def _node_to_text(node: GraphNode) -> str:
     Designed so natural-language queries land on the right node, not just on
     the enclosing class. We include the dotted ``Parent.name`` form, the
     identifier split into words, an explicit ``"in <Parent>"`` phrase, the
-    enclosing module directory, and the language. Tested by the
-    ``multi_hop_retrieval`` benchmark — see ``docs/REPRODUCING.md``.
+    docstring summary when the parser extracted one, the enclosing module
+    directory, and the language. Tested by the ``multi_hop_retrieval``
+    benchmark — see ``docs/REPRODUCING.md``.
     """
     parts: list[str] = []
 
@@ -836,14 +837,22 @@ def _node_to_text(node: GraphNode) -> str:
     if node.return_type:
         parts.append(f"returns {node.return_type}")
 
-    # 7. Module / directory context from the file path — gives queries a
+    # 7. Docstring / doc comment summary — the author's own description of
+    # what the node does, and the only part written in the same natural
+    # language as the queries. Extracted by the parser (first paragraph,
+    # capped) into extra["docstring"].
+    docstring = node.extra.get("docstring") if node.extra else None
+    if docstring:
+        parts.append(str(docstring))
+
+    # 8. Module / directory context from the file path — gives queries a
     # term like "routing" or "client" to anchor against.
     if node.file_path:
         parent_dir = Path(node.file_path).parent.name
         if parent_dir and parent_dir not in (".", "src", "lib"):
             parts.append(parent_dir)
 
-    # 8. Language
+    # 9. Language
     if node.language:
         parts.append(node.language)
 
--- a/code_review_graph/parser.py
+++ b/code_review_graph/parser.py
@@ -785,6 +785,72 @@ def _is_test_function(
     return False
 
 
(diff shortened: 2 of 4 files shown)


tirth8205:
Thanks — this is a useful, well-scoped semantic-search improvement and we want to carry it forward. The exact head conflicts with the current parser, and both workflows stopped before running jobs, so it cannot merge directly.

We are porting it onto current main with your attribution. The port will make Python literal handling match the stated semantics, cover the advertised C# comment form, document or automate the graph reparse needed before re-embedding existing repositories, and run the full suite plus semantic-search benchmark. No further action is required unless you prefer to rebase it yourself.

tirth8205:
Thanks again, Stefan. Final disposition: [PR #653](https://github.com/tirth8205/code-review-graph/pull/653) carries this PR's documentation-summary extraction and documentation-aware embedding intent across Python, JavaScript/Java/Doxygen, C#, Rust, and Go, with your authorship preserved.

This source cannot merge directly because it conflicts with the current parser and its workflows stopped before running jobs. The current-main port also corrects Python literal semantics, documentation markup and wrapped-declaration handling, Go directives, the 400-character provider-input bound, and full-build/incremental persistence and removal behavior.

PR #653 is merged into `main` at [`adf8604`](https://github.com/tirth8205/code-review-graph/commit/adf8604688523287db05e64333281bf48b921dcf). Closing this source PR as superseded; no further action is required.

PR #603, opened 2026-07-03
fix: name community splits from their members and dedupe community names

SHudici:
## Problem

Two related defects make community output hard to trust on real graphs:

1. **Splits are opaque.** `_split_oversized` names every shard `"<parent>-sub<N>"` and hardcodes `"cohesion": 0.0`. On a mid-size production graph (~7k nodes, 103 communities), 26 of the 103 communities were indistinguishable `services-load-subN` entries, all reporting zero cohesion — the numbers looked like detection failures, not results.
2. **Names collide.** Nothing enforces uniqueness: the same graph carried five collision groups (`services-normalize` ×4, `services-job` ×3, `services-fake` ×3, `services-sql` ×2, `services-testclientlazyinit` ×2 — 14 communities sharing 5 names). `get_community` resolves by name match, so 9 of those 14 were unreachable by name.

## Fix

- `_split_oversized` names each shard from its **own members** via the existing `_generate_community_name` (fallback `"<parent>-<id>"` only when members yield nothing), and computes **real cohesion** for all shards in one `_compute_cohesion_batch` pass over the full edge set — same cost profile as the top-level detectors.
- `detect_communities` runs a new `_dedupe_community_names` pass after splitting: within a collision group the largest community keeps the bare name; every other member is suffixed with its most distinctive keyword (skipping keywords already in the name and candidates that would collide with any existing name), with a deterministic numeric fallback.

## Measured effect (same production repo)

Re-ran community detection on the same graph (103 communities before and after — the partition is untouched, only naming and cohesion change):

| metric | before | after |
|---|---|---|
| `-subN` placeholder names | 26 | 0 |
| communities reporting cohesion 0.0 | 26 | 0 |
| duplicate-name groups | 5 (14 communities) | 0 |
| communities unreachable via `get_community` by name | 9 | 0 |

The 26 former `services-load-subN` shards now carry member-derived names with real cohesion, e.g. `services-upload` (85 members, 0.29), `services-rows` (74, 0.42), `lx-raw` (35, 0.27), `iag-market` (8, 0.33), `afklm-parse` (7, 0.41).

## Testing

New tests: a dumbbell-graph oversized community splits into member-named shards (no `-subN`), each with exact expected cohesion (15/16) and parent lineage preserved; dedup keeps the bare name on the largest community, suffixes the rest by keyword, skips keywords already in the name, avoids colliding with existing community names, falls back to a numeric suffix, and leaves unique names untouched. Full suite passes.

Composes with #600 (community naming quality) but does not require it; cut independently from main.

### DIFF
--- a/code_review_graph/communities.py
+++ b/code_review_graph/communities.py
@@ -539,15 +539,33 @@ def _split_oversized(
 
             parent_id = comm.get("id", 0)
             comm_name = comm.get("name", "")
-            for sub_members in sub_communities.values():
+            # Name each split from its own members (the parent's name plus
+            # a "-subN" counter tells a reader nothing), and compute real
+            # cohesion against the full edge set instead of stamping 0.0.
+            qn_to_node = {n.qualified_name: n for n in member_nodes}
+            sub_member_lists = list(sub_communities.values())
+            sub_cohesions = _compute_cohesion_batch(
+                [set(m) for m in sub_member_lists], edges,
+            )
+            for sub_members, sub_cohesion in zip(
+                sub_member_lists, sub_cohesions,
+            ):
+                sub_nodes = [
+                    qn_to_node[qn]
+                    for qn in sub_members
+                    if qn in qn_to_node
+                ]
+                sub_name = _generate_community_name(sub_nodes)
+                if not sub_name or sub_name == "cluster":
+                    sub_name = f"{comm_name}-{next_id}"
                 sub_comm = {
                     "id": next_id,
-                    "name": comm_name + f"-sub{next_id}",
+                    "name": sub_name,
                     "level": comm.get("level", 0) + 1,
                     "parent_id": parent_id,
                     "members": sub_members,
                     "size": len(sub_members),
-                    "cohesion": 0.0,
+                    "cohesion": sub_cohesion,
                     "dominant_language": comm.get(
                         "dominant_language"
                     ),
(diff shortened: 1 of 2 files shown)


tirth8205:
Maintainer disposition: member-derived split names, real split cohesion, and unique community names are viable, but this branch overlaps #600 and #605 and will be replaced by one current-main port with Stefan Hudici attribution.\n\nCombined review found four integration defects that must be corrected before merge:\n1. Oversized splitting filters edges against full qualified-name members before resolving bare TESTED_BY endpoints, so valid test-to-subject evidence can disappear from the split subgraph.\n2. Parallel edges between the same pair can over-weight that relationship during the split Leiden pass; the split graph needs deterministic pair aggregation.\n3. Splitting after #600's test reassignment can separate tests from their subjects again; the final reassignment must run after all splits.\n4. Name-deduplication suffix candidates can reintroduce test-only vocabulary that #600 deliberately removed. Suffix selection must use the same production-member and stop-word policy.\n\nThe replacement also keeps exact batch cohesion, parent lineage, collision avoidance, deterministic numeric fallback, and member-derived names. It will be merged only after focused, full-suite, graph, Linux, and Windows verification.

tirth8205:

Retained from this contribution: naming/cohesion for oversized splits and deterministic community-name deduplication. Stefan Hudici is recognized as a co-author on the replacement commit.

This source branch was not merged directly because it overlaps the same old implementation and loses bare TESTED_BY endpoints while filtering a split subgraph; it also permits repeated call-site edges to bias Leiden. The replacement resolves endpoints before split filtering, reassigns tests after splitting, and aggregates each undirected vertex pair once with a deterministic maximum weight.

PR #604, opened 2026-07-03
feat: attach graph freshness provenance to MCP tool responses

SHudici:
## Problem

Tool responses carry no freshness signal. An agent asking `query_graph(callers_of, X)` gets an answer that may describe a graph built hours ago, on a different branch, at a different commit — and has no way to tell. The build metadata already exists (`last_updated`, `git_branch`, `git_head_sha` in the metadata table); it's just never surfaced where decisions are made.

## Fix

Every graph-backed tool response now carries a compact `_graph` provenance envelope:

```json
"_graph": {
    "updated_at": "2026-07-03T18:22:41",
    "age_seconds": 5121,
    "built_on_branch": "main",
    "built_at_sha": "b72413c9d0aa1e..."
}
```

- `graph_provenance()` (`tools/_common.py`) reads the three metadata keys via a read-only SQLite connection. The db path is escaped with `Path.as_uri()` so URI-significant characters (`#`, `%`) in repo paths cannot derail SQLite's URI parser. **Best-effort by design**: any failure — no graph, unreadable DB, missing metadata — returns `None` and never fails the tool call. `age_seconds` is clamped to ≥ 0 and omitted if the timestamp doesn't parse; branch/sha are included only when present (full SHA, not truncated — agents comparing against `git rev-parse HEAD` shouldn't need prefix semantics).
- `with_provenance()` attaches the envelope to dict responses only, skips results that already carry `_graph`, and passes everything else through untouched.
- All 27 graph-backed MCP tools in `main.py` wrap their returns at the return site (not via a decorator — the #46/#136 regression guards introspect tool source with `inspect.getsource`, and a decorator would defeat them). For the five async tools the wrap runs **inside** the `asyncio.to_thread` worker: the envelope read opens the graph DB, and the event loop must never touch disk (#46, #136). Excluded: `get_docs_section_tool`, `list_repos_tool`, `cross_repo_search_tool` (not backed by a single repo graph).

Cost: one read-only SQLite open + a 3-row SELECT per tool call (off-loop for async tools), ~4 lines of JSON per response.

## Testing

13 new tests: metadata read (all fields), URI-hostile repo paths (`%` and `#`), future-timestamp clamping, unparseable-timestamp omits `age_seconds`, optional branch/sha, `None` for missing `last_updated` / missing graph DB / invalid repo root, envelope attach, no-provenance pass-through, non-dict pass-through, existing-`_graph` preservation, and one end-to-end registered-tool response. The #46/#136 async/to_thread guards still pass. Full suite passes.

### DIFF
--- a/code_review_graph/main.py
+++ b/code_review_graph/main.py
@@ -57,6 +57,7 @@ from .tools import (
     run_postprocess,
     semantic_search_nodes,
     traverse_graph_func,
+    with_provenance,
 )
 
 logger = logging.getLogger(__name__)
@@ -122,14 +123,20 @@ async def build_or_update_graph_tool(
         recurse_submodules: If True, include files from git submodules.
             When None (default), falls back to CRG_RECURSE_SUBMODULES env var.
     """
-    return await asyncio.to_thread(
-        build_or_update_graph,
-        full_rebuild=full_rebuild,
-        repo_root=_resolve_repo_root(repo_root),
-        base=base,
-        postprocess=postprocess,
-        recurse_submodules=recurse_submodules,
-    )
+    root = _resolve_repo_root(repo_root)
+
+    def _run() -> dict:
+        # with_provenance runs inside the worker thread too: it opens the
+        # graph DB, and the event loop must never touch disk (#46, #136).
+        return with_provenance(build_or_update_graph(
+            full_rebuild=full_rebuild,
+            repo_root=root,
+            base=base,
+            postprocess=postprocess,
+            recurse_submodules=recurse_submodules,
+        ), root)
+
+    return await asyncio.to_thread(_run)
 
 
 @mcp.tool()
@@ -154,11 +161,15 @@ async def run_postprocess_tool(
         fts: Rebuild FTS index. Default: True.
         repo_root: Repository root path. Auto-detected if omitted.
     """
-    return await asyncio.to_thread(
-        run_postprocess,
-        flows=flows, communities=communities, fts=fts,
-        repo_root=_resolve_repo_root(repo_root),
-    )
+    root = _resolve_repo_root(repo_root)
+
+    def _run() -> dict:
+        return with_provenance(run_postprocess(
+            flows=flows, communities=communities, fts=fts,
+            repo_root=root,
+        ), root)
+
+    return await asyncio.to_thread(_run)
 
 
 @mcp.tool()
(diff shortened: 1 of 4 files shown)


tirth8205:
Maintainer disposition: graph freshness provenance is viable and will be ported onto current main with Stefan Hudici attribution; this stale branch will not be merged directly because the MCP wrapper surface has moved while other reconciled work landed.\n\nThe replacement will retain the read-only metadata envelope and off-event-loop reads, while correcting one uncovered timestamp case: datetime.fromisoformat accepts timezone-aware values such as 2026-07-17T12:00:00+00:00, but subtracting that value from naive datetime.now raises TypeError. The broad outer handler then drops the entire provenance envelope, including branch and SHA, instead of merely omitting age_seconds as documented.\n\nAcceptance for the replacement:\n- naive and timezone-aware timestamps both work;\n- malformed timestamps retain updated_at and omit only age_seconds;\n- URI-significant and Windows paths remain safe;\n- all graph-backed tool responses are covered;\n- async tools perform the SQLite read inside their worker thread;\n- existing response fields and tool signatures remain compatible.\n\nI will post the replacement PR, full test/CI evidence, merge commit, and attribution here when complete.

tirth8205:
Final maintainer disposition: the safe graph-provenance feature has merged through #644 in commit b0e605c4f34d0980f42a2fd3f7d5f31669c1225f, with Stefan Hudici retained as co-author.\n\nRetained:\n- a top-level _graph envelope containing updated_at, derived age_seconds, and optional build branch/SHA;\n- best-effort behavior that never replaces or removes existing tool response fields;\n- coverage across every single-repository graph tool, while repository-list, cross-repository, and static-doc tools correctly avoid claiming one graph.\n\nCorrected before merge:\n- timezone-aware last_updated values no longer raise a naive/aware datetime TypeError and silently lose provenance;\n- async heavy tools perform both the tool work and provenance SQLite read off the event-loop thread;\n- metadata is opened through an escaped read-only SQLite URI;\n- an exclusive database lock was reproduced as a 5.23-second delay under the default SQLite timeout, then bounded to a 50 ms fail-soft read with a regression test.\n\nVerification on the exact merge base:\n- 147 focused tests passed with 1 native-path skip;\n- 1,611 full-suite tests passed with 1 native-path skip and 2 existing xpasses;\n- Ruff, mypy across 62 files, Bandit, diff checks, graph review, and all 11 required GitHub checks passed, including Windows.\n\nThis source PR is fully superseded by the hardened replacement and can be closed without losing its valid work.

PR #605, opened 2026-07-04
fix: split camelCase and truncate community slugs at word boundaries

SHudici:
## What

`_to_slug()` lowercases its input and hard-slices at 30 characters, so a camelCase class name produces an unreadable, mid-word community name. Observed on a real ~4.8k-node graph:

```
services-testbuildpreanalysispromptbloc   # from TestBuildPreAnalysisPromptBlock...
```

Two problems compound here: camelCase survives lowercasing as one giant "word" (the existing `[^a-z0-9]+` substitution never fires inside it), and the `[:30]` slice cuts wherever it lands.

## How

`_to_slug()` now:

1. normalizes punctuation to spaces,
2. splits camelCase/snake_case into words via the existing `_split_name()` helper,
3. joins with hyphens,
4. truncates at the **last complete word** within the 30-char budget.

A single word longer than the budget still falls back to a hard cut (best effort).

```
TestBuildPreAnalysisPromptBlock  ->  test-build-pre-analysis-prompt   (was: testbuildpreanalysispromptbloc)
AuthService                      ->  auth-service                     (was: authservice)
rate_sheet                       ->  rate-sheet                       (unchanged)
```

Community names derived from dominant classes become hyphenated and readable; names from file stems and already-split keywords are unchanged.

## Testing

- New `TestToSlug` class: 9 tests covering camelCase splitting, exact-boundary cuts, mid-word cuts, overlong single words, punctuation, and empty input.
- `tests/test_communities.py` passes (36 passed on Windows; the teardown errors are the pre-existing Windows unlink issue addressed separately in #596).

## Known trade-offs (kept out of scope)

- Word-boundary truncation can collide slightly more often than a hard cut when two long names share every complete word within the budget (`service_SupercalifragilisticAlpha` / `...Beta` both become `service-supercalifragilistic`). The hard cut already collided for shared 30-char prefixes; name-level dedup is handled in #603.
- `_split_name` does not split consecutive-uppercase acronyms (`HTTPServer` → `httpserver`, unchanged from current behavior for every existing caller). Improving that helper would also change keyword extraction, so it belongs in its own change if wanted.

### DIFF
--- a/code_review_graph/communities.py
+++ b/code_review_graph/communities.py
@@ -149,9 +149,28 @@ def _split_name(name: str) -> list[str]:
     return [p for p in re.split(r"[_\-.\s]+", s) if p]
 
 
+_SLUG_MAX_LEN = 30
+
+
 def _to_slug(s: str) -> str:
-    """Convert a string to a short lowercase slug."""
-    return re.sub(r"[^a-z0-9]+", "-", s.lower()).strip("-")[:30]
+    """Convert a string to a short lowercase slug.
+
+    CamelCase and snake_case inputs are split into hyphenated words, and
+    truncation happens at a word boundary so a slug never ends mid-word:
+    ``TestBuildPreAnalysisPromptBlock`` becomes
+    ``test-build-pre-analysis-prompt`` rather than
+    ``testbuildpreanalysispromptbloc``.
+    """
+    normalized = re.sub(r"[^A-Za-z0-9]+", " ", s)
+    slug = "-".join(w.lower() for w in _split_name(normalized))
+    if len(slug) <= _SLUG_MAX_LEN:
+        return slug
+    cut = slug[:_SLUG_MAX_LEN]
+    if slug[_SLUG_MAX_LEN] != "-":
+        head, sep, _ = cut.rpartition("-")
+        if sep:
+            cut = head
+    return cut.rstrip("-")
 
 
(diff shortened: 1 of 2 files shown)


tirth8205:
Maintainer disposition: camelCase splitting and word-boundary slug truncation are viable. This branch is being combined with overlapping community naming PRs #600 and #603 in one current-main replacement, with Stefan Hudici attribution.\n\nThe slug change itself is retained. The required integration correction is downstream: #603's name-deduplication pass can append raw keyword suffixes after the initial slug is generated, which can reintroduce test vocabulary or bypass the same readable-slug policy. The replacement applies production-member filtering, shared stop words, collision checks, and slug normalization consistently to both initial names and deduplication suffixes.\n\nCoverage includes camelCase, snake_case, punctuation, exact boundaries, long single words, collision groups, test-heavy mixed communities, and deterministic fallbacks. The combined replacement will be merged only after focused, full-suite, graph, Linux, and Windows checks pass.

tirth8205:

Retained from this contribution: stable slug normalization and unique community names. Stefan Hudici is recognized as a co-author on the replacement commit.

This source branch was not merged directly because it overlaps the same old naming path and could reintroduce test-only vocabulary through duplicate-name suffixes. The replacement derives names and suffixes from production nodes when available, uses one bounded slug policy for both base names and suffixes, and keeps results deterministic.

PR #611, opened 2026-07-07
fix: pre-warm local embeddings on all platforms, not just Windows

### LINKED ISSUE #610: macOS/Linux: concurrent first tool calls race the lazy torch import, permanently breaking embedding tools (prewarm only runs on win32)
## Environment

- code-review-graph **2.3.6** (`uvx --from 'code-review-graph[embeddings]==2.3.6' code-review-graph serve`, stdio MCP)
- macOS 15 (arm64), Python 3.13.12 (uv-managed), torch 2.12.1, sentence-transformers 3.4.1
- MCP client: Claude Code (issues tool calls **in parallel**)
- `CRG_EMBEDDING_MODEL=paraphrase-multilingual-MiniLM-L12-v2`

## Symptom

Once it happens, every torch-dependent tool fails for the **entire lifetime of the server process**:

```
Error calling tool 'list_graph_stats_tool': partially initialized module 'torch' from
'.../site-packages/torch/__init__.py' has no attribute 'library' (most likely due to a circular import)
```

- `list_graph_stats_tool` and `embed_graph_tool` fail with the error above (100% reproducible once poisoned)
- `semantic_search_nodes_tool` silently falls back to keyword/FTS mode, so vector search is effectively dead even when the `embeddings` table is fully populated
- Whether a given server process is healthy or poisoned is luck-dependent per session

## Root cause

`main.py` only pre-warms the embedding stack on Windows:

```python
if sys.platform == "win32":
    asyncio.set_event_loop_policy(asyncio.WindowsSelectorEventLoopPolicy())
    ...
    from .embeddings import prewarm_local_embeddings
    prewarm_local_embeddings()
```

On macOS/Linux, `torch` is lazy-imported inside `asyncio.to_thread` executor workers at first use. MCP clients like Claude Code batch independent tool calls in parallel, so the *first two* torch-touching calls regularly arrive together. Two executor threads then race `import torch`; the loser observes a half-initialized module, the import raises, and the partially initialized `torch` stays in `sys.modules` — poisoning every subsequent import in the process.

## Evidence

- `.../bin/python -c 'import torch; import torch.library'` in the exact same uvx-cached venv succeeds (torch 2.12.1 is fine)
- Calling `prewarm_local_embeddings()` on the main thread loads the model in ~7s and encoding works, including multilingual queries
- A graph built earlier by the same setup has 11k+ rows in `embeddings` — i.e. some sessions win the race and work, which matches a nondeterministic import race rather than a broken environment

## Suggested fix

Hoist the `prewarm_local_embeddings()` call out of the `sys.platform == "win32"` guard so it runs on every platform (the event-loop policy line stays Windows-only). The Windows deadlock and the macOS/Linux import race have the same cure: perform the dangerous initialization on the main thread before any worker thread exists. `prewarm_local_embeddings()` is already a no-op for cloud-provider setups without sentence-transformers installed.

Happy to send a PR for this.

## Workaround for other users

Wrap the server start so prewarm runs before `serve`:

```json
"args": [
  "--from", "code-review-graph[embeddings]==2.3.6", "python", "-c",
  "import sys; from code_review_graph.embeddings import prewarm_local_embeddings; prewarm_local_embeddings(); sys.argv = ['code-review-graph', 'serve']; from code_review_graph.cli import main; main()"
]
```

FRONT-JB:
Fixes #610

## Problem

`prewarm_local_embeddings()` only runs on Windows. On macOS/Linux, `torch` is lazy-imported inside `asyncio.to_thread` executor workers at first use. MCP clients (e.g. Claude Code) batch independent tool calls in parallel, so the first two torch-touching calls regularly arrive together: two executor threads race `import torch`, the loser sees a partially initialized module (`module 'torch' has no attribute 'library'`), and the half-initialized module stays poisoned in `sys.modules` for the process lifetime — `list_graph_stats_tool`/`embed_graph_tool` fail on every call and `semantic_search_nodes_tool` silently degrades to keyword mode. Full diagnosis in #610.

## Change

Hoist the `prewarm_local_embeddings()` call out of the `sys.platform == "win32"` guard so the dangerous initialization always happens on the main thread before any worker thread is spawned. The `WindowsSelectorEventLoopPolicy` line stays Windows-only. `prewarm_local_embeddings()` is already a no-op for cloud-provider setups without sentence-transformers, so non-embedding installs are unaffected.

Trade-off: server startup on macOS/Linux with local embeddings now pays the model-load cost (~7s for `paraphrase-multilingual-MiniLM-L12-v2` on an M-series Mac) up front instead of on first use — same behavior Windows already had.

## Verification

- `uv run pytest tests/ --tb=short -q` → **1384 passed, 1 skipped, 2 xpassed**
- `uvx ruff check code_review_graph/` → clean
- Manually verified on macOS arm64 / Python 3.13 / torch 2.12.1: running prewarm on the main thread before `serve` boots a healthy server (MCP initialize handshake OK, no torch errors), while the unpatched lazy path is reproducibly poisoned by concurrent first calls.

### DIFF
--- a/code_review_graph/main.py
+++ b/code_review_graph/main.py
@@ -1049,15 +1049,22 @@ def main(
 
     if sys.platform == "win32":
         asyncio.set_event_loop_policy(asyncio.WindowsSelectorEventLoopPolicy())
-        # Pre-warm sentence-transformers on the main thread before fastmcp's
-        # event loop starts. Lazy-loading ``torch`` + tokenizers inside an
-        # executor worker thread deadlocks ``semantic_search_nodes_tool`` on
-        # Windows stdio MCP (DLL init / OpenMP thread-pool registration grabs
-        # locks the loop needs). #385 added ``asyncio.to_thread`` to peer
-        # tools but cannot fix this case — the dangerous initialization has
-        # to happen on the main thread before any worker thread is spawned.
-        from .embeddings import prewarm_local_embeddings
-        prewarm_local_embeddings()
+
+    # Pre-warm sentence-transformers on the main thread before fastmcp's
+    # event loop starts — on every platform. On Windows, lazy-loading
+    # ``torch`` + tokenizers inside an executor worker thread deadlocks
+    # ``semantic_search_nodes_tool`` on stdio MCP (DLL init / OpenMP
+    # thread-pool registration grabs locks the loop needs); #385 added
+    # ``asyncio.to_thread`` to peer tools but cannot fix this case. On
+    # macOS/Linux, MCP clients batch tool calls in parallel, so two worker
+    # threads can race the first ``import torch``; the loser sees a
+    # partially initialized module that stays poisoned in ``sys.modules``
+    # for the process lifetime, breaking stats/embed tools and silently
+    # degrading semantic search to keyword mode (#610). Either way the
+    # dangerous initialization has to happen on the main thread before any
+    # worker thread is spawned.
+    from .embeddings import prewarm_local_embeddings
+    prewarm_local_embeddings()
 
     try:
         if transport == "stdio":


tirth8205:
Thanks—the concurrent Torch import race is credible, but loading the model unconditionally adds about seven seconds to server startup.

Please make first-use initialization concurrency-safe without imposing that cost on every startup—for example, serialize and cache the first local-embedding initialization. Add a test with concurrent first calls, verify initialization happens once, and add a startup-latency regression showing that unused embeddings do not load Torch or the model.

tirth8205:
Thanks again, FRONT-JB. We adopted the concurrent first-import diagnosis and the requirement for race-free local-embedding initialization in [PR #654](https://github.com/tirth8205/code-review-graph/pull/654), preserving your contribution through co-authorship.

This source is not being merged directly because unconditional prewarming on macOS and Linux moves Torch/model loading—and possible model resolution or download—to every server startup even when embeddings are unused. The replacement instead serializes lazy first use and cache publication while retaining the existing Windows prewarm.

PR #654 is merged into `main` at [`ac4b22b`](https://github.com/tirth8205/code-review-graph/commit/ac4b22b44c978dde0458ef0390d163575a1b3762). Closing this source PR as superseded; no further action is required.

PR #614, opened 2026-07-09
feat: CSS/SCSS/LESS support — stylesheets join the code graph

CodeBlackwell:
Hi! This PR teaches the graph to understand stylesheets, so reviews can see the relationship between components and the CSS that styles them.

### What's new

**Parsing**
- 🧩 `.css` and `.scss` files parse via tree-sitter: selectors, custom properties, `@media`, `@keyframes`, mixins, and SCSS variables all become graph nodes
- 🍃 `.less` support via a regex fallback (no tree-sitter grammar exists for LESS)
- 🖼️ Vue **and** Svelte `<style>` / `<style lang="scss">` blocks are parsed in place, with correct line offsets back into the SFC
- 📦 `@import` chains resolve to real files, including SCSS partials (`_variables.scss`)

**Cross-language edges**
- 🔗 **STYLES** — JSX `className`, Vue/Svelte static template classes, and CSS Modules references (`styles.btnPrimary` → `.btn-primary`, camelCase→kebab handled) link components to the selectors that style them
- ⚔️ **OVERRIDES** — same-file specificity conflicts, source-order overrides, and `!important` wins are tracked per property
- 🚨 **POTENTIAL_CONFLICT** — selectors in *different* files targeting the same class get flagged with both specificities

**Tooling**
- 🔍 Five new `query_graph` patterns: `styles_of`, `styled_by`, `overrides_of`, `overridden_by`, `conflicts_of`
- 📝 Review guidance now warns when a change touches CSS override relationships
- 🎨 Visualization renders the new edge kinds with their own colors, dashes, and legend entries; CSS nodes get a dashed ring

### Design notes
- Cross-file linking runs as a best-effort post-pass alongside the existing resolvers (ReScript/Spring/Temporal), so it can never fail a build and is skipped on incremental updates when no style-relevant file changed
- Query patterns are table-driven — a future edge kind is one dict entry, not a new branch
- Docs updated across README, USAGE, FEATURES, COMMANDS, and the edge-type schema

### Testing
- ✅ 68 new tests (parser fixtures for CSS/SCSS/LESS/TSX/Vue/Svelte + end-to-end linking through `full_build`)
- ✅ Full suite green on top of current `main`: 1437 passed, 1 skipped
- ✅ ruff clean

Happy to adjust anything — naming, edge semantics, or scope. Thanks for the great project!

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -90,7 +90,8 @@ uv run code-review-graph eval               # Run evaluation benchmarks
 - `tests/test_tools.py` — MCP tool integration tests
 - `tests/test_visualization.py` — Export, HTML generation, C++ resolution
 - `tests/test_incremental.py` — Build, update, migration, git ops
-- `tests/test_multilang.py` — Broad language parsing tests, including SFCs, notebooks, SQL, Perl XS, and modern systems/web languages
+- `tests/test_multilang.py` — Broad language parsing tests, including SFCs, notebooks, SQL, CSS/SCSS/LESS, Perl XS, and modern systems/web languages
+- `tests/test_css_linking.py` — Cross-file CSS linking (STYLES edges, CSS Modules, POTENTIAL_CONFLICT detection)
 - `tests/test_custom_languages.py` — Config-driven custom languages (languages.toml loader + end-to-end Erlang parse)
 - `tests/test_embeddings.py` — Vector encode/decode, similarity, store
 - `tests/test_flows.py` — Execution flow detection and criticality
--- a/README.md
+++ b/README.md
@@ -123,7 +123,7 @@ Large monorepos are where token waste is most painful. The graph cuts through th
   <img src="diagrams/diagram9_language_coverage.png" alt="Language coverage organized by category: Web, Backend, Systems, Mobile, Scripting, Config, plus Jupyter and Databricks notebook support" width="90%" />
 </p>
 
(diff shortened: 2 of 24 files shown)


tirth8205:
Thanks for the substantial stylesheet work — this is a valuable direction, but head f7bb0d0 is not safe to merge.

link_css_styles ignores the imported module path, so styles.foo can link to every .foo in the repository. !important and BEM-style prefixes are treated as overrides without proving selectors target the same element; repeated selectors collapse to one identity and can self-edge. CSS/SCSS imports are incomplete, and CSS Modules or scoped Vue/Svelte styles are flagged as high-confidence conflicts from bare-name matches.

Please preserve scope, restrict module links to the imported file, model selector occurrences, support common imports/partials, and bound repository-wide joins. Add negative multi-module/scoped-style tests, scale and incremental rename/deletion coverage, and VS Code support for new edge/query types. Rebase and run full approved CI.

tirth8205:
Thanks for the substantial stylesheet work. After reviewing exact head `f7bb0d0` and running adversarial end-to-end cases, I am closing this version.

CSS-module references ignore the resolved import path, so `styles.foo` can link to every `.foo`; an incremental module rename also linked a stale import to the renamed file. Bare-name matching reports scoped Vue/Svelte and CSS-module selectors as high-confidence conflicts, repeated selectors collapse to one identity and create `OVERRIDES` self-edges, and the `!important`/BEM heuristics infer overrides without proving the selectors can target the same element. A 100×100 repository case produced 10,000 `STYLES` edges, confirming the unbounded cross-repository join. Import/partial resolution and custom-property endpoints also leave unresolved graph data.

These faults are in scope, occurrence identity, and resolver semantics, so there is no safe parser-only subset to merge. The branch is 217 commits behind, non-mergeable, and both workflows are `action_required` with zero jobs.

PR #615, opened 2026-07-09
fix: use ThreadPoolExecutor for MCP stdio on all platforms

HouMinXi:
fix: use ThreadPoolExecutor for MCP stdio on all platforms

## Problem

When code-review-graph runs as an MCP stdio server,
ProcessPoolExecutor creates a fork server (fork+exec) and workers
(fork) that both inherit stdin as a Unix socket to the host. When
the host disconnects, Linux delivers EOF only when all fd references
are closed. The inherited fds keep the socket alive indefinitely --
the server's stdin read loop never terminates, producing zombie
processes (20+ after 8h, ~765MB RSS).

## Existing Workaround

_select_executor_kind() already switches to ThreadPoolExecutor on
Windows when stdin is not a TTY (#46, #136). The docstring
explains the rationale: "Tree-sitter parsing in the worker releases
the GIL during native parsing, so the speedup loss for falling back
to threads is small (typically <30% on the full-build path) and the
trade is worth it."

The same logic applies to all Unix platforms. MCP stdio transport
means stdin is a socket, not a TTY, on every platform.

## Fix

Remove the `sys.platform == "win32"` guard from the thread-fallback
condition. When stdin is not a TTY (MCP stdio mode), use
ThreadPoolExecutor on all platforms.

ThreadPoolExecutor avoids fork entirely:
- No fork server process
- No fd inheritance
- No zombie accumulation
- tree-sitter still runs in parallel (GIL released during native
  parsing)

CRG_PARSE_EXECUTOR=process/thread still overrides.

## Testing

Verified end-to-end on Linux (Python 3.14, Fedora 44):
- MCP server spawned via stdio transport
- build_or_update_graph_tool called with full_rebuild=True
- stdin closed to simulate host disconnect
- Server exited within 1s, zero zombie processes

### DIFF
--- a/code_review_graph/parser.py
+++ b/code_review_graph/parser.py
@@ -4579,11 +4579,13 @@ class CodeParser:
         """Walk ancestors to detect statically-dead guards.
 
         Returns True when ``node`` sits inside an ``if_statement`` whose
-        condition is ``False`` (the literal) or ``TYPE_CHECKING`` (the
-        ``typing`` sentinel).  This covers the two most common Python
-        idioms for compile-time-only code:
+        condition is ``False`` (the literal), the integer ``0``, or
+        ``TYPE_CHECKING`` (the ``typing`` sentinel).  This covers the
+        most common Python idioms for compile-time-only code:
 
             if False:
+                ...
+            if 0:
                 ...
             if TYPE_CHECKING:
                 ...
@@ -4605,6 +4607,12 @@ class CodeParser:
                 if cond is not None:
                     # ``if False:``
                     if cond.type == "false":
+                        return True
+                    # ``if 0:``
+                    if (
+                        cond.type == "integer"
+                        and cond.text == b"0"
+                    ):
                         return True
                     # ``if TYPE_CHECKING:``
                     if (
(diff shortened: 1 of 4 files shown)


tirth8205:
Thanks for the clear report. The one-line change makes sense, but it changes the default executor for every Unix MCP stdio run. Before merge, please add an automated end-to-end regression test: start the server with pipe or socket stdin, trigger parsing, close the parent side, and prove the server exits within a short timeout without leaving worker, fork-server, or zombie processes. Please also keep tests for the `CRG_PARSE_EXECUTOR=process/thread` overrides and TTY behavior. Once that Linux test is green, this is ready for another review.

tirth8205:
Thanks again, Minxi. We adopted the MCP-stdio descriptor-inheritance diagnosis and the all-platform thread-parsing remedy in [PR #648](https://github.com/tirth8205/code-review-graph/pull/648), retaining your co-authorship.

The source head cannot merge directly because it no longer contains the claimed production executor change, and its regression does not exercise the real server entry point, enter the parallel parsing path, or require a clean zero exit. The replacement explicitly scopes threads to active MCP stdio, preserves process pools for ordinary Unix CLI/CI/cron use, and validates a real JSON-RPC build and EOF shutdown end to end.

PR #648 is merged into `main` at [`11ea234`](https://github.com/tirth8205/code-review-graph/commit/11ea2340c8cf6d89fdd44747ee865ac9d909cc87). Closing this source PR as superseded.

Issue #497, opened 2026-05-20
Bug: Double JSON encoding of non-ASCII paths corrupts directory structure on Windows


Dont-know-code:
## Bug Description

When `code-review-graph install` is run in a project whose path contains non-ASCII characters (e.g., Chinese), the generated hooks configuration creates a **corrupted, deeply nested directory tree** on the desktop instead of operating inside the actual project directory.

### What the user sees

After opening the project in Claude Code, a deeply nested directory tree appears on the desktop:
```
Desktop/
  u57fa/              ← Unicode codepoint of "基" used as literal directory name
    u4e8eSTM32/       ← codepoint of "于" + "STM32"
      u7684/           ← codepoint of "的"
        ...
          .code-review-graph/
            graph.db
```

Instead of operating on the actual project at `Desktop/基于STM32的智能厨房安全监测系统设计/`.

## Root Cause: Double JSON Encoding

Two separate bugs combine to cause this:

### Bug 1 — `skills.py` line 553: `json.dumps()` abused as shell quoting

```python
repo_arg = json.dumps(repo_root.resolve().as_posix())
```

`json.dumps()` with `ensure_ascii=True` (default) performs TWO transformations that are wrong for shell quoting:
1. Encodes non-ASCII → `\uXXXX` escape sequences
2. Wraps the result in JSON double quotes

The resulting JSON-encoded string (containing literal `\uXXXX`) is embedded into a shell command string. This command is then JSON-serialized a **second time** when written to `settings.json` (line 721), which escapes the backslashes again.

### Bug 2 — `skills.py` line 386: `ensure_ascii=True` in `.mcp.json` cwd

```python
config_path.write_text(json.dumps(existing, indent=2) + "\n", encoding="utf-8")
```

The `cwd` field in `.mcp.json` is written with Unicode escapes (`\uXXXX`). While this is technically valid JSON, some MCP hosts / process launchers fail to properly decode `\uXXXX` sequences when setting the child process working directory.

### How the two bugs interact

```
skills.py:553   json.dumps(path)        ← encode Chinese → \uXXXX + JSON quotes
                     ↓
              embedded in shell command string
                     ↓
skills.py:721   json.dumps(hooks_config) ← second encode: \ → \ 
                     ↓
            written to settings.json with \uXXXX (literal)
                     ↓
Claude Code    json.loads()             ← ONE decode: \uXXXX → \uXXXX (literal string)
                     ↓
[...]

PR #618, opened 2026-07-10
fix: preserve literal paths in Git change discovery

teilomillet:
## Linked issue

No exact issue exists for these Git discovery functions; the defect was found with Ordeal while auditing the current Python Trending surface. #497 is related Unicode-path work, but it concerns generated host configuration rather than incremental Git output. PR #566 overlaps these subprocess blocks while addressing decode crashes, but its replacement decoding does not preserve Git's quoted paths.

## What & why

code-review-graph keeps a map of a repository so its review tools know which files changed and what those changes may affect. When someone edited a valid source file such as `café.py`, Git sometimes reported the name in an encoded form. The updater then searched for that encoded text instead of the real filename, so it could silently miss the changed file. This patch reads Git's unambiguous machine format, allowing the updater to find the real file while preserving its existing rename and copy behavior.

Before this change, both the working-tree and commit-diff paths produced:

```python
['"caf\\303\\251.py"']
```

instead of:

```python
["café.py"]
```

Downstream incremental parsing then looked up a filename that did not exist and could silently miss the changed source file. The line-based status parser also treated a literal ` -> ` inside a filename as a rename delimiter and truncated that path.

Git quotes unusual paths in its newline-delimited human-readable output. This change requests NUL-delimited output from `git diff --name-only -z` and `git status --porcelain=v1 -z`, keeps the subprocess output as bytes, and decodes each complete path with `os.fsdecode`. Rename and copy records retain the existing destination-only behavior by skipping the following source-path record.

The changelog now records both corrected behaviors. Ordinary ASCII paths and the existing rename behavior remain unchanged; no schema or migration change is required.

## Ordeal evidence

A contract-specific Ordeal `fuzz` property generated valid non-ASCII Python filenames and required each Git discovery API to return the literal path.

- `get_staged_and_unstaged`: 12 examples, minimized failure `{"filename": "café.py"}`.
- `get_changed_files`: 12 examples, minimized failure `{"filename": "café.py"}`.
- Direct reproduction outside the discovery harness returned `['"caf\\303\\251.py"']` from both APIs.
- The exact target-native regressions failed before the production change and passed afterward.
- Both Ordeal properties passed all 12 examples after the fix.

The semantic-delta check also covers simple paths, destination-only rename and copy records, and a literal `path -> literal.py` filename.

## How it was tested

```bash
uv lock --check
uv run pytest tests/test_integration_git.py -k unicode_path -q
uv run pytest tests/test_incremental.py::TestGitOperations -q
uv run pytest --tb=short -q --cov=code_review_graph --cov-report=term-missing --cov-fail-under=65
uv run ruff check tests/test_integration_git.py
uv run bandit -r code_review_graph/ -c pyproject.toml
git diff --check
```

Results:

- Full suite: 1,386 passed, 1 skipped, 2 xpassed.
- Coverage suite: 73.50%, above the required 65% threshold.
- Ruff, mypy, Bandit, lockfile, and diff checks passed.
- A self-hosted `code-review-graph build` parsed all 179 files, and `detect-changes --base origin/main --brief` analyzed the intended four-file patch.

The local run used macOS and Python 3.13. Repository CI will provide the configured Python 3.10-3.13 matrix. The coverage run emitted existing SQLite `ResourceWarning`s in unrelated parser, visualization, and wiki tests; they did not fail the suite.

## Checklist

- [x] Docs updated where behavior changed (`CHANGELOG.md`)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 ## [Unreleased]
 
+### Fixed
+
+- Incremental Git change discovery now preserves literal non-ASCII and
+  arrow-containing paths instead of returning quoted or truncated filenames.
+
 ## [2.3.6] - 2026-06-10
 
 **Community-response release.** Built from a full audit of every open PR,
--- a/code_review_graph/incremental.py
+++ b/code_review_graph/incremental.py
@@ -473,6 +473,11 @@ _SAFE_GIT_REF = re.compile(r"^[A-Za-z0-9_.~^/@{}\-]+$")
 _SAFE_SVN_REV = re.compile(r"^r?\d+(:r?\d+|:HEAD|:BASE|:COMMITTED)?$", re.IGNORECASE)
 
 
+def _decode_nul_paths(output: bytes) -> list[str]:
+    """Decode Git's NUL-delimited path output without quote mangling."""
+    return [os.fsdecode(path) for path in output.split(b"\0") if path]
+
+
 def _store_vcs_metadata(repo_root: Path, store: "GraphStore") -> None:
     """Persist VCS branch/revision info into the graph metadata table."""
     vcs = detect_vcs(repo_root)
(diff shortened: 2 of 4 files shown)


tirth8205:
Thanks—NUL-delimited byte output is the strongest direction here. Before this comes out of draft, please:

- Rebase on current main and preserve the existing branch/revision and all-tracked-file decoding behavior in this overlapping area.
- Add real-Git Windows tests for Unicode paths, literal ` -> ` filenames, rename/copy destination-only records, and NUL-delimited output.
- Run the full CI matrix and mark the PR ready once it is green.

The mocked tests are useful, but this path handling needs target-native Windows proof.

tirth8205:
Final maintainer disposition: the corrected implementation is merged through #650 at 4f9f035b49515fa79e4c084aafb10828d607ba7b, so this draft source PR is superseded.

Adopted from this contribution:
- NUL-delimited byte output for both git diff and porcelain status;
- filesystem-aware decoding that preserves literal path bytes;
- destination-first handling for rename and copy records.

Changes required before merge:
- rebased the focused change onto current main without carrying unrelated stale files;
- reject failed diff fallback and status commands instead of parsing misleading stdout;
- added real-Git Unicode tests that run on Windows;
- added a real staged-rename test proving destination-only ordering;
- added Unix real-Git tests for literal arrow and newline filenames;
- added mocked copy, whitespace, literal-arrow, NUL, and failed-command regressions.

Final evidence:
- all seven new regressions failed on the old implementation before the fix;
- 78 focused Git/incremental tests passed after the final rebase;
- 1,653 broader local tests passed, with 1 expected skip and 2 existing xpasses;
- Ruff and diff checks passed;
- all 11 GitHub checks passed, including Python 3.10 through 3.13 and native Windows.

Teïlo Millet is retained as co-author on the merged commit. Thank you for identifying and fixing the literal-path failure mode.

PR #660, opened 2026-07-18
fix(parser): end-to-end dead-guard fix with consumer filtering (#576)

### LINKED ISSUE #576: CALLS edges are emitted for calls inside statically-dead code (`if False:`, `if TYPE_CHECKING:`, `#if 0`)

HouMinXi:
## Summary

Fix #576 end-to-end. PR #580 tagged CALLS edges with `extra.reachable=False` but no consumer checked it — every caller, impact, flow, and dead-code path still traversed dead edges as live.

This fix takes the opposite approach: **dead CALLS edges are never emitted at all**. All downstream consumers also gained `is_live()` filtering for backward compatibility with old graphs.

### Parser (parser.py)

**Branch semantics fix** — only the consequence (true branch) of a dead guard is dead; else/elif branches are live:

```python
# BEFORE (PR #580): entire if_statement treated as dead
# AFTER: only consequence branch is dead
if False:
    dead_call()      # NOT emitted
else:
    live_call()      # emitted
```

**TYPE_CHECKING import verification** — only treat `TYPE_CHECKING` as the typing sentinel when `import_map` confirms it was imported from `typing`. A coincidentally-named local variable is not a dead guard.

**Byte-range comparison** — tree-sitter Python bindings create fresh `Node` objects per `child_by_field_name` call, so `is` identity comparison failed. Switched to byte-range matching in `_node_is_in_child()`.

**Scope boundaries** — added `lambda` and `async_function_definition` to the scope-boundary list so nested definitions under dead guards don't propagate dead status.

### Graph (graph.py)

- `GraphEdge.is_live()` — backward-compat predicate with `None`/`str`/`dict` defense for `extra`
- SQL filters use `COALESCE(json_extract(extra, '$.reachable'), 1) != 0` to handle new graphs (no `reachable` key) and old graphs (`reachable: False`)
- Filters added to: `get_all_call_targets` (2 queries), `_build_networkx_graph`, `get_impact_radius_sql` CTE, `get_edges_among`, `get_transitive_tests` (CALLS + 3 TESTED_BY queries), `load_flow_adjacency` (unified CALLS+TESTED_BY)

### Consumers (6 files)

Every consumer that traverses CALLS or TESTED_BY edges now filters via `is_live()`:

| File | Functions |
|------|-----------|
| analysis.py | `find_hub_nodes`, `find_knowledge_gaps`, `find_surprising_connections` (degree + edge loop), `find_bridge_nodes` (NetworkX) |
| changes.py | `compute_risk_score`, `analyze_changes` (TESTED_BY) |
| enrich.py | `_format_node_context` (callers + callees) |
| refactor.py | `rename_preview` (calls + bare-name fallback), `find_dead_code` (TESTED_BY × 3 + polymorphic SQL + member SQL), `suggest_refactorings` |
| tools/query.py | `query_graph` (callers_of calls + fallback, callees_of), `traverse_graph_func` (BFS/DFS) |

### Tests

- `test_dead_guard_omits_dead_edges` — verifies dead consequences are omitted, else/elif calls are live, total count = 5
- `test_type_checking_false_positive` — local `TYPE_CHECKING` variable (not from typing) must not suppress calls, total count = 1

## End-to-End Verification

Built a real graph from fixtures and verified at store level:

```
=== get_all_call_targets: 5 targets ===
  live_elif_call / live_final_else / live_in_else / live_helper / some_condition
  Leaked dead targets: NONE

=== is_live() backward compat ===
  reachable=False  -> False  (correct)
  no key           -> True   (correct)
  JSON string      -> False  (correct)
  None             -> True   (correct)

=== SQL filter ===
  Old dead edge (reachable=False): No (correctly filtered)
  New live edge (no key): Yes (correctly included)

ALL CHECKS PASSED
```

## Known Limitations (Follow-up)

1. **`typing.TYPE_CHECKING` attribute access** — `if typing.TYPE_CHECKING:` (attribute node) is not recognized. Needs attribute-node handling in `_is_in_static_dead_guard`.
2. **TYPE_CHECKING alias** — `from typing import TYPE_CHECKING as TC` is not recognized. Needs alias resolution from `import_map`.
3. **Cross-language** — dead-guard detection is Python-focused. `if (false) { ... }` in JavaScript is not handled (intentional scope for this PR).

## Review Process

This PR went through 9 rounds of automated architecture review (deepseek-v4-flash via code-forge with code-review-graph context injection). Rounds 1–8 found and fixed real bugs; round 9 returned zero new findings.

Key bugs caught during review:
- The original SQL filter used `IS NOT 0`, which silently excluded all CALLS edges on new graphs (no `reachable` key yields NULL, and `NULL IS NOT 0` is falsy). Fixed by switching to `COALESCE(..., 1) != 0`.
- `import typing` was incorrectly treated as a dead guard — when a file does `import typing` (the module itself), the local name `typing` was added to the sentinel set, causing any `if typing:` block to suppress calls. Fixed by requiring `local == "TYPE_CHECKING"` in addition to `qualified == "typing"`.
- `GraphEdge.is_live()` crashed on old databases where `extra` is stored as a JSON string or NULL. Added `None` and `isinstance(str)` guards.
- `async_function_definition` was missing from the scope-boundary list, so async functions nested under dead guards could have their calls incorrectly marked dead.
- The `search_edges_by_target_name` fallback path in both `query_graph` and `rename_preview` was not filtering dead edges from old graphs.
- TESTED_BY edges in `get_transitive_tests`, `find_dead_code`, and `analyze_changes` were not filtered, allowing dead test coverage to appear real.

### DIFF
--- a/code_review_graph/analysis.py
+++ b/code_review_graph/analysis.py
@@ -22,6 +22,8 @@ def find_hub_nodes(store: GraphStore, top_n: int = 10) -> list[dict]:
     in_degree: dict[str, int] = Counter()
     out_degree: dict[str, int] = Counter()
     for e in edges:
+        if not e.is_live():
+            continue
         out_degree[e.source_qualified] += 1
         in_degree[e.target_qualified] += 1
 
@@ -129,6 +131,8 @@ def find_knowledge_gaps(store: GraphStore) -> dict[str, list[dict]]:
     degree: dict[str, int] = Counter()
     tested_nodes: set[str] = set()
     for e in edges:
+        if not e.is_live():
+            continue
         degree[e.source_qualified] += 1
         degree[e.target_qualified] += 1
         if e.kind == "TESTED_BY":
@@ -228,9 +232,11 @@ def find_surprising_connections(
 
     node_map = {n.qualified_name: n for n in nodes}
 
-    # Build degree map
+    # Build degree map (skip dead edges)
     degree: dict[str, int] = Counter()
     for e in edges:
+        if not e.is_live():
+            continue
         degree[e.source_qualified] += 1
         degree[e.target_qualified] += 1
 
@@ -243,6 +249,8 @@ def find_surprising_connections(
 
     scored_edges = []
     for e in edges:
+        if not e.is_live():
+            continue
         src = node_map.get(e.source_qualified)
         tgt = node_map.get(e.target_qualified)
         if not src or not tgt:
@@ -285,12 +293,12 @@ def find_surprising_connections(
             reasons.append("peripheral-to-hub")
 
         # Cross-file-type: test <-> non-test (+0.15)
-        if src.is_test != tgt.is_test and e.kind == "CALLS":
+        if src.is_test != tgt.is_test and e.kind == "CALLS" and e.is_live():
             score += 0.15
             reasons.append("cross-test-boundary")
 
         # Non-standard edge kind (+0.15)
-        if e.kind == "CALLS" and src.kind == "Type":
+        if e.kind == "CALLS" and e.is_live() and src.kind == "Type":
             score += 0.15
             reasons.append("unusual-edge-kind")
 
--- a/code_review_graph/changes.py
+++ b/code_review_graph/changes.py
@@ -238,7 +238,7 @@ def compute_risk_score(store: GraphStore, node: GraphNode) -> float:
 
     # --- Community crossing (cap 0.15) ---
     callers = store.get_edges_by_target(node.qualified_name)
-    caller_edges = [e for e in callers if e.kind == "CALLS"]
+    caller_edges = [e for e in callers if e.kind == "CALLS" and e.is_live()]
 
     cross_community = 0
     node_cid = store.get_node_community_id(node.id)
@@ -356,7 +356,9 @@ def analyze_changes(
         # parser, so a changed production function finds its tests by source.
         # See: #515
         tested = store.get_edges_by_source(node.qualified_name)
-        if not any(e.kind == "TESTED_BY" for e in tested):
+        if not any(
+            e.kind == "TESTED_BY" and e.is_live() for e in tested
+        ):
             test_gaps.append({
                 "name": _sanitize_name(node.name),
                 "qualified_name": _sanitize_name(node.qualified_name),
--- a/code_review_graph/enrich.py
+++ b/code_review_graph/enrich.py
@@ -133,7 +133,7 @@ def _format_node_context(
     callers: list[str] = []
     seen: set[str] = set()
     for e in store.get_edges_by_target(qn):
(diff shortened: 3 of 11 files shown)


HouMinXi:
@tirth8205 Hi Tirth, could you take a look at this approach? The key design decision is: instead of tagging dead edges with `reachable: False` and asking every consumer to filter them (PR #580), we skip emitting dead CALLS edges at the parser level entirely, and add `is_live()` filtering to all consumers for backward compatibility with old graphs.

A few things I'd like your input on:

1. **Omit vs tag** — Is dropping dead edges at parse time acceptable, or would you prefer keeping them tagged for tooling that needs to inspect dead code paths?

2. **SQL filtering strategy** — The `COALESCE(json_extract(extra, '$.reachable'), 1) != 0` pattern is applied in 8 SQL locations. Would you prefer a dedicated `live` column instead?

3. **Scope** — TYPE_CHECKING attribute access (`typing.TYPE_CHECKING`) and aliases (`as TC`) are left as follow-ups. Is that OK for this PR, or should they be included?

Happy to iterate on whatever direction you prefer.

HouMinXi:
re-spin a new PR

PR #662, opened 2026-07-18
fix(search): preserve cross-repository result visibility

### LINKED ISSUE #619: [Bug]: cross_repo_search_tool drops one repo's results — incomparable score scales merged, then truncated to per-repo limit
### Summary

`cross_repo_search_tool` can silently drop **all** results from one repo, even when that repo's graph contains exact-name matches. Two compounding causes: results from different repos carry incomparable score scales, and the merged list is truncated to the *per-repo* `limit`.

Version: 2.3.6 (pipx, Python 3.12, macOS).

### Reproduction

Two registered repos (an Android/Kotlin app and an iOS Swift/ObjC app), both with fresh graphs and rebuilt FTS indexes, no embeddings.

```
cross_repo_search_tool(query="splash", limit=15)
→ "Found 22 result(s) across 2 repo(s)" — but all 15 returned results are from the iOS repo.
```

The Android repo has 7 matching nodes (`SplashWelcomeScreen`, a class literally named `Splash`, etc.) — single-repo `semantic_search_nodes_tool(query="splash")` finds all of them. They only appear in cross-repo output with `limit=30`, ranked *below every iOS result*.

### Root cause

**1. The same query takes different search paths per repo, producing incomparable scores** (`search.py`, `hybrid_search`, ~L308):

- Repo A (Android): FTS5 matches (the repo happens to have a node named exactly `Splash` and a `splash/` directory path segment) → results go through `rrf_merge` (~L150) → scores ≈ `1/(60+rank)` ≈ **0.015–0.033**.
- Repo B (iOS): FTS5 returns **0 rows** — `_fts_search` (~L192) wraps the query in double quotes (exact token match) and identifiers are indexed as single tokens, so `splash` does not match `SplashViewController` (token: `splashviewcontroller`). Verified directly: `SELECT count(*) FROM nodes_fts WHERE nodes_fts MATCH '"splash"'` → 7 on repo A, 0 on repo B. With no embeddings either, `hybrid_search` falls back to `_keyword_search` (~L255), which does `LIKE '%splash%'` and assigns **integer scores 1.0–3.0**.

**2. `cross_repo_search_tool` merges by raw score and truncates** (`tools/registry_tools.py`, ~L110–117):

```python
all_results.sort(key=lambda r: r.get("score", 0), reverse=True)
...
"results": all_results[:limit],
```

Every keyword-fallback score (≥ 1.0) beats every RRF score (≤ 0.033), so the repo whose FTS *worked* is buried; the global `[:limit]` slice then removes it from the output entirely — even though `limit` is documented as "Maximum results per repo".

The ironic effect: the better-indexed repo loses.

### Related symptom

Multi-word queries (e.g. `"splash screen launch"`) return 0 across all repos: FTS treats the quoted string as an exact phrase, and the keyword fallback requires *every* word as a substring of `name`/`qualified_name` (AND logic).

### Suggested fixes

1. In `cross_repo_search_tool`, merge per-repo result lists by **rank** (e.g. RRF across repos), not raw score — the scales are only meaningful within one repo's search path.
2. Don't truncate the merged list to the per-repo `limit` — either return up to `limit × n_repos`, or guarantee each repo keeps its top-N.
3. Longer term: split camelCase/snake_case identifiers into sub-tokens at FTS index time, so token queries match compound identifiers and the LIKE fallback stops being the deciding factor between repos.

Happy to provide more detail from the reproduction if useful.

HsiangNianian:
## Linked issue

Closes #619

## What & why

`cross_repo_search_tool` merged raw scores from repository-local search paths that use different scales. A repository using keyword fallback could therefore outrank every FTS/RRF result, and the final global slice could hide another repository entirely.

This change interleaves results by each repository's local rank, uses registry order as a deterministic tie-breaker, and returns the complete set already bounded by the per-repository `limit`. Original scores and repository provenance are preserved. The public tool description now documents the ordering and response cardinality.

The related camelCase tokenization and multi-word query behavior remain out of scope.

## How it was tested

```bash
uv run pytest tests/test_registry.py -q
uv run ruff check code_review_graph/ tests/test_registry.py
```

- Registry tests: 27 passed
- Full suite: 1962 passed, 5 skipped, 2 xpassed
- Ruff: passed
- Mypy: no issues in 66 source files

## Checklist

- [x] Public tool documentation updated for the behavior change

### DIFF
--- a/code_review_graph/main.py
+++ b/code_review_graph/main.py
@@ -940,8 +940,10 @@ def cross_repo_search_tool(
 ) -> dict:
     """Search for code entities across all registered repositories.
 
-    Runs hybrid search on each registered repo's graph database and merges
-    the results by score. Register repos first with the CLI 'register' command.
+    Runs hybrid search on each registered repo's graph database and interleaves
+    results by repository-local rank. Equal ranks follow registry order, and up
+    to ``limit`` results per searched repo may be returned. Register repos first
+    with the CLI 'register' command.
 
     Args:
         query: Search string to match against node names.
--- a/code_review_graph/tools/registry_tools.py
+++ b/code_review_graph/tools/registry_tools.py
@@ -79,10 +79,10 @@ def cross_repo_search_func(
                 "results": [],
             }
 
-        all_results: list[dict[str, Any]] = []
+        ranked_results: list[tuple[int, int, dict[str, Any]]] = []
         searched_repos: list[str] = []
 
-        for repo_entry in repos:
+        for repo_index, repo_entry in enumerate(repos):
             repo_path = Path(repo_entry["path"])
             db_path = get_db_path(repo_path)
             if not db_path.exists():
@@ -95,10 +95,10 @@ def cross_repo_search_func(
                         store, query, kind=kind, limit=limit
                     )
                     alias = repo_entry.get("alias", repo_path.name)
-                    for r in results:
+                    for local_rank, r in enumerate(results):
                         r["repo"] = alias
                         r["repo_path"] = str(repo_path)
-                    all_results.extend(results)
+                        ranked_results.append((local_rank, repo_index, r))
                     searched_repos.append(alias)
                 finally:
                     store.close()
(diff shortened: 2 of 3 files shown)


rajpratham1 (approved):
This pull request provides a well-focused improvement to cross-repository search behavior by replacing global score-based ranking with repository-local ranking. Since relevance scores from different repositories are not directly comparable, the new approach produces more consistent and predictable search results while preserving visibility across repositories. The implementation updates the search logic, aligns the documentation with the new behavior, and includes dedicated test coverage to validate the ranking strategy. The changes are coherent, well-scoped, and no blocking issues were identified. LGTM.

PR #666, opened 2026-07-19
fix(parser): omit calls from dead guards in C/C++, Go, TS and JS

HouMinXi:
## Summary

`f84b7d1` was the right call, and using `ast` for Python gets things a
syntactic walk cannot: alias resolution, shadowing, and boolean
combinations. This PR does not touch that path, and does not try to
replace it. It only covers the languages `ast` cannot reach.

Every non-Python language still emits dead edges:

```
# on main (v2.3.7)
C    void c(){ #if 0  dead(); #endif  live(); }   -> ['dead', 'live']
Go   func c(){ if false { dead() }; live() }      -> ['dead', 'live']
TS   function c(){ if(false){dead();} live(); }   -> ['dead', 'live']
Py   if False: dead()  / live()                   -> ['live']       (already fixed)
```

`ast.parse()` raises `SyntaxError` on C, Go, TypeScript and Rust, and
`tree-sitter-language-pack` is the only multi-language parser in the
dependency set, so the non-Python path has to be a tree-sitter walk.
This PR adds one, alongside the Python pass rather than replacing it.
`_python_unreachable_call_positions` keeps ownership of Python.

## What it detects

| Language | Guard |
|---|---|
| Go | `if false { ... }` |
| TypeScript / JavaScript | `if (false) { ... }`, `if (0) { ... }` |
| C / C++ | `#if 0`, `#elif 0` |

The TS/JS condition arrives wrapped in a `parenthesized_expression` and
is unwrapped before matching; `0` is node type `number` there, not
Python's `integer`. The C/C++ walk deliberately ignores function and
class scope, because the preprocessor deletes the text before the
compiler sees it, so a call inside a whole function wrapped in `#if 0`
is dead too.

## What stays live

Only the consequence is dead. These keep their edges, and each has a
test:

- `else`, `#else`, `#elif` branches of a dead guard
- `if (true)`, `if (1)`, and any non-literal condition
- `if (a === false)`, `if (foo(false))`, which are not literal conditions
- calls after the dead block, and sibling declarations

Direction matters here: the risk worth guarding against is suppressing a
live call, so every ambiguous case is left live.

## Semantics for nested declarations

A function or class declared inside a dead branch is never evaluated, so
calls in its body are dead. That matches the Python pass, which already
drops calls inside a `def` or `class` under `if False:`. JS/TS class
declarations are not hoisted, so no reachable symbol is lost by treating
them this way.

## Scope

Numeric detection is the literal `0` only. `0x0` and `0.0` are equally
falsy but stay live: a missed guard just drops a suppression, whereas
evaluating arbitrary numeric literals risks the opposite error. Noted in
a comment at the check.

Not included: `#ifdef` / `#ifndef` (needs macro state, not available to
a syntactic walk), and Rust, which is ready but held back to keep this
diff reviewable.

## Tests

`tests/test_parser.py`, one per language plus:

- `test_dead_guard_covers_declarations_nested_in_dead_branch`:
  nested-declaration semantics above.
- `test_dead_guard_calls_absent_from_graph_store`: builds a real
  `GraphStore` from each fixture and asserts the dead target never
  appears in `get_all_call_targets()`, mirroring the consumer-level
  check in `test_python_reachability.py`.

Each test was verified by injecting the bug it targets, confirming the
failure, then reverting; same drill for the 24 direct helper tests added
in bc1936e (gate disabled: the integration tests fail; helper broken:
the literal and fixture tests fail; restored: all green).

Full suite at bc1936e: 1989 passed, 5 skipped, 2 xpassed, plus one
failure in test_embedding_initialization that is pre-existing: it fails
identically at the v2.3.7 base in a clean worktree, and only on Python
3.14 Linux, which the CI matrix does not run. The 1966 figure I first
posted was from a run on a different branch; corrected.
Python behavior is unchanged; the ast path is untouched.

## Questions for you

I would rather match your design than land my own, so please push back
on any of this:

1. **Placement.** I put the helpers next to
   `_python_unreachable_call_positions` so the two reachability passes
   sit together, and added a single gate in `_extract_calls`. If you
   would prefer them in their own module, or the gate somewhere else,
   say the word and I will move them.
2. **Scope.** Is per-language literal matching the shape you want, or
   would you rather have one small table of "statically false condition"
   node types per language that this and future languages read from? I
   kept it inline because there are only three shapes today, but I am
   happy to build the table if you expect more languages.
3. **Rust.** It is written and tested (`if false`, and `cfg!(any())`
   distinguished from `cfg!(any(feature = "x"))`), held back only to
   keep this diff small. Want it here, as a follow-up, or not at all?
4. **`#ifdef` / `#ifndef`.** Left out because it needs macro state
   rather than syntax. If you think an approximation is worth having
   for the common `#ifdef UNDEFINED_MACRO` case, I will look at it.
5. **Anything you would rather solve differently.** If you already have
   a direction in mind for non-Python reachability, I am glad to drop
   this and implement yours instead.

Happy to split, rebase, or rework this however suits the project.

### DIFF
--- a/code_review_graph/parser.py
+++ b/code_review_graph/parser.py
@@ -322,6 +322,111 @@ def _python_unreachable_call_positions(
     visitor.visit(tree)
     return frozenset(visitor.positions)
 
+
+# ---------------------------------------------------------------------------
+# Non-Python static dead-guard detection (tree-sitter ancestor walk).
+#
+# ``_python_unreachable_call_positions`` above uses the ``ast`` module and so
+# only covers Python.  The helpers below cover languages ``ast`` cannot parse,
+# walking the tree-sitter ancestor chain of a call node:
+#
+#   * Go / TypeScript / JavaScript:  ``if false { ... }`` / ``if (0) { ... }``
+#   * C / C++:                       ``#if 0`` / ``#elif 0`` preprocessor blocks
+#
+# Only the consequence (true branch) is dead; ``else`` / ``#else`` / ``#elif``
+# branches stay live.
+# ---------------------------------------------------------------------------
+
+
+def _node_is_in_child(node, child_node) -> bool:
+    """Return True if *node* is *child_node* or one of its descendants.
+
+    Compares byte ranges because the tree-sitter Python bindings create a
+    fresh ``Node`` object on every ``child_by_field_name`` call, so ``is``
+    identity fails even when both sides refer to the same tree node.
+    """
+    start_byte, end_byte = child_node.start_byte, child_node.end_byte
+    cursor = node
+    while cursor is not None:
+        if cursor.start_byte == start_byte and cursor.end_byte == end_byte:
+            return True
+        cursor = cursor.parent
+    return False
+
+
+def _is_statically_false_condition(cond) -> bool:
+    """Return True if *cond* is a statically-false literal (non-Python).
+
+    * ``parenthesized_expression`` (TS/JS wrap ``if (expr)``) is unwrapped.
+    * ``false`` -- Go and TS/JS boolean literal.
+    * ``number`` equal to ``0`` -- TS/JS ``if (0)``.
+    """
+    if cond.type == "parenthesized_expression":
+        inner = cond.named_children
+        return _is_statically_false_condition(inner[0]) if inner else False
+    if cond.type == "false":
+        return True
+    # Literal ``0`` only. ``0x0`` / ``0.0`` are equally falsy but are left
+    # undetected on purpose: missing one is a dropped suppression, never a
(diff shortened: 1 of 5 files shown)


rajpratham1 (approved):
This pull request introduces a well-designed enhancement for detecting and excluding calls located in statically unreachable code across C/C++, Go, TypeScript, and JavaScript. The parser changes are focused on dead-guard detection without impacting live execution paths, and the implementation is supported by comprehensive fixture-based, integration, and helper-level unit tests covering multiple edge cases. The changes improve the accuracy of generated call graphs while maintaining consistency across supported languages. No blocking issues were identified during review. LGTM and approved.

HouMinXi:
@tirth8205
Do you have any comments regarding this patch?

PR #672, opened 2026-07-19
feat(parser): capture JS/TS member-assigned function expressions

avzamelek:
## Problem
Prototype- and module-augmentation assignments — `app.handle = function () {}`,
`Router.prototype.dispatch = () => {}`, `exports.render = function () {}` — are how
Express, Koa and many older JS/TS modules define their entire public API. The parser
only captured `const x = fn` (`variable_declarator`) and class fields, so these members
produced **no Function node at all** — invisible to `callers_of`, impact, dead-code and
search. This is part of the README's "JavaScript … flow detection needs work", and
matches the note in `eval/configs/express.yaml` ("JS modules use prototypes +
module.exports heavily, so most methods are not represented").

## Change
Add `_extract_js_member_functions`, mirroring the existing `_extract_js_var_functions`:
an `assignment_expression` whose LHS is a `member_expression` and whose RHS is an
arrow/function value becomes a Function node qualified by its full member path
(`file::app.handle`, the same shape as `file::Class.method`), with a `CONTAINS` edge and
body-call recursion. Computed access (`obj[key] = fn`) is skipped.

## Impact
On `expressjs/express` this adds **89 Function nodes (1771 → 1860)**, surfacing the
prototype-augmented public API (`app.use`, `app.handle`, `app.route`, `app.param`,
`req.accepts*`, `res.*`). Scope is structural coverage only; it does **not** by itself
move `flow_completeness` recall, which is gated on JS member-call resolution. `app.get`/
`app.post` remain uncaptured because express defines them via a computed
`methods.forEach(m => app[m] = …)` loop, which this handler intentionally skips.

## Tests
`TestJsMemberAssignedFunctions` in `tests/test_parser.py` (6 cases; fail before, pass
after). Full suite green (`1965 passed`); `ruff check code_review_graph/` clean.

### DIFF
--- a/code_review_graph/parser.py
+++ b/code_review_graph/parser.py
@@ -5422,6 +5422,18 @@ class CodeParser:
             ):
                 continue
 
+            # --- JS/TS member-assigned functions (app.handle = function () {}) ---
+            if (
+                language in ("javascript", "typescript", "tsx")
+                and node_type == "expression_statement"
+                and self._extract_js_member_functions(
+                    child, source, language, file_path, nodes, edges,
+                    enclosing_class, enclosing_func,
+                    import_map, defined_names, _depth,
+                )
+            ):
+                continue
+
             # --- Functions ---
             if node_type in func_types and self._extract_functions(
                 child, source, language, file_path, nodes, edges,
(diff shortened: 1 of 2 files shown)


avzamelek:
Follow-up added: static same-file JS/TS member calls now resolve to member-assigned function definitions.
For example, after app.handle = function () {}, a later app.handle() now creates a CALLS edge to file::app.handle instead of the bare handle target.
The resolver keeps the existing fallback for unresolved external calls such as thirdParty.handle().
Added regression coverage; tests/test_parser.py passes (112 passed).
Dynamic/computed assignments (app[m] = fn) and alias/destructuring resolution remain intentionally out of scope for this PR.

avzamelek (commented):
Follow-up added for optional chaining.
app?.handle() is now normalized to the same static member path as app.handle(), so when app.handle = fn exists in the same file, the CALLS edge resolves to file::app.handle.
Computed optional access such as app?.[key]() remains intentionally unresolved.
Added regression coverage; tests/test_parser.py passes (113 passed).

tirth8205:
The seven focused tests pass, but function-local member assignments are conflated. Parsing function a(){ x.run=function(){} } and function b(){ x.run=function(){} } creates two functions with the same file::x.run qualified name and duplicate containment targets. Please include the enclosing function scope, or restrict extraction to safe module/class/prototype contexts, and add this collision regression.

avzamelek:
[4e6793f](https://github.com/tirth8205/code-review-graph/pull/672/commits/4e6793f9cad5700eee1bf135d09a8533f10508c6) 

Addressed member-assigned function extraction is now limited to module/class contexts, so function-local assignments no longer collide as file::x.run. Added a regression test for sibling local assignments.. tests/test_parser.py passes (114 passed)

tirth8205:
Revalidated on current `main` at `9fed471`: the branch cleanly merges and 27/27 targeted member-assignment plus current-main PHP scoped-call tests pass on the merged tree. The function-local regression is fixed, but two identity bugs remain.

Two sibling lexical blocks, each declaring its own `const x` and assigning `x.run = function(){}`, produce two Function nodes and two containment edges with the identical `file::x.run` target. Dynamic receivers are also treated as static: `factory().handle = function(){}` is indexed as a stable member and a later `factory().handle()` resolves to it, although separate factory calls may return different objects.

Please restrict accepted member paths to stable identifier/dotted paths in direct safe module/prototype contexts, or carry lexical block identity, and add these two negative regressions. After that, this feature is suitable to merge.

tirth8205:
The corrected integration is now in #768. All four contributor commits and attribution are preserved.

It adds the missing negative cases for sibling lexical blocks and dynamic receivers, plus a real full-build persistence/call-resolution regression. The rebased branch passed 867 broad parser/graph tests (1 skipped) and all 147 parser tests after the final end-to-end addition. Full hosted CI is running now.

I am keeping this original PR open until the corrected integration passes the complete Python, Windows, type, lint, schema, security, and CodeQL matrix.

tirth8205:
Thank you, @Avza. This contribution is now integrated and merged through #770 at `329dbb91`.

Your four patches and author attribution are preserved in the merged history. During validation we added a stability correction so function-local, sibling-block, and dynamic member assignments do not receive unstable graph identities, plus an end-to-end graph storage/caller regression.

The corrected standalone integration #768 passed its full hosted matrix. The exact combined #770 state then passed Python 3.10–3.13, Windows, lint, type-check, security, schema sync, review, CodeQL, and GitGuardian before merge.

PR #673, opened 2026-07-19
fix(flows): rematch incremental flows across node replacement

### LINKED ISSUE #569: Incremental flow re-detection never finds new entry points (relative vs absolute path mismatch)

helasaoudi:
## Summary
- Capture and clear affected flows before file node IDs churn during incremental updates
- Remap community assignments by qualified name and purge orphan flow memberships after replacement
- Normalize relative git paths against absolute graph `file_path` values so new entry points are detected

Fixes #569

## Test plan
- [x] `tests/test_flows.py` relative-vs-absolute new entry point + replacement lifecycle
- [x] `tests/test_incremental.py` end-to-end incremental flow lifecycle
- [x] `tests/test_communities.py` community capture/remap regression
- [x] Existing flow/community/incremental suites (`143` related tests passed locally)

### DIFF
--- a/code_review_graph/communities.py
+++ b/code_review_graph/communities.py
@@ -9,7 +9,9 @@ from __future__ import annotations
 import logging
 import random
 import re
+import sqlite3
 from collections import Counter, defaultdict
+from pathlib import Path
 from typing import Any
 
 from .graph import GraphEdge, GraphNode, GraphStore, _sanitize_name
@@ -846,10 +848,111 @@ def detect_communities(
     return results
 
 
+def capture_community_assignments(
+    store: GraphStore,
+    file_paths: list[str],
+) -> dict[str, int]:
+    """Capture ``qualified_name -> community_id`` for nodes in *file_paths*.
+
+    Must run **before** node replacement so assignments can be remapped onto
+    the new node rows after IDs churn (#569).
+    """
+    if not file_paths:
+        return {}
+
+    conn = store._conn
+    mapping: dict[str, int] = {}
+    for i in range(0, len(file_paths), _SQL_BATCH):
+        batch = file_paths[i:i + _SQL_BATCH]
+        placeholders = ",".join("?" * len(batch))
+        rows = conn.execute(
+            f"SELECT qualified_name, community_id FROM nodes "  # nosec B608
+            f"WHERE community_id IS NOT NULL AND file_path IN ({placeholders})",
+            batch,
+        ).fetchall()
+        for qn, cid in rows:
+            if qn and cid is not None:
+                mapping[qn] = int(cid)
+    return mapping
+
+
+def remap_community_assignments(
+    store: GraphStore,
+    community_by_qn: dict[str, int],
+) -> int:
+    """Re-apply captured community IDs onto current nodes by qualified name.
+
+    Returns the number of nodes updated.
+    """
+    if not community_by_qn:
+        return 0
+
+    conn = store._conn
+    updated = 0
+    items = list(community_by_qn.items())
+    if conn.in_transaction:
+        conn.commit()
+    conn.execute("BEGIN IMMEDIATE")
+    try:
+        for qn, cid in items:
+            cur = conn.execute(
+                "UPDATE nodes SET community_id = ? WHERE qualified_name = ?",
+                (cid, qn),
+            )
+            if cur.rowcount:
+                updated += cur.rowcount
+        conn.commit()
+    except BaseException:
+        conn.rollback()
+        raise
+    return updated
+
+
+def purge_empty_communities(store: GraphStore) -> int:
+    """Delete community rows that no longer have any assigned nodes.
+
+    Also refreshes ``size`` for communities that still have members so
+    incremental deletion stays consistent with a fresh full rebuild.
+    Returns the number of empty community rows removed.
+    """
+    conn = store._conn
+    if conn.in_transaction:
+        conn.commit()
+    conn.execute("BEGIN IMMEDIATE")
+    try:
+        # Refresh sizes from live node assignments.
+        conn.execute(
+            "UPDATE communities SET size = ("
+            "SELECT COUNT(*) FROM nodes "
+            "WHERE nodes.community_id = communities.id)"
+        )
+        cur = conn.execute(
+            "DELETE FROM communities WHERE id NOT IN ("
+            "SELECT DISTINCT community_id FROM nodes "
+            "WHERE community_id IS NOT NULL)"
+        )
+        deleted = cur.rowcount if cur.rowcount is not None and cur.rowcount >= 0 else 0
+        # Drop summary rows for removed communities when the table exists.
+        try:
+            conn.execute(
+                "DELETE FROM community_summaries WHERE community_id NOT IN ("
+                "SELECT id FROM communities)"
+            )
+        except sqlite3.OperationalError:
+            pass
+        conn.commit()
+    except BaseException:
+        conn.rollback()
+        raise
+    return deleted
+
+
 def incremental_detect_communities(
     store: GraphStore,
     changed_files: list[str],
     min_size: int = 2,
+    repo_root: str | Path | None = None,
+    force: bool = False,
 ) -> int:
     """Re-detect communities only if changed files affect existing communities.
 
@@ -861,19 +964,26 @@ def incremental_detect_communities(
         store: The GraphStore instance.
         changed_files: List of file paths that have changed.
         min_size: Minimum number of nodes for a community to be included.
+        repo_root: Optional repo root used to expand relative/absolute paths.
+        force: When True, always re-run detection (used after pre-replacement
+            capture showed communities were affected even if path matching
+            would otherwise miss them).
 
     Returns:
         Number of communities detected, or 0 if skipped.
     """
-    if not changed_files:
+    if not changed_files and not force:
         return 0
 
+    from .flows import expand_changed_file_paths
+
+    expanded = expand_changed_file_paths(store, changed_files, repo_root=repo_root)
     conn = store._conn
 
     # Check if any communities are affected (batch to stay under SQLite limit)
     affected_count = 0
-    for i in range(0, len(changed_files), _SQL_BATCH):
-        batch = changed_files[i:i + _SQL_BATCH]
+    for i in range(0, len(expanded), _SQL_BATCH):
+        batch = expanded[i:i + _SQL_BATCH]
         placeholders = ",".join("?" * len(batch))
         row = conn.execute(
             f"SELECT COUNT(DISTINCT community_id) FROM nodes "  # nosec B608
@@ -882,9 +992,8 @@ def incremental_detect_communities(
         ).fetchone()
         if row:
             affected_count += row[0]
-    affected = (affected_count,) if affected_count else None
 
-    if not affected or affected[0] == 0:
+    if not force and affected_count == 0:
         return 0  # No communities affected, skip
 
     # Re-run full community detection (correct and fast enough)
--- a/code_review_graph/communities.py
+++ b/code_review_graph/communities.py
@@ -891,7 +891,8 @@ def remap_community_assignments(
     updated = 0
     items = list(community_by_qn.items())
     if conn.in_transaction:
-        conn.commit()
+        logger.warning("Rolling back uncommitted transaction before BEGIN IMMEDIATE")
+        conn.rollback()
     conn.execute("BEGIN IMMEDIATE")
     try:
         for qn, cid in items:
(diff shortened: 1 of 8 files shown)


tirth8205:
This is a useful lifecycle fix, but the four new focused tests are not enough for a 646-line stateful change. Path expansion uses suffix and SQL LIKE matching, yet there is no coverage for deleted or renamed entry points, different paths sharing a suffix, or literal % and _ in filenames. Please add those cases and prove incremental flow/community state matches a full rebuild before merge.

tirth8205:
Revalidated head `98239a7` against current `main` at `9fed471`; this remains blocked and still conflicts in `code_review_graph/incremental.py`. The four focused lifecycle tests pass on the PR head, but deletion parity and path matching still fail.

After a full build creates a two-node community for an isolated file, deleting that file and running the incremental full-postprocess path leaves the old community row at `size=2` with zero assigned nodes; a fresh full rebuild correctly has no communities. Deleted nodes are removed before affected state is captured, and replacement paths include only files being parsed.

Path expansion also uses unescaped SQL `LIKE`: `a_b.py` matches `acb.py`, and `a%z.py` matches `axz.py`. Please rebase onto the current lifecycle code, capture deleted and renamed old paths before removal, replace suffix-LIKE matching with exact normalized matching or escape `%` and `_`, and add deletion/rename/suffix-collision/literal-character tests asserting incremental state equals a fresh full rebuild for flows, memberships, community rows, and assignments.

tirth8205:
Revalidated head `98239a7` against exact current main `329dbb91`. It still conflicts in `code_review_graph/incremental.py`, directly overlapping the newly merged last-synced-base work. The four submitted tests pass, but the missing lifecycle cases still fail.

- `get_files_matching("a_b.py")` also returns `acb.py`, and `"a%z.py"` also returns `axz.py`, because SQL `LIKE` wildcards are unescaped.
- Deleting an isolated two-node file leaves a community row at `size=2` with zero assigned nodes.
- Linked #569 still reproduces: `incremental_trace_flows(store, ["b.py"])` finds zero flows when stored paths are absolute; passing the absolute path finds one.

Please rebase, capture deleted/renamed old paths before removal, use exact normalized or correctly escaped path matching, and add incremental-vs-clean-rebuild parity for flows, memberships, community rows, and node assignments. This head is not mergeable.

helasaoudi:
Thanks for the detailed revalidation addressed on the latest head 97caf68 (rebased onto current main).

Conflicts
Resolved the code_review_graph/incremental.py overlap with the last-synced-base work and force-pushed the rebased branch. The PR should now be mergeable from a conflict perspective.

Deletion / rename lifecycle
Affected flow and community state is now captured before permanent removal for deleted, renamed, and stale paths (not only to_parse). After replacement/removal we:

remap surviving community assignments by qualified name
purge orphan flow memberships / dead entry_point_ids
remove empty community rows so incremental deletion no longer leaves size=2 communities with zero assigned nodes
Path matching
Unescaped suffix-LIKE expansion is replaced with exact normalized path variants plus escaped LIKE (ESCAPE '\'), so:

a_b.py no longer matches acb.py
a%z.py no longer matches axz.py
incremental_trace_flows(store, ["b.py"]) resolves absolute stored file_paths (including without an explicit repo_root)
Tests added

relative changed files without repo_root
literal _ / % filename matching
shared-basename suffix collision
deleted isolated community cleanup
rename vs fresh full-rebuild parity for flows, memberships, and community assignments
Locally, the focused lifecycle suites and related flow/community/incremental tests are green. CI checks are waiting on maintainer workflow approval — happy to follow up if anything still fails after that.

tirth8205:
Verified head 97caf68 locally: the rebase resolved the incremental.py conflict, the full suite passes (2314 passed, 5 skipped, 2 xpassed), ruff is clean, and the #569 repro is fixed: incremental_trace_flows(store, ["b.py"]) now finds the flow when stored paths are absolute. Two items remain before merge.

1. The Windows-separator LIKE pattern in expand_changed_file_paths is wrong. f"%\\{escaped}" produces the SQL pattern %\<first-char>..., and under ESCAPE '\' that backslash escapes the first character of the filename instead of matching a literal backslash. The pattern degenerates to an unanchored suffix match. Verified: expand_changed_file_paths(store, ["b.py"]) also returns /repo/ab.py, and ["utils.py"] also returns /repo/data_utils.py, with or without repo_root. Unrelated files that merely share a basename suffix get their flows cleared and re-traced on every update, and with postprocess="none" their flows are deleted without being re-traced. Use an escaped backslash (%\\ in SQL) for the separator anchor or drop that pattern and normalize stored backslash paths, and add a basename-suffix collision test (b.py vs ab.py).

2. The community parity assertions are partly vacuous. test_incremental_delete_clears_orphan_community_rows computes clean_ids from the fresh rebuild and then discards it (_ = clean_ids), and test_incremental_rename_parity_with_full_rebuild ends with assert _assigned(store) >= 0, which is always true. Please assert real equality of community assignments between the incremental store and the clean rebuild for the deletion and rename cases.

Minor: incremental_update builds lifecycle_inputs variants that duplicate what expand_changed_file_paths already does, so paths are expanded multiple times per update, and clear_flows_for_files re-runs the node/flow ID queries that capture_affected_flow_entry_points just executed. Worth collapsing while you are in there.

helasaoudi:
Thanks for the revalidation on 97caf68 — the two remaining items (plus the minor cleanup) are addressed on head 6450e28
(rebased onto current main, conflicts clear).

1. Windows-separator LIKE / basename-suffix collisions

Separator pattern is now a correctly escaped literal backslash: %\\… under ESCAPE '\' (f"%\\\\{escaped}" in Python).
Path expansion also goes through shared POSIX normalize_file_path (#774), so matching uses the same identity as stored graph paths.
Added coverage that b.py does not match ab.py, and utils.py does not match data_utils.py (with and without repo_root).
2. Community parity asserts

test_incremental_delete_clears_orphan_community_rows and test_incremental_rename_parity_with_full_rebuild now assert real equality of community assignments / signatures between the incremental store and a fresh full rebuild (no discarded clean_ids, no vacuous >= 0).
3. Minor: collapse duplicate work

incremental_update passes raw lifecycle inputs once through expand_changed_file_paths.
clear_flows_for_files returns entry-point QNs from a single _flows_touching_files query (no separate capture + clear round-trip).
Focused lifecycle / path-matching tests pass locally. Happy to follow up if anything still fails after CI runs on this head.

tirth8205:
Verified head 6450e28 against current main: merge is clean, full suite passes (2408 passed, 5 skipped, 2 xpassed), ruff is clean, and all ten new tests fail without the fix. Two items block merge.

1. mypy fails: flows.py:738 reuses `batch` across the expanded-paths loop (list[str]) and the affected-ids loop (list[int]). The type-check CI job rejects this. Rename one loop variable.

2. expand_changed_file_paths still over-matches. The %/name LIKE fallback runs even when repo_root is provided, so a changed root-level util.py also matches an unchanged left/util.py. Repro: both files hold flows, then incremental_update(changed_files=["util.py"]) with postprocess none deletes the left/util.py flow and nothing re-traces it. Skip the LIKE fallback when repo_root is known; that also removes three unindexed scans per changed file, about 2s per 300 files on a 50k-node graph.

Minor: the new helpers commit an ambient transaction before BEGIN IMMEDIATE while store_flows rolls back; pick one convention.

helasaoudi:
Thanks for the revalidation on `6450e28` — the two blockers (plus the minor transaction note) are addressed on head `bd17845`.

**1. mypy `batch` reuse**
- Renamed the loop variables in `incremental_trace_flows` so path expansion uses `path_batch: list[str]` and affected flow IDs use `id_batch: list[int]` (no shared `batch` across incompatible types).

**2. `expand_changed_file_paths` over-match with `repo_root`**
- The `%/name` LIKE fallback now runs only when `repo_root` is unknown.
- With `repo_root` set, matching uses exact absolute/relative variants only, so root-level `util.py` no longer matches `left/util.py`.
- Added coverage:
  - `test_expand_with_repo_root_skips_nested_same_basename`
  - `test_incremental_root_basename_does_not_clear_nested_flows` (incremental update / postprocess-none path)

**3. Minor: ambient transaction convention**
- New helpers now roll back ambient transactions before `BEGIN IMMEDIATE`, matching `store_flows` / `store_communities` (with the same warning).

Locally, the focused lifecycle/path-matching suites, mypy on the touched modules, and ruff on the changed files are green. Happy to follow up if anything still fails after CI on this head.

tirth8205:
Changes required: a failed replacement parse removes a valid existing flow while retaining its old nodes. With `CRG_SERIAL_PARSE=1`, build a `routes.py` containing `def handler(): return helper()` and `def helper(): return 1` using `full_build`, save `trace_flows` with `store_flows`, append another function, then call `incremental_update(..., changed_files=['routes.py'])` while `CodeParser.parse_bytes` raises `RuntimeError`; `get_flows` changes from one flow to an empty list on this PR. Preserve flow and community state until replacement parsing succeeds.

tirth8205:

```bash
# resolve, then
uv run pytest tests/ -q
git push
```

When resolving:
- Rebase onto current origin/staging and resolve code_review_graph/incremental.py: keep staging's known_text/stored_paths=True/content-mismatch flow and layer the PR's remove=False deferral, stale-file union, and remap/purge calls on top of it.
- After rebase, re-run the full gate set (pytest 3.13, ruff, mypy) and the edge-case review; none of that was assessed here.

tirth8205:
This no longer merges into `staging`. 2 files conflict and the branch is 319 commits behind:

- `code_review_graph/tools/build.py`

```bash
git fetch origin
git merge origin/staging
# resolve, then
uv run pytest tests/ -q
git push
```

PR #626, opened 2026-07-17
fix(windows): validate Codex hooks through cmd.exe

tirth8205:
## Latest refresh — 18 July 2026

- rebased patch-identically onto main 0a3bd6c; current head bdc95c9
- all 15 hosted checks passed after this rebase, including the remote Windows job
- released Codex-on-Windows client validation remains required; this stays draft and is not mergeable by policy## Status

**Draft — do not merge until the released Codex Windows-client checks below pass.**

## Change

This keeps the safe Windows hook work from source PR #621, with Adam White's original authorship preserved. It adds native `commandWindows` bodies for `PostToolUse` and `SessionStart`, drains the full stdin event, upgrades matching Unix-only CRG hooks in place, and preserves unrelated user hooks.

## Why source PR #621 was closed

The source double-wrapped `cmd.exe`, did not upgrade an older Unix-only CRG configuration, and tested strings instead of executing the generated Windows commands. This draft fixes those gaps. Source PR #621 is closed as superseded.

## Automated validation

- patch-equivalent rebase onto current `main` (`0a3bd6c`); current head `bdc95c9`
- focused Codex-hook tests: **12 passed, 4 Windows-only skipped locally**
- complete non-daemon suite: **1,899 passed, 9 skipped, 2 expected deselections, 2 xpassed**
- both multiprocessing parity checks: **2 passed**
- Ruff and `git diff --check`: clean
- GitHub CI run 29619153235: every job passed, including native Windows execution on Windows Server 2025
- PR Review run 29619153205: passed

Automated Windows CI proves the generated command bodies execute through `COMSPEC /C`, drain a 2 MiB event, fail open outside a Git repository, and upgrade the older configuration without duplication.

## Required real-client validation

Do not mark this ready or merge it until all items below are recorded with a released Codex client on Windows:

- [ ] reinstall over an older Unix-only `hooks.json` and confirm the client accepts the installed hooks without an untrusted-hook or configuration error
- [ ] startup and resume invoke the `SessionStart` hook successfully
- [ ] a Write action invokes the `PostToolUse` hook successfully
- [ ] an Edit action invokes the `PostToolUse` hook successfully
- [ ] a Bash action invokes the `PostToolUse` hook successfully
- [ ] reinstall preserves unrelated user hooks and does not duplicate code-review-graph hooks

Addresses #620.
Carries PR #621 with original contributor attribution.

PR #675, opened 2026-07-20
fix(codex): emit commandWindows override for Windows hooks (fixes #620)

### LINKED ISSUE #620: Bug: Codex installer generates Unix-only hooks on Windows
## Description

Running `code-review-graph install --platform codex` on Windows generates Unix-only commands in `~/.codex/hooks.json`.

Codex then reports:

```text
PostToolUse hook (failed)
error: hook exited with code 1
```

## Environment

- OS: Windows
- Shell: PowerShell
- code-review-graph: 2.3.6
- Python: 3.14.4
- Installation: pipx
- Platform: Codex

## Steps to reproduce

1. Install with `pipx install code-review-graph`.
2. Run `code-review-graph install --platform codex`.
3. Restart Codex.
4. Run an operation that triggers the generated `PostToolUse` hook.

## Actual behavior

The generated `~/.codex/hooks.json` contains Unix shell commands:

```json
{
  "type": "command",
  "command": "cat >/dev/null || true; git rev-parse --git-dir >/dev/null 2>&1 && code-review-graph update --skip-flows || true"
}
```

This uses `cat`, `/dev/null`, and Unix shell chaining without a Windows-specific override. The generated `SessionStart` hook has the same problem.

## Expected behavior

On Windows, the generated hooks should include the Codex-supported `commandWindows` override. For example:

```json
{
  "type": "command",
  "command": "cat >/dev/null || true; git rev-parse --git-dir >/dev/null 2>&1 && code-review-graph update --skip-flows || true",
  "commandWindows": "powershell.exe -NoProfile -NonInteractive -Command \"[Console]::In.ReadToEnd() | Out-Null; code-review-graph update --skip-flows; exit 0\""
}
```

The `SessionStart` hook should similarly use:

```json
"commandWindows": "powershell.exe -NoProfile -NonInteractive -Command \"[Console]::In.ReadToEnd() | Out-Null; code-review-graph status; exit 0\""
```

## Root cause

`generate_codex_hooks_config()` generates only the Unix `command` field. It does not generate `commandWindows`.

## Workaround

Manually adding `commandWindows` to both generated hooks resolves the failure. After restarting Codex, both hooks exit successfully and the MCP graph loads correctly.

Alaa-Taieb:
## Root cause

`generate_codex_hooks_config()` in `code_review_graph/skills.py` builds `~/.codex/hooks.json` but only emitted the Unix `command` field for both `PostToolUse` and `SessionStart` hooks. On Windows, Codex runs the hook via the OS shell, where the Unix `command` string is invalid in PowerShell, so every hook fails with `hook exited with code 1` (issue #620).

Codex natively supports a per-hook `commandWindows` override that takes precedence on Windows. The function never emitted it.

## Fix

This is a small, additive change:

- Added a module-level helper constant `_CODEX_WINDOWS_HOOK_PREFIX = "powershell.exe -NoProfile -NonInteractive -Command "`.
- Each hook dict now also emits a `commandWindows` field that drains stdin (`[Console]::In.ReadToEnd() | Out-Null`), runs the matching CRG command, and `exit 0` so the hook never fails on payload size and always returns 0.
- The existing Unix `command` strings, `type`, `timeout`, and `statusMessage` are **untouched** — macOS/Linux behavior is completely unchanged.

## Tests

- Added two regression tests (`test_command_windows_present_for_both_hooks`, `test_command_windows_uses_powershell`) that assert the generated JSON contains `commandWindows` (PowerShell prefix + correct CRG command + `exit 0`) for both hooks.
- The test suite intentionally never spawns `powershell.exe`, so it passes on any OS CI runner.

## Verification

- `uvx ruff check code_review_graph/skills.py` -> clean.
- `uv run pytest tests/test_skills.py` -> 177 passed (only the unrelated `bash`-based stdin test is skipped on this Windows host; it passes on Linux/macOS CI).

Closes #620

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 ## [Unreleased]
 
+### Fixed
+- Fixed Codex hooks on Windows: `code-review-graph install --platform codex` now
+  emits a `commandWindows` override (PowerShell) so hooks no longer fail with
+  `hook exited with code 1` (fixes #620).
+
 ## [2.3.7] - 2026-07-18
 
 **Maintainer-reconciliation release.** This release packages the verified work
--- a/code_review_graph/skills.py
+++ b/code_review_graph/skills.py
@@ -21,6 +21,10 @@ from typing import Any
 
 logger = logging.getLogger(__name__)
 
+# Codex hook command used on Windows (PowerShell). Drains stdin so large
+# tool payloads never break the hook, then runs the CRG command and exits 0.
+_CODEX_WINDOWS_HOOK_PREFIX = "powershell.exe -NoProfile -NonInteractive -Command "
+
 
 # --- Multi-platform MCP install ---
 
@@ -788,6 +792,12 @@ def generate_codex_hooks_config(repo_root: Path) -> dict[str, Any]:
                                 " && code-review-graph update --skip-flows"
                                 " || true"
                             ),
+                            "commandWindows": (
+                                _CODEX_WINDOWS_HOOK_PREFIX
+                                + "[Console]::In.ReadToEnd() | Out-Null; "
+                                "code-review-graph update --skip-flows; "
+                                "exit 0"
+                            ),
                             "timeout": 30,
                             "statusMessage": "Updating code-review-graph",
                         },
@@ -805,6 +815,12 @@ def generate_codex_hooks_config(repo_root: Path) -> dict[str, Any]:
                                 "git rev-parse --git-dir >/dev/null 2>&1"
                                 " && code-review-graph status"
                                 " || echo 'Not a git repo, skipping'"
+                            ),
+                            "commandWindows": (
+                                _CODEX_WINDOWS_HOOK_PREFIX
+                                + "[Console]::In.ReadToEnd() | Out-Null; "
+                                "code-review-graph status; "
+                                "exit 0"
                             ),
                             "timeout": 10,
                             "statusMessage": "Checking code-review-graph status",
--- a/tests/test_skills.py
+++ b/tests/test_skills.py
@@ -545,6 +545,27 @@ class TestGenerateCodexHooksConfig:
         assert "code-review-graph update --skip-flows" in post_cmd
         assert "code-review-graph status" in session_cmd
 
+    def test_command_windows_present_for_both_hooks(self, tmp_path):
+        config = generate_codex_hooks_config(tmp_path)
+        post = config["hooks"]["PostToolUse"][0]["hooks"][0]
+        session = config["hooks"]["SessionStart"][0]["hooks"][0]
+        assert "commandWindows" in post
+        assert "commandWindows" in session
+
+    def test_command_windows_uses_powershell(self, tmp_path):
+        config = generate_codex_hooks_config(tmp_path)
+        post = config["hooks"]["PostToolUse"][0]["hooks"][0]["commandWindows"]
+        session = config["hooks"]["SessionStart"][0]["hooks"][0]["commandWindows"]
+        prefix = "powershell.exe -NoProfile -NonInteractive -Command "
+        assert post.startswith(prefix)
+        assert "code-review-graph update --skip-flows" in post
+        assert "code-review-graph status" in session
+        # stdin must be drained and the hook must always exit 0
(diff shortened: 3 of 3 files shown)


tirth8205:
The focused suite passes 195 tests, but a required upgrade reproduction still fails. When an existing CRG hook contains only the older Unix `command`, reinstall treats it as a duplicate and skips it, so `commandWindows` is never added.

Please upgrade matching entries in place and add this exact reinstall regression. The branch also needs a rebase and released Codex-on-Windows evidence showing both hooks are discovered and execute successfully after upgrading an existing configuration. This overlaps draft #626, where real-client testing already showed that shell assumptions matter, so the two implementations should be consolidated before merge.

tirth8205:
Revalidated this Windows-hook cluster against exact current main `329dbb91`. This head is not mergeable for two concrete reasons:

1. Reinstalling over current Unix-only CRG hook entries treats them as duplicates, so `commandWindows` remains missing.
2. The command body is not safely quoted for the released Codex PowerShell runner. Draft #626's `cmd.exe` assumption already failed in real Codex 0.144.5: `&` launched the stdin drain as a background PowerShell job and injected the job table into hook context.

The two branches overlap and need one implementation: upgrade owned entries in place and use a quoted PowerShell command that drains stdin synchronously and fails open. That combined shape passes local tests, but it still requires the Windows-only tests and a released Codex-on-Windows install/reinstall/session/PostToolUse/uninstall run before merge.

tirth8205:
Thanks - the root cause is correct and the generator change is cleanly additive, and the two regression tests do fail without the fix. Three things block merge: (1) install_codex_hooks dedups merged entries on the command field, so reinstalling over an existing Unix-only hooks.json skips the new entries and commandWindows is never added - I reproduced this locally with your branch; matching CRG-owned entries need to be upgraded in place, with a reinstall regression test. (2) The emitted commandWindows drops the quotes around the -Command body that the verified workaround in #620 includes; under the released Codex Windows runner the unquoted pipe is parsed by the outer shell, so please match the quoted shape or provide real Codex-on-Windows evidence that the unquoted form executes. (3) The branch conflicts with main (CHANGELOG.md) and needs a rebase, and this work overlaps draft #626, so the two should be consolidated into one implementation.

tirth8205:
Rechecked against current main (1a010de). The generator change still applies cleanly and your two tests do fail without the fix, but the same three things block merge.

1. Upgrade path: install_codex_hooks (skills.py, dedup around lines 1105-1124) keys on the command field, so reinstalling over an existing Unix-only hooks.json skips your entries and commandWindows is never added. That is exactly the situation of the reporter in #620. Matching CRG-owned entries need to be upgraded in place, with a reinstall regression test.

2. Quoting: the emitted commandWindows drops the quotes around the -Command body that the verified workaround in #620 uses. Under the released Codex Windows runner the unquoted pipe and semicolons are parsed by the outer shell. Match the quoted shape or attach real Codex-on-Windows output showing the unquoted form executing.

3. The branch conflicts with main in CHANGELOG.md and is about 200 commits behind, so please rebase. It also overlaps draft #626, which already has the upgrade-in-place logic, so the two should be consolidated into one implementation.

tirth8205:
Changes required: existing Codex users do not receive the new Windows override on reinstall. Reproduce by generating a config with `code_review_graph.skills.generate_codex_hooks_config`, removing every inner hook's `commandWindows`, writing it to the hooks.json location used by `install_codex_hooks`, and calling that installer again; the overrides remain absent because matching commands are skipped. Upgrade owned entries in place and verify the command quoting in the released Windows client before consolidating this with the other hook PRs.

tirth8205:
This no longer merges into `staging`. Conflicts in `CHANGELOG.md`.

```bash
# resolve, then
uv run pytest tests/ -q
git push
```

When resolving:
- Rebase the PR onto current staging and resolve the CHANGELOG.md [Unreleased] section conflict (staging's Unreleased block gained many entries since the PR branched); then re-run tests/ruff/mypy and the edge-case review.

tirth8205:
This no longer merges into `staging`. The one conflict is `CHANGELOG.md`, and the branch is 511 commits behind.

```bash
git fetch origin
git merge origin/staging
# resolve, then
uv run pytest tests/ -q
git push
```

PR #686, opened 2026-07-20
Fix incremental update leaving stale nodes after a file rename

### LINKED ISSUE #684: [Bug]: File rename leaves stale nodes/edges — incremental update diverges from full rebuild
### Summary

After a file rename, an incremental update leaves the old path's nodes and edges in the graph, so the incremental result diverges from a full rebuild.

- **Version:** 2.3.7
- **OS:** macOS (platform-independent — it's in the git-diff parsing)
- **Python:** 3.10+
- **Found via:** source review + a reproducing test (details below)

### Root cause

`get_changed_files()` (`code_review_graph/incremental.py:559`) discovers changes with:

```python
result = subprocess.run(["git", "diff", "--name-only", "-z", base, "--"], ...)
```

`git diff --name-only` reports a rename `a.py → b.py` as **only the new path** `b.py`. So `incremental_update()` never sees `a.py`.

`incremental_update()`'s purge loop (`incremental.py:1048`) only removes a path when it is present in the changed/dependent set **and** missing on disk:

```python
for rel_path in all_files:            # all_files = changed ∪ dependents
    ...
    abs_path = repo_root / rel_path
    if not abs_path.is_file():
        store.remove_file_data(str(abs_path))
```

The old path `a.py` is in neither `changed` nor `dependents`, so it is never purged. Unlike `full_build()` (`incremental.py:915`), which computes `existing_files - current_abs` and purges, `incremental_update()` has no "purge files no longer on disk" step. The old File/Function/Class nodes and every edge keyed to `a.py` survive.

### Steps to reproduce

1. Build a graph of a repo containing `a.py` (e.g. `def foo(): return 1`).
2. `git mv a.py b.py && git commit -m rename`
3. Run an incremental update across the rename (`incremental_update(repo, store)` with `base="HEAD~1"`).

### Expected vs actual

- **Expected:** the graph contains only `b.py::foo` — identical to a full rebuild of the post-rename tree.
- **Actual:** the graph contains **both** `a.py::foo` (stale, plus dangling importer edges) **and** `b.py::foo`. Node/edge counts exceed a full rebuild.

### Suggested fix

Use `git diff --name-status -z` and parse it so renames/copies (`R`/`C` records) emit **both** the old and new path. The old path then flows through the existing purge loop and is removed, making an incremental update equal a full rebuild. (Apply to the `--cached` fallback branch too.)

I have a fix + tests ready and will open a PR referencing this issue.

thejesh23:
## Linked issue

Closes #684

## What & why

`get_changed_files()` discovered changes with `git diff --name-only`, which reports a rename `a.py → b.py` as **only the new path** `b.py`. `incremental_update()` therefore never saw `a.py`, and its purge loop only removes paths that are in the changed/dependent set **and** missing on disk — the old path is in neither. Unlike `full_build()` (which purges `existing_files - current_abs`), the incremental path had no "purge files no longer on disk" step, so the renamed-away file's File/Function/Class nodes and every edge keyed to it survived. The incremental graph diverged from a full rebuild.

This switches change discovery to `git diff --name-status -z` and adds `_parse_name_status_z()`, which emits **both** the old and new path for rename/copy (`R`/`C`) records. The old path then flows through the existing purge loop and is removed, so an incremental update again equals a full rebuild. The `--cached` fallback branch gets the same treatment. The now-unused `_decode_nul_paths()` helper is removed.

## How it was tested

```bash
uv run pytest tests/test_incremental.py tests/test_integration_git.py -q   # 85 passed
uv run pytest tests/ --tb=short -q   # 1962 passed (1 pre-existing Windows/py3.14 failure, unrelated)
uv run mypy code_review_graph/incremental.py --ignore-missing-imports --no-strict-optional   # Success
```

New tests (both fail on `main`, pass here):
- `test_incremental.py::TestGitOperations::test_get_changed_files_reports_both_rename_paths` — unit test of the name-status parser.
- `test_integration_git.py::test_incremental_update_purges_renamed_source` — real-git test asserting the renamed-away source is purged and node/edge/file counts match a fresh `full_build`.

## Checklist

- [x] All tests pass: `uv run pytest tests/ --tb=short -q` (one pre-existing, unrelated Windows/py3.14 failure)
- [x] Type checking passes: `uv run mypy code_review_graph/incremental.py --ignore-missing-imports --no-strict-optional`
- [x] Docs updated where behavior changed (docstrings on the new/changed functions)

### DIFF
--- a/code_review_graph/incremental.py
+++ b/code_review_graph/incremental.py
@@ -534,9 +534,37 @@ _SAFE_GIT_REF = re.compile(r"^[A-Za-z0-9_.~^/@{}\-]+$")
 _SAFE_SVN_REV = re.compile(r"^r?\d+(:r?\d+|:HEAD|:BASE|:COMMITTED)?$", re.IGNORECASE)
 
 
-def _decode_nul_paths(output: bytes) -> list[str]:
-    """Decode Git's NUL-delimited path output without quote mangling."""
-    return [os.fsdecode(path) for path in output.split(b"\0") if path]
+def _parse_name_status_z(output: bytes) -> list[str]:
+    """Decode ``git diff --name-status -z`` output to a list of changed paths.
+
+    Unlike ``--name-only``, ``--name-status`` distinguishes renames/copies. Its
+    ``-z`` records are NUL-separated fields: ``<status>`` then the path(s).
+    A rename/copy (``R``/``C``) emits ``<status>\\0<old>\\0<new>``; every other
+    status emits ``<status>\\0<path>``.
+
+    Both the old *and* new path of a rename are returned, so the old path is
+    seen by :func:`incremental_update` and purged when it no longer exists on
+    disk — keeping an incremental update equivalent to a full rebuild (a plain
+    ``--name-only`` diff reports only the new path, leaving the old file's nodes
+    and edges stale).
+    """
+    fields = [os.fsdecode(f) for f in output.split(b"\0") if f]
+    paths: list[str] = []
+    i = 0
+    n = len(fields)
+    while i < n:
+        status = fields[i]
+        i += 1
+        # Rename (R) and copy (C) statuses carry a similarity score (e.g. "R100")
+        # and are followed by TWO paths (old, new); all others by a single path.
+        if status[:1] in ("R", "C") and i + 1 < n:
+            paths.append(fields[i])      # old path (renamed away / copy source)
+            paths.append(fields[i + 1])  # new path
+            i += 2
+        elif i < n:
+            paths.append(fields[i])
+            i += 1
+    return paths
 
 
 def _store_vcs_metadata(repo_root: Path, store: "GraphStore") -> None:
@@ -572,7 +600,7 @@ def get_changed_files(repo_root: Path, base: str = "HEAD~1") -> list[str]:
         return []
     try:
         result = subprocess.run(
-            ["git", "diff", "--name-only", "-z", base, "--"],
+            ["git", "diff", "--name-status", "-z", base, "--"],
             capture_output=True,
             cwd=str(repo_root),
             timeout=_GIT_TIMEOUT,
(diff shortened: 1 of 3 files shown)


jondkinney:
I reviewed this alongside #729, which fixes the same issue with a very similar approach — flagging that here in case it's helpful for the maintainer to compare them side by side, since they'd conflict if both merged.

Having read both, I think each has one thing the other would benefit from:

**This PR's tests are the stronger of the two.** `test_get_changed_files_reports_both_rename_paths` asserts the exact expected ordering (`["old/a.py", "new/b.py", "keep.py"]`), and adding the `--name-status` assertion to the existing test is a nice touch — it pins the behaviour rather than just the output. #729 doesn't have an equivalent rename-specific case.

**#729's parser implementation may be slightly more robust.** It slices (`fields[i + 1 : i + 3]`) rather than indexing, so a truncated final record can't misbehave, and it dedupes with a `seen` set. In this PR the `status[:1] in ("R", "C") and i + 1 < n` guard means a malformed trailing `R100` with only one following path falls through to the `elif` and gets treated as a normal single-path entry. That's a degenerate case that git shouldn't produce, so it may well not matter in practice — I mention it only because the two implementations differ there.

A combination — #729's parser with this PR's tests — would probably be the strongest result, if that's useful.

I verified the underlying assumption both PRs rest on, and it holds: `incremental.py:1053` does `if not abs_path.is_file(): store.remove_file_data(...)`, so once the old path is in the changed set it does get purged. Both correctly leave `D` records alone, since `--name-only` already reported deletions.

One scope note that applies to both: this fixes the git-diff path, but `start_watch_thread()`'s `GraphUpdateHandler` has no `on_moved` handler, so `--watch` users would still see stale nodes after a rename. I've opened #742 for that separately — not a gap in this PR, just adjacent.

Thanks for putting this together.

tirth8205:
This bug is now fixed and shipped through #729 at `5a1a2a2`. The merged implementation removes the renamed-away path and was validated with real committed-rename coverage plus exact incremental-vs-fresh-rebuild parity for files, nodes, and edges; 92 local tests and all 11 hosted checks passed. Your root-cause report and proposed old/new-path handling were correct. I am closing this overlapping implementation now that the selected fix is merged.

PR #694, opened 2026-07-21
fix: anchor incremental update to the last-synced commit

utkuozdemir:
Incremental update discovered changed files by diffing against a fixed HEAD~1, so it only looked one commit back. A per-commit hook keeps ordinary local edits and commits in sync, but commits that arrive another way never pass through it. After a multi-commit pull, a rebase, or a branch switch, a default update reconciled only the newest commit and left the earlier ones missing from the graph with no error and no warning. The docs already suggest running update after a rebase to refresh a stale graph, but that is exactly the path that under-caught, so the quick remedy did not fully work and only a full build reliably recovered.

The graph already records the commit it was last built at, so use that as the diff base when no base is given. A single update now catches everything since the graph was last in sync, which makes the documented post-rebase update behave as expected. When that stored commit is no longer a usable base (a fresh or legacy database, or a commit lost to a history rewrite or shallow clone) the update falls back to a full rebuild instead of a HEAD~1 diff that could report the graph as up to date while it is stale.

An explicit base is still honored verbatim, and SVN and non-git working copies keep their existing behavior. The change is applied at the shared build entry point and at the MCP tool and CLI defaults so every path benefits, including update with the change-impact summary.

This does not make routine updates more expensive. Because the base now tracks where the graph actually is rather than a fixed commit back, a normal update reparses the same or fewer files than before, and a single uncommitted edit examines one file instead of also re-examining the previous commit. The full rebuild path is reserved for the case where no usable base exists and never runs in steady state. The only heavier case is the first update after a large pull, rebase, or branch switch, which reparses exactly the files that changed since the graph was last in sync and were previously being dropped, and even then a content check skips files whose contents did not change.

### DIFF
--- a/code_review_graph/cli.py
+++ b/code_review_graph/cli.py
@@ -667,7 +667,11 @@ def main() -> None:
 
     # update
     update_cmd = sub.add_parser("update", help="Incremental update (only changed files)")
-    update_cmd.add_argument("--base", default="HEAD~1", help="Git diff base (default: HEAD~1)")
+    update_cmd.add_argument(
+        "--base",
+        default=None,
+        help="Git diff base (default: the commit the graph was last built at)",
+    )
     update_cmd.add_argument("--repo", default=None, help="Repository root (auto-detected)")
     update_cmd.add_argument("-q", "--quiet", action="store_true", help="Suppress output")
     update_cmd.add_argument(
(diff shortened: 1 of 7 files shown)


tirth8205:
The production approach is sound, but the branch is not merge-ready yet. Its divergent-branch test assumes git init creates main; with the standard master default, git checkout main fails unchecked and the test fails (36 passed, 1 failed). It passes only when Git is explicitly initialized on main (37 passed). Please use git init -b main or discover the branch, assert subprocess return codes, then rebase after #729 and add one end-to-end case where the stored SHA spans multiple commits including a rename and matches a full rebuild.

utkuozdemir:
> The production approach is sound, but the branch is not merge-ready yet. Its divergent-branch test assumes git init creates main; with the standard master default, git checkout main fails unchecked and the test fails (36 passed, 1 failed). It passes only when Git is explicitly initialized on main (37 passed). Please use git init -b main or discover the branch, assert subprocess return codes, then rebase after #729 and add one end-to-end case where the stored SHA spans multiple commits including a rename and matches a full rebuild.

Good point, I changed it exactly the way you suggested and amended the commit. Waiting for https://github.com/tirth8205/code-review-graph/pull/729 to land, then I'm going to add the e2e case you asked.

tirth8205:
Rechecked against `main` at `9fed471`: 176 tests pass with 1 expected skip. The requested multi-commit audit also passes: starting from a stored SHA three commits behind, with a rename, importer change, and new file across those commits, the default update catches every file, removes the old path, and matches a fresh full rebuild exactly for nodes and edges.

No production-code blocker remains. The branch still conflicts with current main and does not commit that multi-commit regression. Please rebase, retain current rename/lifecycle coverage, and add the end-to-end regression before merge.

tirth8205:
The corrected integration is now in #769. The contributor patch and Utku Özdemir’s attribution are preserved unchanged.

The added regression splits a module rename, importer edit, and new importing file across three commits. It proves the default stored-SHA update sees all four old/new paths, purges the old graph state, and matches a fresh rebuild’s complete node and edge snapshots. All 244 affected tests passed after rebasing on the merged forget/unbind work; full hosted CI is running now.

I am keeping this original PR open until the corrected integration passes the complete Python, Windows, type, lint, schema, security, and CodeQL matrix.

tirth8205:
Thank you, @utkuozdemir. This contribution is now integrated and merged through #770 at `329dbb91`.

Your original patch and author attribution are preserved in the merged history. During validation we added an end-to-end three-commit regression covering a rename, an importer update, and a new file, comparing incremental output with a clean full build.

The corrected standalone integration #769 passed its full hosted matrix. The exact combined #770 state then passed Python 3.10–3.13, Windows, lint, type-check, security, schema sync, review, CodeQL, and GitGuardian before merge.

PR #701, opened 2026-07-21
fix: honor --repo pointing at a module's own .code-review-graph dir

### LINKED ISSUE #697: search/query/impact ignore explicit --repo inside a monorepo (git-root walk overrides it)
# `search`/`query`/`impact` ignore explicit `--repo` inside a monorepo (git-root walk overrides it)

**Version:** code-review-graph 2.3.7 (PyPI)
**OS:** Windows 11 (amd64), Python 3.14.4
**Repo layout tested:** `llvm-project` (single top-level `.git`, 21 subdirectories used as independently `build`/`register`ed CRG projects, e.g. `llvm/`, `clang/`)

## Summary

`build`, `register`, `daemon add`, and `status` all correctly treat an explicit path as the
project root — even when that path is a *subdirectory* of a larger git checkout, as long as it
has its own `.code-review-graph` marker. `search`, `query`, and `impact` do **not**: they resolve
`--repo` through `find_project_root()` → `find_repo_root()`, which walks *up* the directory tree
looking only for `.git`/`.svn` and ignores `.code-review-graph`. In a monorepo (one `.git` at the
top, modules as subdirectories — a very common layout, e.g. `llvm-project`), this always resolves
to the monorepo root instead of the module you asked for, and the command fails with:

```
No graph found at <monorepo_root>\.code-review-graph\graph.db. Run `code-review-graph build` first.
```

even though the graph exists exactly where `--repo` pointed.

## Repro

```powershell
# llvm-project has a single .git at its root; llvm/ and clang/ are subdirectories.
code-review-graph build --repo "D:\llvm-project\llvm"      # succeeds, creates
                                                             #   D:\llvm-project\llvm\.code-review-graph\graph.db
code-review-graph status --repo "D:\llvm-project\llvm"     # OK — respects the path, shows real stats
code-review-graph search "raw_ostream" --repo "D:\llvm-project\llvm"
# -> No graph found at D:\llvm-project\.code-review-graph\graph.db. Run `code-review-graph build` first.
#                      ^^^^^^^^^^^^^^^^ resolved to the monorepo root, not the --repo path
```

`query` and `impact` fail the same way (they're all wired through the same
`_GRAPH_TOOL_COMMANDS` branch in `cli.py`).

## Root cause

`cli.py`, in the `_GRAPH_TOOL_COMMANDS` dispatch:

```python
requested_root = Path(args.repo).expanduser() if args.repo else None
repo_root = find_project_root(requested_root)
```

`find_project_root()` (in `incremental.py`) passes `requested_root` into `find_repo_root()` as the
*starting point for an upward walk*, not as an already-resolved root:

```python
def find_repo_root(start=None, stop_at=None):
    current = start or Path.cwd()
    while current != current.parent:
        if (current / ".git").exists():
            return current
        ...
        current = current.parent
    ...
```

So an explicit `--repo` is only ever used as a *starting point* to search upward for `.git`/`.svn`
— never validated/accepted on its own, and never checked for `.code-review-graph`. This is
inconsistent with `tools/_common.py::_validate_repo_root()`, which `build`/`register`/`status` use
and which correctly accepts `.git` **or** `.svn` **or** `.code-review-graph` at the exact given
path.

## Suggested fix

Make the `_GRAPH_TOOL_COMMANDS` branch in `cli.py` use `_validate_repo_root()` when `--repo` is
explicitly given, matching `build`/`register`/`status`, and only fall back to the upward
`find_project_root()` walk when no `--repo` was passed (cwd auto-detect):

```python
if args.repo:
    from .tools._common import _validate_repo_root
    repo_root = _validate_repo_root(Path(args.repo).expanduser())
else:
    repo_root = find_project_root(None)
```

I've verified this fix locally against a 21-module build of `llvm-project` (largest module,
`llvm/`, ~80K files / 130K nodes / 1.5M edges) — `search`, `query callers_of`, and
`impact --files` all return correct results with `--repo` pointing at a module subdirectory, with
no behavior change for the no-`--repo` cwd-based auto-detect path. Happy to open a PR with this
change plus a regression test (build two nested fake repos under one `.git`, assert `search --repo
<child>` doesn't resolve to `<parent>`) if useful.

## Workaround (for anyone hitting this before a fix ships)

Set `CRG_REPO_ROOT` (highest-precedence override in `find_project_root`) instead of `--repo`:

```powershell
$env:CRG_REPO_ROOT = "D:\llvm-project\llvm"
code-review-graph search "raw_ostream"
```

Khusan-Khujakeldiev:
Fixes #697.

## Problem

`search`/`query`/`impact` (and the rest of `_GRAPH_TOOL_COMMANDS`) resolve an
explicit `--repo` via `find_project_root()`, which only walks *up* looking for
`.git`/`.svn`. In a monorepo with a single `.git` at the root and independently
`build`/`register`-ed module subdirectories (each carrying its own
`.code-review-graph` marker but no `.git` of their own — e.g. `llvm-project`'s
`llvm/` and `clang/`), this silently climbs past the module's own graph to the
monorepo root and reports "No graph found" even though the graph exists
exactly where `--repo` pointed.

## Fix

When `--repo` points directly at a directory that already has its own
`.code-review-graph` subdirectory, honor that path as-is instead of using it
only as a starting point for the upward git-root walk. This mirrors how
`build`/`register`/`status` already treat `.code-review-graph` as sufficient
evidence of a project root via `tools._common._validate_repo_root`.

When `--repo` has no `.code-review-graph` of its own, behavior is unchanged
(falls through to the existing `find_project_root` walk), so the existing
"nested subdirectory of a single repo" case is unaffected.

## Testing

- Added `test_tool_command_honors_repo_flag_over_ancestor_git_in_monorepo`,
  which reproduces the monorepo shape (one `.git` at the top, a module
  subdirectory with its own `.code-review-graph/graph.db`) and asserts
  `search --repo <module>` resolves to `<module>`, not the ancestor with
  `.git`. Verified it fails on `main` and passes with this fix.
- Ran the full test suite locally on Windows before and after the change;
  no new failures (only pre-existing, unrelated Windows-only failures around
  `chmod`/`symlink` privileges and git submodule fixtures remain, present on
  `main` too).
- Verified against the real-world case that surfaced this: a 21-module,
  ~180K-file `llvm-project` checkout built/registered module-by-module with
  CRG, where `search`/`query callers_of`/`impact --files` now return correct
  results with `--repo` pointing at a module (e.g. the largest module,
  `llvm/`, ~80K files).

### DIFF
--- a/code_review_graph/cli.py
+++ b/code_review_graph/cli.py
@@ -1162,7 +1162,22 @@ def main() -> None:
         from .incremental import find_project_root, get_db_path
 
         requested_root = Path(args.repo).expanduser() if args.repo else None
-        repo_root = find_project_root(requested_root)
+        if requested_root is not None and (requested_root / ".code-review-graph").is_dir():
+            # An explicit --repo that already has its own .code-review-graph
+            # directory must be honored as-is. This matters for monorepos
+            # with a single .git at the top and independently
+            # build/register-ed modules as subdirectories (each carrying
+            # its own .code-review-graph marker but no .git of their own):
+            # find_project_root() below only walks *up* looking for
+            # .git/.svn, so it would silently climb past the module's own
+            # marker to the monorepo root and miss the graph that was
+            # built right here. build/register/status avoid this via
+            # tools._common._validate_repo_root; do the equivalent here,
+            # scoped to the case where evidence of a real build already
+            # exists at the requested path.
+            repo_root = requested_root.resolve()
+        else:
+            repo_root = find_project_root(requested_root)
         db_path = get_db_path(repo_root)
         if not db_path.exists():
             print(
--- a/tests/test_cli_tool_commands.py
+++ b/tests/test_cli_tool_commands.py
@@ -144,3 +144,38 @@ def test_tool_command_missing_graph_exits_nonzero(tmp_path, monkeypatch, capsys)
 
     assert exc_info.value.code == 1
     assert "No graph found" in capsys.readouterr().err
+
+
+def test_tool_command_honors_repo_flag_over_ancestor_git_in_monorepo(
+    tmp_path, monkeypatch, capsys,
+):
+    """Regression test: a single ``.git`` at a monorepo root must not shadow
+    a module subdirectory's own ``.code-review-graph`` graph when ``--repo``
+    points directly at that module (e.g. ``llvm-project`` with ``llvm/`` and
+    ``clang/`` independently built/registered as CRG projects)."""
+    monorepo = tmp_path / "monorepo"
+    module = monorepo / "llvm"
+    module.mkdir(parents=True)
+    (monorepo / ".git").mkdir()
+
+    # A graph exists for the module itself...
+    module_data = module / ".code-review-graph"
+    module_data.mkdir()
+    (module_data / "graph.db").touch()
+
+    # ...but NOT at the monorepo root, so the buggy resolution (walking up
+    # to the nearest .git) would report "no graph found" there instead.
+    monorepo_data = monorepo / ".code-review-graph"
+    monorepo_data.mkdir()
+
+    argv = ["code-review-graph", "search", "raw_ostream", "--repo", str(module)]
+    result = {"status": "ok", "tool": "semantic_search_nodes"}
+
+    with patch.object(sys, "argv", argv):
+        with patch(
+            "code_review_graph.tools.semantic_search_nodes", return_value=result,
+        ) as tool:
+            cli.main()
+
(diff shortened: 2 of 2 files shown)


tirth8205:
Issue #697 is now fixed and shipped through #730 at `3a27c65`. The selected implementation covers both an explicit module root and a nested path inside that registered subproject, rejects paths with no project marker cleanly, and includes a real graph-build plus CLI-search regression. It passed 16 focused local tests and all 11 hosted checks, including Windows. Your diagnosis of the ancestor `.git` override was correct; I am closing this overlapping implementation now that the broader validated fix is merged.

PR #710, opened 2026-07-22
fix(eval): build the vector index so semantic benchmarks can report a number; stop tests writing to $HOME

### LINKED ISSUE #711: [Bug]: agent_baseline (and search_quality, multi_hop_retrieval) always report no_graph_results — eval never builds the vector index
### code-review-graph version

2.3.7

### Operating system

Linux

### Python version

3.12.3

### AI platform

claude-code

```shell
Nodes: 6287
Edges: 32036
Files: 1122
Languages: python
Last updated: 2026-07-22
Built at commit: 22381558446c
```

(Graph for the `fastapi` eval checkout at the pinned SHA `22381558446c5d1ac376680a6581dd63b3a04119`, built by the eval runner itself.)

### Steps to reproduce

```
1. code-review-graph eval --benchmark agent_baseline --repo fastapi
2. cat evaluate/results/fastapi_agent_baseline_*.csv
```

**Expected:** a `baseline_to_graph_ratio` per question, as described in README.md ("The `agent_baseline` eval benchmark measures that realistic baseline — a pure-python grep over the corpus, top-3 files by match count, token-counted and compared to the graph query cost").

**Actual:** every row comes back `status="no_graph_results"` with `graph_tokens=0`, and `aggregate()` excludes those rows — so the run reports `ok_rows: 0` / `median: None` rather than an error.

```csv
repo,question,...,baseline_tokens,graph_tokens,baseline_to_graph_ratio,status
fastapi,How does include_router register routes on the application,...,132479,0,,no_graph_results
fastapi,Where does APIRoute build its route handler,...,192817,0,,no_graph_results
fastapi,How does solve_dependencies resolve Depends parameters,...,150384,0,,no_graph_results
```

### Additional context

**Root cause.** `eval/runner.py` builds the graph and calls `run_post_processing(store)`, but that step's embedding refresh is a *refresh*, not a bootstrap: `embeddings.refresh_embeddings()` returns early on a graph with no existing vectors — deliberately, per its docstring, so that no build path can silently load a model or incur API cost. Nothing else in the eval framework populates the index, so the `embeddings` table stays at 0 rows.

`search.hybrid_search` therefore falls through to FTS5 alone, which scores a full sentence against no matching document and returns nothing. This affects every benchmark that puts a natural-language question through it: `agent_baseline`, `search_quality`, `multi_hop_retrieval`.

The graph itself is fine — 6287 nodes, FTS5 populated with 6287 rows. Bare identifiers match; sentences do not:

```
'How does include_router register routes on the application' -> 0 hits
'include_router'                                             -> 5 hits
'APIRoute'                                                   -> 5 hits
'solve_dependencies'                                         -> 1 hit
```

After building the index manually with `code-review-graph embed`, the same three questions produce usable rows: **28.1x / 82.4x / 68.0x**.

This may explain why no `agent_baseline` CSVs exist under `evaluate/results/` even though README.md cites the path `evaluate/results/<repo>_agent_baseline_*.csv` as the realistic-baseline measurement.

A fix is proposed in #710 (opt-in `--embed` flag, plus a warning when a semantic benchmark runs against an unindexed graph, plus excluded-row counts in `aggregate()` so the failure is visible rather than arriving as an empty median).

### LINKED ISSUE #712: [Bug]: running the test suite writes pytest temp paths into the developer's real ~/.code-review-graph/registry.json
### code-review-graph version

2.3.7

### Operating system

Linux

### Python version

3.12.3

### AI platform

claude-code

```shell
Nodes: 4418
Edges: 34123
Files: 218
Last updated: 2026-07-22T10:08:04
Built on branch: main
Built at commit: 6ce25b4e53f9
```

(Graph of code-review-graph itself, at `6ce25b4`. The bug is in the test suite, so the graph is incidental.)

### Steps to reproduce

```
1. rm -rf ~/.code-review-graph      # start from a clean home directory
2. uv sync --extra dev
3. uv run pytest -q
4. cat ~/.code-review-graph/registry.json
```

**Expected:** running the test suite leaves the developer's home directory untouched.

**Actual:** `~/.code-review-graph/` is created and `registry.json` is written with pytest temp paths:

```json
{
  "repos": [
    {
      "path": "/tmp/pytest-of-<user>/pytest-0/test_registry_data_dir_overrid0/project",
      "data_dir": "/tmp/pytest-of-<user>/pytest-0/test_registry_data_dir_overrid0/external"
    },
    {
      "path": "/tmp/pytest-of-<user>/pytest-0/test_registry_data_dir_overrid1/project",
      "data_dir": "/tmp/pytest-of-<user>/pytest-0/test_registry_data_dir_overrid1/registry-data"
    },
    {
      "path": "/tmp/pytest-of-<user>/pytest-0/test_data_dir_auto_creates_dir0/project",
      "data_dir": "/tmp/pytest-of-<user>/pytest-0/test_data_dir_auto_creates_dir0/nonexistent/nested/path"
    }
  ]
}
```

### Additional context

**Two independent routes reach the real home directory.**

1. `Registry()` defaults to `~/.code-review-graph/registry.json` (`registry.py:19-20`), and `incremental.get_data_dir()` constructs one internally (`incremental.py:316`). The three tests in `TestDataDirRegistry` (`tests/test_incremental.py:419,439,489`) call `Registry()` with no path argument, so they both **read and write** the registry of whoever runs the suite.

   The write is the visible symptom; the read is the more interesting half. It makes those tests depend on machine state — a developer with a real registered repo can get a different result from one without, and the failure would look unrelated to the registry.

2. `daemon.py:41-43` builds `CONFIG_PATH`, `PID_PATH` and `STATE_PATH` from `Path.home()`, and `DaemonConfig.log_dir` defaults the same way (`daemon.py:69`). Bisecting by test file, `tests/test_daemon.py` is what recreates `~/.code-review-graph/` even once the registry half is fixed.

**Common root:** all of these are import-time constants. A constant evaluated at import cannot be redirected afterwards — by the time a fixture could set anything, the value is already frozen. That is why an env-var override has to be resolved per call rather than captured once.

Verified by deleting `~/.code-review-graph` and re-running the full suite: it is recreated every time on `main`.

A fix is proposed in #710 (a `constants.crg_home()` helper reading `$CRG_HOME` — same convention as the existing `CRG_DATA_DIR` — with `registry.py` and `daemon.py` routed through it, the legacy `daemon` constant names preserved via a PEP 562 module `__getattr__`, and an autouse fixture in a new `tests/conftest.py`).

nadiadatepe-eng:
# Pull Request

## Linked issue

Closes #711 — `agent_baseline` / `search_quality` / `multi_hop_retrieval` always report `no_graph_results`, because the eval framework never builds the vector index.

Closes #712 — running the test suite writes pytest temp paths into the developer's real `~/.code-review-graph/registry.json`.

Both were found while reproducing the README benchmarks, not from a user report.

## What & why

Two independent fixes, one commit each.

### 1. The eval framework never built the vector index

`run_eval` builds the graph and calls `run_post_processing(store)`, but that step's embedding refresh is a *refresh*, not a bootstrap: `refresh_embeddings()` returns early on a graph with no existing vectors — deliberately, so no build path can silently load a model or incur API cost. Nothing else in the eval framework populated the index.

So every benchmark that puts a natural-language question through `hybrid_search` — `agent_baseline`, `search_quality`, `multi_hop_retrieval` — fell through to FTS5, which scores a full sentence against no matching document and returns nothing. Every row came back `status="no_graph_results"`, and `aggregate()` excludes those rows, so the run reported `ok_rows: 0` / `median: None` rather than an error.

Reproduced on the shipped `fastapi` config. The graph itself was healthy — 6287 nodes, FTS5 populated with 6287 rows — but all three `agent_questions` returned 0 hits:

```
'How does include_router register routes on the application' -> 0 hits
'Where does APIRoute build its route handler'                -> 0 hits
'How does solve_dependencies resolve Depends parameters'     -> 0 hits
```

Bare identifiers worked fine (`'APIRoute'` → 5 hits), which is what makes the failure quiet: FTS5 is doing its job, it just can't match a sentence.

After building the index, the same three questions produce usable rows:

| question | baseline_tokens | graph_tokens | ratio |
|---|---:|---:|---:|
| include_router register routes | 132479 | 4711 | 28.1x |
| APIRoute build route handler | 192817 | 2339 | 82.4x |
| solve_dependencies / Depends | 150384 | 2213 | 68.0x |

This is likely why no `agent_baseline` CSVs exist under `evaluate/results/` even though README.md cites the path `evaluate/results/<repo>_agent_baseline_*.csv` as the realistic-baseline measurement.

Changes:

- `run_eval(embed=..., embedding_provider=..., embedding_model=...)`, plus `_build_embedding_index()`, which reads the graph through the runner's already-open `GraphStore`. **Default off**, so the cost invariant `refresh_embeddings` protects is unchanged.
- `_warn_if_semantic_index_missing()` runs before the benchmarks and names the affected ones, so the failure announces itself instead of arriving as an empty aggregate.
- `agent_baseline.aggregate()` reports `no_graph_results_rows` / `no_baseline_match_rows`. Excluded rows are counted, not just dropped — "no result" and "every query failed" had the same signature before.
- CLI: `eval --embed` / `--embed-provider` / `--embed-model`.

`run_eval`'s new parameters are optional and trailing, so the signature stays source-compatible.

### 2. The test suite wrote into the developer's home directory

Running `pytest` left entries like this in the real `~/.code-review-graph/registry.json`:

```json
{"path": "/tmp/pytest-of-<user>/pytest-0/test_registry_data_dir_overrid0/project",
 "data_dir": "/tmp/pytest-of-<user>/pytest-0/.../external"}
```

Two routes reached it:

- `Registry()` defaults to `~/.code-review-graph/registry.json`, and `incremental.get_data_dir()` constructs one internally (`incremental.py:316`). Tests covering data-dir resolution therefore both *read and wrote* the registry of whoever ran the suite. Besides the pollution, that made those tests machine-dependent: a developer with a registered repo could get a different result from one without.
- `daemon.py` built `CONFIG_PATH`, `PID_PATH`, `STATE_PATH` and `DaemonConfig.log_dir` from `Path.home()` as import-time constants.

The common root is that an import-time constant cannot be redirected afterwards — by the time a fixture could set anything, the value is frozen. So the paths resolve per call:

- `constants.crg_home()` reads `$CRG_HOME`, falling back to `~/.code-review-graph`. Same convention as the existing `CRG_DATA_DIR`.
- `registry.default_registry_path()` and four `daemon.default_*()` helpers route through it.
- `daemon.CONFIG_PATH` / `PID_PATH` / `STATE_PATH` keep working via a PEP 562 module `__getattr__`, with `__dir__` so they stay visible to introspection. Not covered: `import *` — the module defines no `__all__`, and adding one would change what the star exports for every other name.
- `tests/conftest.py` points `$CRG_HOME` at a tmp directory for every test. Autouse and unconditional: an opt-in fixture stops protecting a test the day someone forgets to request it.

Scoped deliberately — the editor-integration installers in `skills.py` write to `~/.codex`, `~/.cursor` and `~/.config/opencode`, which are outside CRG state and whose tests already patch `Path.home()` themselves.

## How it was tested

```bash
# 1980 passed, 5 skipped, 2 xpassed in 30.70s

# All checks passed!

# Success: no issues found in 66 source files
```

End-to-end, from a cold start with the graph database deleted:

```bash
# without --embed: the failure now announces itself
$ code-review-graph eval --benchmark agent_baseline --repo fastapi
  no vector index — agent_baseline will score natural-language questions
  against FTS5 alone and return zero hits. Re-run with --embed.

# with --embed: 3 of 3 rows usable
$ code-review-graph eval --benchmark agent_baseline --repo fastapi --embed
  28.1x / 82.4x / 68.0x
```

The home-directory fix was verified by deleting `~/.code-review-graph` and running the full suite: it is no longer recreated. `test_daemon.py` was the file recreating it (bisected).

Each new test was checked against a reverted source tree to confirm it actually fails without the fix, rather than passing vacuously.

## Note on reproducibility

`docs/REPRODUCING.md` states two runs on different machines produce identical numbers. That holds for the parsing/FTS/graph benchmarks, but the embedding-backed ratios drift slightly between runs on the same machine (82.3 → 82.4, 67.9 → 68.0). Not addressed here — flagging it in case you want the claim narrowed to the deterministic benchmarks.

## Checklist

### DIFF
--- a/code_review_graph/cli.py
+++ b/code_review_graph/cli.py
@@ -840,6 +840,28 @@ def main() -> None:
     eval_cmd.add_argument("--all", action="store_true", dest="run_all", help="Run all benchmarks")
     eval_cmd.add_argument("--report", action="store_true", help="Generate report from results")
     eval_cmd.add_argument("--output-dir", default=None, help="Output directory for results")
+    eval_cmd.add_argument(
+        "--embed",
+        action="store_true",
+        help=(
+            "Build the vector index after each graph build. Required by the "
+            "agent_baseline, search_quality and multi_hop_retrieval "
+            "benchmarks: without it their natural-language questions hit "
+            "FTS5 only and return zero results (default: disabled)"
+        ),
+    )
+    eval_cmd.add_argument(
+        "--embed-provider",
+        choices=["local", "openai", "google", "minimax"],
+        default=None,
+        help="Provider for --embed (default: local, needs "
+             "code-review-graph[embeddings])",
+    )
+    eval_cmd.add_argument(
+        "--embed-model",
+        default=None,
+        help="Model for --embed (default: the provider's own default)",
+    )
 
     # detect-changes
     detect_cmd = sub.add_parser(
@@ -1262,6 +1284,9 @@ def main() -> None:
                 repos=repos,
                 benchmarks=benchmarks,
                 output_dir=getattr(args, "output_dir", None),
+                embed=getattr(args, "embed", False),
+                embedding_provider=getattr(args, "embed_provider", None),
+                embedding_model=getattr(args, "embed_model", None),
             )
             print(f"\nCompleted {len(results)} benchmark(s).")
             print("Run 'code-review-graph eval --report' to generate tables.")
--- a/code_review_graph/constants.py
+++ b/code_review_graph/constants.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 
 import math
 import os
+from pathlib import Path
 
 
 def _bounded_float_env(
@@ -71,3 +72,28 @@ IMPACT_DEPTH_DECAY = _bounded_float_env(
 IMPACT_SCORE_FLOOR = _bounded_float_env(
     "CRG_IMPACT_SCORE_FLOOR", 0.05, lower=0.0, upper=1.0,
 )
+
+
+#: Overrides the per-user state directory that holds ``registry.json``,
+#: ``watch.toml``, ``daemon.pid``, ``daemon-state.json`` and ``logs/``.
+#: Follows the same convention as CRG_DATA_DIR.
+CRG_HOME_ENV = "CRG_HOME"
+
+_DEFAULT_CRG_HOME = Path.home() / ".code-review-graph"
+
+
+def crg_home() -> Path:
+    """Return the per-user state directory for code-review-graph.
+
+    ``$CRG_HOME`` wins when set and non-empty; otherwise
+    ``~/.code-review-graph``.
+
+    Resolved per call rather than captured in a module-level constant. An
+    import-time constant cannot be redirected afterwards, which is what let
+    the test suite write into the real home directory of whoever ran it: by
+    the time a fixture set the variable, the value had already been frozen.
+    """
+    override = os.environ.get(CRG_HOME_ENV, "").strip()
+    if override:
+        return Path(override).expanduser()
+    return _DEFAULT_CRG_HOME
--- a/code_review_graph/daemon.py
+++ b/code_review_graph/daemon.py
@@ -32,15 +32,71 @@ else:
     except ImportError:
         tomllib = None  # type: ignore[assignment]
 
+from .constants import crg_home
+
 logger = logging.getLogger(__name__)
 
 # ---------------------------------------------------------------------------
 # Config file location
 # ---------------------------------------------------------------------------
 
-CONFIG_PATH: Path = Path.home() / ".code-review-graph" / "watch.toml"
-PID_PATH: Path = Path.home() / ".code-review-graph" / "daemon.pid"
-STATE_PATH: Path = Path.home() / ".code-review-graph" / "daemon-state.json"
+def default_config_path() -> Path:
+    """Path to ``watch.toml`` under the per-user state directory."""
+    return crg_home() / "watch.toml"
+
+
+def default_pid_path() -> Path:
+    """Path to the daemon PID file."""
+    return crg_home() / "daemon.pid"
+
+
+def default_state_path() -> Path:
+    """Path to the persisted daemon state."""
+    return crg_home() / "daemon-state.json"
+
+
+def default_log_dir() -> Path:
+    """Directory for per-repo daemon logs."""
+    return crg_home() / "logs"
+
+
+# These four were module-level constants built from Path.home(). They resolve
+# per call now so $CRG_HOME can redirect them: an import-time constant is
+# frozen before any caller — a test fixture, a sandboxed run — gets the chance
+# to set the variable, which is how the test suite ended up writing into the
+# real home directory of whoever ran it.
+#
+# The PEP 562 shim below keeps the old attribute names working for anything
+# that already imported them: both ``daemon.CONFIG_PATH`` and
+# ``from …daemon import CONFIG_PATH`` route through ``__getattr__``, and
+# ``__dir__`` keeps them visible to introspection. Not covered: ``import *``
+# (this module defines no ``__all__``, and adding one would change what the
+# star exports for every other name) and static analysers, which cannot see
+# dynamic attributes. Both are acceptable — these were never public API, and
+# the alternative is deleting the names outright.
+_LAZY_PATHS = {
+    "CONFIG_PATH": default_config_path,
+    "PID_PATH": default_pid_path,
+    "STATE_PATH": default_state_path,
+}
+
+
+def __getattr__(name: str) -> Path:
+    if name in _LAZY_PATHS:
+        return _LAZY_PATHS[name]()
+    raise AttributeError(f"module {__name__!r} has no attribute {name!r}")
+
+
+def __dir__() -> list[str]:
+    """Include the lazy names so ``dir()`` and tab-completion still find them.
+
+    ``__getattr__`` alone covers attribute access and ``from … import X``,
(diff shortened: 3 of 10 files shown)


jordandunmire97-ai (approved):

jordandunmire97-ai (commented):
Idealization of what is prospering

tirth8205:
The vector bootstrap and test-home fixes are valid, but end-to-end validation found two remaining blockers: custom provider/model indexes were built under one identity while all three benchmarks queried the default identity (`no_graph_results`), and embedding bootstrap failures leaked the open `GraphStore`. I preserved both contributor-authored commits, resolved the registry conflict without dropping #682’s Unicode coverage, added provider/model propagation plus failure cleanup tests, and opened the corrected integration as #757. Exact-head validation: 203 focused tests plus real multiprocessing parity passed, Ruff/diff passed, and a 243-file graph build completed with zero errors. I will keep this PR open until #757 passes hosted checks and is merged.

tirth8205:
The corrected integration #757 is now merged at `03daa269` with both of your original commits and authorship preserved. The final version also propagates the exact embedding provider/model through all three benchmarks and closes `GraphStore` on bootstrap failure. It passed 203 focused tests, real multiprocessing parity, a full graph build, CodeQL, and the complete hosted matrix including Windows. Closing this conflicted PR as fully incorporated.

PR #715, opened 2026-07-22
fix(graph): preserve same-file nodes on qualified_name collision (#585)

### LINKED ISSUE #585: Silent node loss: same-file qualified_name collisions are dropped on insert

SarveshaKumarKS:
# Pull Request

## Linked issue

Closes #585

## What & why

Two distinct definitions in the same file could produce the same base qualified name. Because `nodes.qualified_name` is unique and `upsert_node()` updated on that conflict, the later definition silently replaced the earlier node. Its line range disappeared, and distinct graph edges could collapse onto the surviving node.

Minimal reproduction:

```python
nodes = [
    NodeInfo(kind="Function", name="render", file_path="/repo/app.py", line_start=10, line_end=12),
    NodeInfo(kind="Function", name="render", file_path="/repo/app.py", line_start=20, line_end=24),
]
store.store_file_nodes_edges("/repo/app.py", nodes, [])
```

Before:

```text
[("/repo/app.py::render", 20, 24)]
```

After:

```text
[("/repo/app.py::render", 10, 12),
 ("/repo/app.py::render:S<semantic-fingerprint>", 20, 24)]
```

## Identity strategy

The first definition keeps the existing bare qualified name. Only genuine collisions are disambiguated, so all unambiguous identities remain unchanged.

A collided definition receives a deterministic suffix derived from semantic node fields rather than its source line. If multiple colliding definitions have identical semantic fields, a deterministic occurrence suffix is added. This preserves both definitions while keeping their identities stable when harmless lines are inserted above them during an incremental rebuild.

That stability matters because qualified names are also edge keys. A line-based suffix changed after a line shift, leaving cross-file incoming edges pointed at a deleted identity. The semantic suffix keeps those identities—and therefore their incoming edges—valid across the real `store_file_nodes_edges()` remove-and-replace lifecycle.

## Edge-resolution trade-off

Batch storage aligns an edge endpoint with a collided identity only when the edge provides enough evidence:

- outgoing edges use the source definition whose line range contains the edge;
- `CONTAINS` edges use the target definition's start line; and
- recursive calls use the same disambiguated identity as their source definition.

An incoming call with no evidence about which duplicate it targets deliberately remains on the canonical bare identity. It is not assigned arbitrarily to a disambiguated definition.

## How it was tested

The regression coverage now verifies:

- two same-file definitions with the same base qualified name both survive with distinct line ranges;
- direct re-indexing of the same definition updates in place without creating a duplicate;
- a harmless line insertion above duplicate definitions does not change their identities;
- distinct cross-file callers to every duplicate still resolve after re-indexing;
- recursive caller/callee queries resolve a disambiguated node; and
- ambiguous incoming targets remain explicitly attached to the canonical identity.

Validation on current `main` (`805c4c6`):

```bash
uv run pytest
```

Results: 2,070 passed, 5 skipped, 2 xpassed; Ruff clean; mypy clean across 66 source files.

## Checklist

- [x] All tests pass: `uv run pytest`
- [x] Implementation docstrings describe the behavior

### DIFF
--- a/code_review_graph/graph.py
+++ b/code_review_graph/graph.py
@@ -11,6 +11,7 @@ import hashlib
 import json
 import logging
 import os
+import re
 import sqlite3
 import threading
 import time
@@ -45,6 +46,9 @@ logger = logging.getLogger(__name__)
 # ``EXTENSION_TO_LANGUAGE``; TSX keeps its own grammar name.
 _JAVASCRIPT_LANGUAGE_FAMILY = ("javascript", "typescript", "tsx")
 _JAVASCRIPT_LANGUAGE_FAMILY_SET = frozenset(_JAVASCRIPT_LANGUAGE_FAMILY)
+_COLLISION_SUFFIX_RE = re.compile(
+    r"^[0-9a-f]{12}(?::(?:[2-9]|[1-9][0-9]+))?$",
+)
 
 
 def _compatible_edge_languages(language: str) -> tuple[str, ...]:
@@ -290,7 +294,7 @@ class GraphStore:
         candidates = [
             row for row in rows
             if row["qualified_name"] == base
-            or row["qualified_name"].startswith(f"{base}:S")
+            or self._is_collision_identity(base, row["qualified_name"])
         ]
         for row in candidates:
             if row["line_start"] == node.line_start:
@@ -346,6 +350,18 @@ class GraphStore:
             bool(row["is_test"]),
         )
 
+    @staticmethod
+    def _is_collision_identity(base: str, qualified_name: str) -> bool:
+        """Return whether *qualified_name* has this store's exact suffix shape.
+
+        Checking the full generated suffix avoids treating colon-bearing symbol
+        names, such as Objective-C selectors, as collision identities.
+        """
+        prefix = f"{base}:S"
+        if not qualified_name.startswith(prefix):
+            return False
+        return _COLLISION_SUFFIX_RE.fullmatch(qualified_name[len(prefix):]) is not None
+
     def _plan_file_identities(
         self, file_path: str, nodes: list[NodeInfo],
     ) -> list[str]:
@@ -370,7 +386,7 @@ class GraphStore:
                 row
                 for row in old_rows
                 if row["qualified_name"] == base
-                or row["qualified_name"].startswith(f"{base}:S")
+                or self._is_collision_identity(base, row["qualified_name"])
             ]
 
         planned: list[str | None] = [None] * len(nodes)
@@ -378,6 +394,14 @@ class GraphStore:
         assigned: set[str] = set()
         for base, indices in new_by_base.items():
             candidates = old_by_base[base]
+            used_old_ids: set[int] = set()
+            if len(indices) == 1 and len(candidates) == 1:
+                index = indices[0]
+                row = candidates[0]
+                planned[index] = row["qualified_name"]
+                assigned.add(row["qualified_name"])
+                used_old_ids.add(row["id"])
+
             pairs: list[tuple[int, int, int, sqlite3.Row]] = []
             for index in indices:
                 signature = self._semantic_signature(nodes[index])
@@ -387,7 +411,6 @@ class GraphStore:
                     distance = abs(nodes[index].line_start - row["line_start"])
                     pairs.append((distance, index, row["id"], row))
 
-            used_old_ids: set[int] = set()
             for _, index, row_id, row in sorted(pairs):
                 if planned[index] is not None or row_id in used_old_ids:
                     continue
@@ -468,14 +491,24 @@ class GraphStore:
         """Resolve a bare edge endpoint to one lone historical suffix."""
         if self.get_node(qualified_name) is not None:
             return qualified_name
-        prefix = f"{qualified_name}:S"
+        lower = f"{qualified_name}:S"
+        upper = f"{qualified_name}:T"
         rows = self._conn.execute(
             "SELECT qualified_name FROM nodes "
-            "WHERE substr(qualified_name, 1, ?) = ? LIMIT 2",
-            (len(prefix), prefix),
-        ).fetchall()
-        if len(rows) == 1:
-            return rows[0]["qualified_name"]
+            "WHERE qualified_name >= ? AND qualified_name < ? "
+            "ORDER BY qualified_name",
+            (lower, upper),
+        )
+        matches = []
+        for row in rows:
+            candidate = row["qualified_name"]
+            if not self._is_collision_identity(qualified_name, candidate):
+                continue
+            matches.append(candidate)
+            if len(matches) == 2:
+                break
+        if len(matches) == 1:
+            return matches[0]
         return qualified_name
 
     def remove_file_data(self, file_path: str) -> None:
--- a/code_review_graph/graph.py
+++ b/code_review_graph/graph.py
@@ -230,8 +230,13 @@ class GraphStore:
 
     def upsert_node(self, node: NodeInfo, file_hash: str = "") -> int:
         """Insert or update a node. Returns the node ID."""
+        node_id, _ = self._upsert_node(node, file_hash)
+        return node_id
+
+    def _upsert_node(self, node: NodeInfo, file_hash: str = "") -> tuple[int, str]:
+        """Insert or update a node and return its ID and stored identity."""
         now = time.time()
-        qualified = self._make_qualified(node)
+        qualified = self._qualified_for_upsert(node)
         extra = json.dumps(node.extra) if node.extra else "{}"
 
         self._conn.execute(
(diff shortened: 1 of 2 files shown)


tirth8205:
I independently applied this head to current main and ran the submitted collision tests; all 3 pass. It is not mergeable yet because the missing incremental lifecycle case loses graph meaning. Exact failing regression: store two same-file `render` definitions, create an incoming CALLS edge to the second identity (`target.py::render:L20`), insert one harmless line above both definitions, and re-index the target file. The second identity becomes `:L21`; the `:L20` node is gone, while the cross-file incoming edge still targets `:L20`, so it is dangling. Root cause: the identity encodes `line_start`, and `store_file_nodes_edges()` removes the file rows before identities are recomputed. The direct `upsert_node()` test does not exercise that real replace/re-index path. Please replace the line-based identity with a semantic stable identity, or implement explicit old-to-new identity migration for incoming edges, and add regressions for harmless line shifts, callers to every duplicate, and explicit ambiguous-target behavior. I am keeping both the PR and #585 open until those cases pass.

SarveshaKumarKS:
Thanks for the detailed regression. I reproduced the dangling-edge failure and pushed a revised implementation.
The line-based :L<line_start> identity has been replaced with a deterministic semantic fingerprint, so harmless line shifts no longer change collided node identities or leave incoming edges dangling.
I added regressions covering the real store_file_nodes_edges() lifecycle, callers to every duplicate, line-shift re-indexing, direct upsert idempotency, recursive resolution, and explicit ambiguous-target behavior.
The PR is rebased onto current main, and the updated validation is:
2,070 passed, 5 skipped, 2 xpassed
Ruff clean
mypy clean across 66 source files
Updated head: 1abf73b
The PR description has also been updated with the identity and edge-resolution trade-offs. When you have a chance, could you please re-review the revised implementation?

tirth8205:
I revalidated the updated head after the lifecycle reconciliation work. The original same-file collision is improved, but one identity transition still leaves a dangling edge: index two same-file `render` definitions, point a caller at the second semantic-suffix identity, then remove the first definition and reindex the surviving second with a harmless line shift. The survivor changes from `target.py::render:S9354c815bcc1` to bare `target.py::render`, while the caller still targets the removed suffixed identity. The new lifecycle regression fails exactly on that mismatch. This remains blocked until survivor identity is stable when a collision set shrinks; please also cover growth and reordering of the collision set.

tirth8205:
Revalidated head `1abf73b` after applying it to exact current main `329dbb91`. The submitted and relevant storage/lifecycle suites pass (94 tests), but collision-set transitions still corrupt graph identity.

- Shrink: `[A = bare, B = suffix]` becoming only `B` changes `B` to the bare key and leaves an existing caller's suffix target dangling.
- Growth: starting with unique `B` at the bare key and then inserting `A` before it gives `A` the old bare key and moves `B` to a suffix. The existing caller silently changes meaning from `B` to `A`.

This is not mergeable until identity is independent of collision-set size/order, or incoming edges are migrated atomically. Please add shrink, growth, and reorder lifecycle regressions plus explicit ambiguity behavior.

SarveshaKumarKS:
Thanks for the precise shrink and growth cases. I reproduced all three collision-set transitions and pushed a new revision.
The replace path now reconciles definitions with their previous identities before deleting the old file rows:

- semantic matches retain their identities when definitions reorder;
- semantically identical definitions use nearest previous source location across harmless line shifts;
- growth preserves the existing definition’s bare identity and gives the new definition a suffix; and
- shrink preserves the surviving definition’s suffix identity, so its existing callers remain valid.

A newly indexed bare edge endpoint resolves to that historical suffix only when it is the sole candidate. If multiple definitions remain possible, the target stays canonical and ambiguous rather than being routed arbitrarily.
Added regressions cover shrink, growth, reorder, callers to every surviving identity, bare resolution after shrink, and explicit ambiguous-target behavior.
Validation on current main (cbb74ed):

- 2,313 passed, 5 skipped, 2 xpassed
- all 75 graph storage tests passed
- Ruff clean
- mypy clean across 70 source files

Updated head: 392651a
Could you please re-run the lifecycle cases against this revision?

tirth8205:
Thanks for the continued iteration; the shrink, growth, and reorder cases from my last review now pass, and the 8 new regressions genuinely fail without the fix. Two new blockers found at head 392651a. First, a correctness regression on the common no-collision path: index a single unique function render(a) at line 10, then re-index the file with render(a, b) at line 12 (an ordinary edit that changes the signature while lines shift). The node is permanently re-keyed from /repo/app.py::render to /repo/app.py::render:Sda4664594a6e, get_node on the bare name returns None, an existing cross-file caller edge to the bare name dangles, and repeated re-indexes never restore the bare key. Root cause: in _plan_file_identities the fallback denies the bare key because it is in `reserved`, after the semantic match fails (signature changed) and the exact-location rescue fails (line moved). When the collision set for a base has exactly one old row and one new definition, they should be paired regardless of signature or line changes. Please add a regression for a lone definition whose params and line_start both change, asserting the bare identity is retained. Second, a performance regression: _resolve_stored_identity uses "substr(qualified_name, 1, ?) = ?", which EXPLAIN QUERY PLAN shows as a full index SCAN, and it runs for every edge endpoint that does not match an existing node (unresolved library-call targets are common). Measured 0.86 ms per unresolved edge at 20k nodes; a full build of this repo goes from about 3.8s on main to about 6.4s on this branch. Please replace it with an index-seekable range or LIKE query. One smaller note: symbol names containing colons (Objective-C selectors, which the parser supports via .m files) can false-match the ":S" prefix heuristic; worth a guard or at least a comment.

SarveshaKumarKS:
Thanks for the precise reproductions. I confirmed all three issues and pushed a revised implementation.
For the no-collision lifecycle, a base with exactly one old row and one new definition is now paired unconditionally. This preserves the bare identity even when both the signature and source location change, so existing incoming callers remain valid across repeated re-indexes.
_resolve_stored_identity() now searches the qualified-name index using the bounded range [base:S, base:T) instead of applying substr() to every row. EXPLAIN QUERY PLAN reports:
SEARCH nodes USING COVERING INDEX idx_nodes_qualified (qualified_name>? AND qualified_name<?)
Collision candidates are also validated against the complete generated suffix format rather than only checking the :S prefix. This prevents colon-bearing symbols such as Objective-C selectors from being mistaken for disambiguated identities.
Added regressions cover:

- a lone definition whose parameters and line both change across repeated re-indexes;
- preservation of its bare identity and existing cross-file caller;
- use of the index-bounded lookup; and
- Objective-C-style colon symbols not being treated as collision suffixes.

Validation on current main (1a010de):

- 2,409 passed, 5 skipped, 2 xpassed
- all 81 graph tests passed
- Ruff clean
- mypy clean across 70 source files

Updated head: 398adda
Could you please revalidate this revision?

tirth8205:
The lifecycle cases from my last review all pass now, and pytest, ruff, and mypy are clean on current main. Two blockers remain.

First, _plan_file_identities (graph.py:374) queries WHERE file_path = ? with its raw argument while remove_file_data normalizes. incremental.py passes str(full_path), backslash-spelled on Windows, so the plan finds no old rows and reconciliation silently no-ops there. Repro on any OS: store two same-file render defs via the spelling C:\t\target.py, point a caller at the suffixed survivor, shrink to one def, reindex with the same spelling. The survivor re-keys to bare and the caller dangles. Fix is normalize_file_path at the top of the function, plus a regression using a backslash spelling.

Second, the plan is quadratic. The old_by_base loop at graph.py:384 rescans every old row per base: reindexing 1500/3000/6000 unique defs takes 0.30/1.16/4.74s versus about 0.02s on main, and 1000 identical same-name defs cost 1.44s in the pairs loop. Bucket old rows by base in one pass.

Minor: direct upsert_node of a unique def whose line shifted now creates a suffixed duplicate instead of updating in place.

tirth8205:
Changes required: same-file collision identities still change when the file path uses backslashes. Using `GraphStore.store_file_nodes_edges`, store two `render` nodes for `C:\t\target.py` at lines 10 and 20 with params `(a)` and `(b)`, then replace that file with only `(b)` at line 21; the surviving node's qualified name changes from its semantic suffix to the bare name, leaving incoming references behind. Normalise the planning path and preserve the surviving identity before addressing the remaining repeated scans.

tirth8205:
This no longer merges into `staging`. Conflicts in `code_review_graph/graph.py`, `tests/test_graph.py`.

```bash
# resolve, then
uv run pytest tests/ -q
git push
```

When resolving:
- Rebase PR #715 onto current origin/staging and resolve conflicts in code_review_graph/graph.py (replace_file_data / replace_file_data_batch and the new _replace_file_data bulk-insert helper from the #721 performance work vs.
- After rebase, re-run pytest / ruff / mypy and re-assess edge-case coverage for qualified_name collisions.

tirth8205:
This no longer merges into `staging`. 2 files conflict and the branch is 313 commits behind:

- `code_review_graph/graph.py`
- `tests/test_graph.py`

- `code_review_graph/graph.py`: staging added the corrupt-database classifiers and a `crg_parent_dir` SQLite function registered outside the open guard. Keep that guard wrapped around exactly the statements it covers.

```bash
git fetch origin
git merge origin/staging
# resolve, then
uv run pytest tests/ -q
git push
```

PR #716, opened 2026-07-22
fix(parser): track static Class::method calls as callers in PHP and Rust (#567)

### LINKED ISSUE #567: Static/scoped calls (`Class::method`) aren't tracked as callers in PHP and Rust

kishanguptab:
## Summary

Static/scoped calls — `Mailer::dispatch($x)` in PHP, `Mailer::dispatch(x)` / `Self::new()` in Rust — were stored as `CALLS` edges whose target was the intermediate `Class::method` string. That key matches neither the canonical node name (`<file>::Class.method`) nor a bare method name, so the edge dangled and `callers_of`, `get_impact_radius`, and `tests_for` reported **zero callers** for a method that is obviously being called. In Rust this is especially costly because `Type::method()` / `Self::method()` is the *normal* way to call associated functions and constructors, so a large share of Rust call edges silently went missing.

Fixes #567.

## Approach

Adds a post-build resolver (`code_review_graph/scoped_resolver.py`) in the same style as the existing Spring / Temporal / ReScript / HCL resolvers, wired into both full build and incremental update (incremental gated to `.php`/`.rs` changes). It rewrites resolvable `Class::method` CALLS targets to the canonical qualified name of the defining method node.

It is deliberately conservative so it never fabricates an edge:

- Only genuine two-segment `Class::method` targets are considered; already-resolved `<file>::Class.method` targets are left alone (idempotent).
- Resolves on a single matching `(class, method)` node in the graph, or disambiguates between same-named definitions via the caller file's `IMPORTS_FROM` edges (resolved path or FQN namespace-segment match).
- **PHP:** namespaces are stripped before lookup (`App\Mail\Mailer` → `Mailer`); `self` / `static` (case-insensitive) resolve to the enclosing class.
- **Rust:** `method_map` is keyed by language so a PHP class never resolves a same-named Rust type; multi-segment module paths (`crate::mailer::Mailer::dispatch`) are left alone since resolving by their last two segments would point at an unrelated type; `Self` resolves to the enclosing type, while lowercase `self` is a module path (never a type method) and is left untouched.
- Unresolved external targets (`Vec::new`, `Redis::get`, …) are left exactly as they behave today — recorded, not falsely resolved.
- Rewritten edges are tagged `confidence_tier = INFERRED` to distinguish a heuristically resolved scoped call from a directly-extracted one.

## Before / after repro

Two files per language: a class/struct `Mailer` defining `dispatch`, and a function `register` calling `Mailer::dispatch(...)`.

**PHP**

| | `callers_of dispatch` | stored CALLS target |
|---|---|---|
| before (main) | `0 results` | `Mailer::dispatch` (dangling, `EXTRACTED`) |
| after | `1 result` → `register` | `<...>/Mailer.php::Mailer.dispatch` (`INFERRED`) |

**Rust**

| | `callers_of dispatch` | stored CALLS target |
|---|---|---|
| before (main) | `0 results` | `Mailer::dispatch` (dangling, `EXTRACTED`) |
| after | `1 result` → `register` | `<...>/mailer.rs::Mailer.dispatch` (`INFERRED`) |

`get_impact_radius(changed_files=["src/Mailer.php"])` / `["src/mailer.rs"]` now includes `register` in the impacted nodes.

## Tests

- `tests/test_php_scoped_calls.py` (8 tests): cross-file resolution, `INFERRED` tagging, impact-radius includes the caller, namespaced `use`-import resolution, import-based disambiguation of same-named classes, external target left untouched, incremental re-resolution, idempotency.
- `tests/test_rust_scoped_calls.py` (7 tests): cross-file resolution, `INFERRED` tagging, `Type::new()` constructor, `Self::` → enclosing type, impact-radius includes the caller, external stdlib target (`Vec::new`) left untouched, multi-segment module path left untouched.

`pytest` (full suite), `ruff check code_review_graph/`, `mypy code_review_graph/ --ignore-missing-imports --no-strict-optional`, and `bandit -r code_review_graph/ -c pyproject.toml` all pass. (One pre-existing, unrelated failure — `test_windows_server_still_prewarms_before_mcp_run`, a Windows event-loop test — fails identically on clean `main` in this macOS/Python 3.14 environment.)

## Note on #574 (PHP `use` imports)

Verified as **already fixed on current `main`** — no residual gap, so this PR does not touch PHP imports. `_extract_import` has a dedicated `elif language == "php":` branch that records clean FQNs (handling `as` aliases, grouped `use A\{B, C}`, and `use function`/`use const`), and `_do_resolve_module` has the matching PHP branch that maps the FQN to an absolute `.php` path. Running the #574 reproducer against `main`, `IMPORTS_FROM` edges now target the resolved path of the imported class file and `importers_of(Job.php)` returns the two importing files (`MatchService.php`, `JobController.php`) — it no longer stores the raw `use ...;` statement text.

## Scope

PHP and Rust only — the two languages that actually produce a dangling `Class::method` edge. C++/Ruby produce no CALLS edge for scoped calls (a separate extraction gap) and R's `::` is external-package access; all out of scope here, consistent with the issue.

### DIFF
--- a/code_review_graph/incremental.py
+++ b/code_review_graph/incremental.py
@@ -122,6 +122,16 @@ def _run_hcl_resolver(store: GraphStore) -> Optional[dict]:
         logger.warning("Terraform/HCL resolver failed: %s", exc)
         return None
 
+
+def _run_scoped_resolver(store: GraphStore) -> Optional[dict]:
+    """Resolve static/scoped ``Class::method`` calls without failing a build."""
+    try:
+        from .scoped_resolver import resolve_scoped_calls
+        return resolve_scoped_calls(store)
+    except Exception as exc:  # noqa: BLE001 - best-effort post-pass
+        logger.warning("Scoped call resolver failed: %s", exc)
+        return None
+
 # Default ignore patterns (in addition to .gitignore).
 #
 # ``**/<dir>/**`` patterns are safe-anywhere directory exclusions.  A leading
@@ -986,6 +996,7 @@ def full_build(
     spring_event_stats = _run_spring_event_resolver(store)
     temporal_stats = _run_temporal_resolver(store)
     hcl_stats = _run_hcl_resolver(store)
+    scoped_stats = _run_scoped_resolver(store)
 
     return {
         "files_parsed": len(files),
@@ -997,6 +1008,7 @@ def full_build(
         "event_resolution": spring_event_stats,
         "temporal_resolution": temporal_stats,
         "hcl_resolution": hcl_stats,
+        "scoped_resolution": scoped_stats,
     }
 
 
@@ -1131,6 +1143,8 @@ def incremental_update(
     temporal_stats = _run_temporal_resolver(store) if spring_changed else None
     hcl_changed = any(rp.endswith((".tf", ".hcl")) for rp in all_files)
     hcl_stats = _run_hcl_resolver(store) if hcl_changed else None
+    scoped_changed = any(rp.endswith(".php") for rp in all_files)
+    scoped_stats = _run_scoped_resolver(store) if scoped_changed else None
 
     return {
         "files_updated": len(all_files),
@@ -1144,6 +1158,7 @@ def incremental_update(
         "event_resolution": spring_event_stats,
         "temporal_resolution": temporal_stats,
         "hcl_resolution": hcl_stats,
+        "scoped_resolution": scoped_stats,
     }
 
 
--- a/code_review_graph/incremental.py
+++ b/code_review_graph/incremental.py
@@ -1143,7 +1143,7 @@ def incremental_update(
     temporal_stats = _run_temporal_resolver(store) if spring_changed else None
     hcl_changed = any(rp.endswith((".tf", ".hcl")) for rp in all_files)
     hcl_stats = _run_hcl_resolver(store) if hcl_changed else None
-    scoped_changed = any(rp.endswith(".php") for rp in all_files)
+    scoped_changed = any(rp.endswith((".php", ".rs")) for rp in all_files)
     scoped_stats = _run_scoped_resolver(store) if scoped_changed else None
 
     return {
--- a/code_review_graph/scoped_resolver.py
+++ b/code_review_graph/scoped_resolver.py
@@ -1,0 +1,257 @@
+"""Post-build resolver for static/scoped ``Class::method`` calls.
+
+Tree-sitter extraction records a scoped/static call such as ``Mailer::send($x)``
+(PHP) or ``Mailer::send(x)`` / ``Self::new()`` (Rust) as a ``CALLS`` edge whose
+target is the intermediate ``Class::method`` string.  That string matches
+neither the canonical node key (``<file>::Class.method``) nor a bare method
+name, so ``callers_of``, ``get_impact_radius`` and ``tests_for`` never see the
+edge and report zero callers for a method that is obviously being called
+(GitHub #567).
+
+This module runs after the graph is built and rewrites the resolvable
+``Class::method`` targets to the canonical qualified name of the defined
+method node, in the same style as the Spring/Temporal/ReScript resolvers.
+
+It is deliberately conservative so it never fabricates an edge:
+
+* Only genuine ``Class::method`` targets are considered; already-resolved
+  targets (``<file>::Class.method``) are left alone.
+* ``self`` / ``static`` (PHP) and ``self`` / ``Self`` (Rust) resolve to the
+  enclosing class/type of the caller.
+* A target resolves when exactly one ``(class, method)`` node exists in the
+  graph, or when the caller file's ``IMPORTS_FROM`` edges disambiguate between
+  several same-named definitions.
+* Ambiguous or unknown targets (external types such as ``Vec::new`` or
+  ``Redis::get`` with no in-graph definition) are left untouched, keeping the
+  behaviour of unresolved external calls exactly as it is today.
+
+Rewritten edges are tagged ``confidence_tier = INFERRED`` to distinguish a
+heuristically resolved scoped call from a directly-extracted one.
+"""
+
+from __future__ import annotations
+
+import json
+import logging
+from typing import TYPE_CHECKING, Optional
+
+if TYPE_CHECKING:
+    from .graph import GraphStore
+
+logger = logging.getLogger(__name__)
+
+# Languages whose scoped/static calls dangle as ``Class::method`` targets.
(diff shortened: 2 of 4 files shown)


jordandunmire97-ai (approved):
Hope to assist with any reviews

tirth8205:
The 15 focused PHP/Rust tests pass, but the resolver case-folds both class and method names for Rust. Rust identifiers are case-sensitive, so this can create a CALLS edge to a differently cased definition. The import fallback also matches only one namespace segment and can select an unrelated same-named definition. Please case-fold PHP only, keep exact Rust identifiers, suffix-match imported namespaces/paths, and add case-collision and unrelated-path tests.

tirth8205:
Maintainer validation found one remaining correctness bug: unrelated imports sharing only a partial `Queue/Mailer` suffix could fabricate a CALLS link in both PHP and Rust. I preserved this contribution against current `main`, credited you as co-author, added complete-suffix/Windows-path safeguards and RED→GREEN regressions, and opened the tested integration as #752. Exact-head validation: 822 affected tests passed, 26 focused tests passed, Ruff/diff passed, and the full graph build completed with zero errors. I will leave this PR open until #752 passes hosted checks and is merged.

tirth8205:
The corrected integration #752 is now merged at `d6b4ad1` with your contribution credited. The final resolver keeps PHP case-insensitive and Rust case-sensitive behavior, normalizes Windows paths, and refuses ambiguous partial import-suffix matches. It passed 822 affected tests, 26 focused regressions, a full graph build, CodeQL, and every hosted job including Windows. Closing this original PR as fully incorporated.

PR #718, opened 2026-07-23
fix(parser): detect PHPUnit @test docblocks and #[Test] attributes

### LINKED ISSUE #693: PHP: tests marked with /** @test */ or #[Test] are classified as production functions (no PHP branch in decorator extraction)
## Summary

PHP test methods are only recognised by the `test_` name prefix. Neither the PHPUnit docblock annotation `/** @test */` nor the PHP 8 attribute `#[Test]` is detected, so those methods are classified as production `Function` nodes with `is_test: false`.

This is the same class of bug as #295 (C# `attribute_list`), which was fixed by adding a language branch to the decorator extraction.

Version: 2.3.7

## Reproduction

`tests/ExampleTest.php` in an otherwise empty repo:

```php
<?php

namespace Tests;

use PHPUnit\Framework\TestCase;
use PHPUnit\Framework\Attributes\Test;

class ExampleTest extends TestCase
{
    // control: prefix -> detected
    public function test_prefixed_method_should_be_detected(): void
    {
        $this->assertTrue(true);
    }

    /** @test */
    public function docblock_annotated_method(): void
    {
        $this->assertTrue(true);
    }

    #[Test]
    public function php8_attribute_annotated_method(): void
    {
        $this->assertTrue(true);
    }

    // control: no marker -> correctly not detected
    public function helperNotATest(): void
    {
    }
}
```

```
sqlite3 .code-review-graph/graph.db \
  "SELECT kind, name, is_test FROM nodes WHERE kind IN ('Function','Test') ORDER BY name;"
```

### Actual

```
Function|docblock_annotated_method|0            <-- expected Test
Function|helperNotATest|0                       <-- correct
Function|php8_attribute_annotated_method|0      <-- expected Test
Test|test_prefixed_method_should_be_detected|1  <-- correct
```

### Expected

Both annotated methods classified as `Test` with `is_test: 1`. The two controls confirm the check discriminates in both directions, so the two misses are not a measurement artifact.

## Root cause

`parser.py` `_extract_functions` (~L9160-9190) builds `deco_list` with a branch per language — Python `decorated_definition`, Rust `attribute_item`, C# `attribute_list` (added in #295), Java `modifiers > annotation`. There is no PHP branch, so `decorators` is always empty for PHP and `_is_test_function` falls through to name matching only.

`_TEST_ANNOTATIONS` already contains `"Test"`, so `#[Test]` would work as soon as PHP attributes are extracted.

## Impact

`detect_changes` lists the test methods themselves under `test_gaps` as untested production code, and their risk contribution inflates the overall `risk_score`. On a real Laravel service (164 files, 794 nodes) 59 of 88 test methods use `/** @test */` and are invisible, so the graph reports 29 tests and 167 `TESTED_BY` edges built from a third of the suite. Reviewers get pointed at code that is in fact covered.

## Suggested fix

Two separate paths, since they are different node shapes:

1. `#[Test]` attributes — add a PHP branch to the decorator extraction, reading `attribute_list > attribute > name` from the method's children. Mirrors the C# fix in #295; `_TEST_ANNOTATIONS` needs no change.
2. `/** @test */` docblocks — this is a `comment` node preceding the method, not an attribute, so it needs its own check: inspect the previous sibling comment for an `@test` tag. Worth handling because it is still the dominant style in existing PHPUnit suites (deprecated in PHPUnit 11, but widespread in the installed base).

Happy to send a PR for either or both if useful.

arimu1:
## Summary

PHP test methods are currently recognized only by the `test_` name prefix. Neither PHPUnit's legacy `/** @test */` docblock tag nor the PHP 8 `#[Test]` attribute is detected, so methods using either style are classified as production `Function` nodes with `is_test: false` instead of `Test` nodes.

This is the same class of gap as the C# `attribute_list` fix (#295): PHP just never got a language branch in the decorator/annotation extraction in `_extract_functions`, so `decorators` was always empty for PHP and `_is_test_function` fell through to name matching only. `_TEST_ANNOTATIONS` already contains `"Test"`, so `#[Test]` works as soon as PHP attributes are captured.

## Fix

Two separate paths, since they're different node shapes in tree-sitter-php:

- **`#[Test]` attributes** — a new `_php_attribute_names` helper reads `attribute_list > attribute_group > attribute > name` from the method's children (one level deeper than C#'s `attribute_list > attribute`, since PHP wraps each `#[...]` group in an intermediate `attribute_group` node).
- **`/** @test */` docblocks** — not an attribute at all; it's a `comment` node that's a *preceding sibling* of the `method_declaration`, not a child. A new `_php_docblock_marks_test` helper checks the previous sibling for the `@test` tag and, when found, folds `"Test"` into the existing decorator list so it flows through the existing `_TEST_ANNOTATIONS` check unchanged.

Both wire into the existing PHP branch alongside C#'s in `_extract_functions`.

## Test evidence

Added `TestPHPTestAnnotations` in `tests/test_multilang.py`, covering all four cases from the issue in one fixture:

- `test_`-prefixed method — still detected (control)
- `/** @test */` docblock — now detected
- `#[Test]` attribute — now detected
- plain method with no marker — still not detected (control)

```
$ uv run pytest tests/test_multilang.py -k PHPTestAnnotations -v
tests/test_multilang.py::TestPHPTestAnnotations::test_name_prefix_still_detected PASSED
tests/test_multilang.py::TestPHPTestAnnotations::test_docblock_annotation_detected PASSED
tests/test_multilang.py::TestPHPTestAnnotations::test_php8_attribute_detected PASSED
tests/test_multilang.py::TestPHPTestAnnotations::test_plain_method_not_detected PASSED
4 passed
```

Confirmed the two new-detection tests fail without the fix (stashed `parser.py` only, tests unchanged):

```
tests/test_multilang.py::TestPHPTestAnnotations::test_docblock_annotation_detected FAILED
tests/test_multilang.py::TestPHPTestAnnotations::test_php8_attribute_detected FAILED
2 failed, 2 passed
```

Full suite: `ruff check` clean, `mypy --ignore-missing-imports --no-strict-optional` clean, `pytest --cov=code_review_graph --cov-fail-under=65`: **1965 passed, 5 skipped, 2 xpassed**, coverage 79.81%.

Closes #693.

### DIFF
--- a/code_review_graph/parser.py
+++ b/code_review_graph/parser.py
@@ -1732,6 +1732,55 @@ def _csharp_attribute_names(node) -> list[str]:
     return names
 
 
+def _php_attribute_names(node) -> list[str]:
+    """Return PHP attribute names from ``attribute_list`` children of *node*.
+
+    PHP 8 attributes (``#[Test]``, ``#[DataProvider('x')]``) are
+    ``attribute_list`` nodes wrapping one or more ``attribute_group`` nodes,
+    each wrapping one or more ``attribute`` nodes whose ``name`` child is the
+    attribute name -- one level deeper than C#'s ``attribute_list >
+    attribute``. See: #693
+    """
+    names: list[str] = []
+    for sub in node.children:
+        if sub.type != "attribute_list":
+            continue
+        for group in sub.children:
+            if group.type != "attribute_group":
+                continue
+            for attr in group.children:
+                if attr.type != "attribute":
+                    continue
+                for ident in attr.children:
+                    if ident.type == "name":
+                        names.append(
+                            ident.text.decode("utf-8", errors="replace").strip(),
+                        )
+                        break
+    return names
+
+
+_PHP_TEST_DOC_TAG_RE = re.compile(r"@test\b")
+
+
+def _php_docblock_marks_test(node) -> bool:
+    """Return True if a PHPUnit ``/** @test */`` docblock precedes *node*.
+
+    PHPUnit's legacy convention marks a test method with an ``@test`` tag in
+    the doc comment immediately above it, instead of a ``test_`` name prefix
+    or a ``#[Test]`` attribute. Tree-sitter represents that comment as a
+    preceding sibling of the ``method_declaration``, not a child, so it needs
+    its own check rather than reuse of the attribute-based extraction above.
+    See: #693
+    """
+    sib = node.prev_sibling
+    return bool(
+        sib is not None
+        and sib.type == "comment"
+        and _PHP_TEST_DOC_TAG_RE.search(sib.text.decode("utf-8", errors="replace"))
+    )
+
+
 def _csharp_namespaces(root_node) -> list[str]:
     """Return all namespaces declared in a C# compilation unit.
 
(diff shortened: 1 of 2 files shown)


tirth8205:
The supplied four PHPUnit tests pass, but two correctness cases still fail. A qualified attribute such as #[\PHPUnit\Framework\Attributes\Test] is classified as Function, while @test-case is falsely classified as @test. Please normalize qualified, aliased, and grouped attribute names, recognize only a standalone PHPDoc @test tag, and add those positive and negative cases.

PR #729, opened 2026-07-23
fix(incremental): purge old paths on rename — parse git diff --name-status

### LINKED ISSUE #684: [Bug]: File rename leaves stale nodes/edges — incremental update diverges from full rebuild

Qalipso:
Fixes #684.

## Problem
`git diff --name-only` reports a rename `a.py → b.py` as **only the new path**, so `incremental_update()` never sees `a.py`. The purge loop only removes paths present in the changed/dependent set, so the old path's File/Function/Class nodes and every edge keyed to it survive — an incremental update permanently diverges from a full rebuild (stale nodes, dangling importer edges, inflated counts).

## Fix
- Change discovery now uses `git diff --name-status -z` (both in the primary call and the `--cached` fallback).
- New `_decode_name_status_paths()` parses the NUL-delimited records and emits **both** sides of `R*`/`C*` entries (order-preserving, deduped). The old path then flows through the existing purge loop's `is_file()` check and is removed — no changes needed to the purge logic itself.
- `_decode_nul_paths` had no remaining callers and is removed.

## Tests
- Unit: record parser — `M/A/D`, `R100` (two paths), `C75`, dedupe, empty input.
- `get_changed_files` with a mocked rename returns both paths.
- **End-to-end repro from the issue:** real git repo, `git mv a.py b.py` + commit, `get_changed_files(base="HEAD~1")` → `{a.py, b.py}`, then `incremental_update` — asserts the old path is fully purged and the new one is present (full-rebuild parity).
- Updated the three existing `get_changed_files` mocks to the `--name-status` output format.

Full suite: **1966 passed, 5 skipped**. `ruff check` clean.

AI-assisted; every change reviewed and understood.

### DIFF
--- a/code_review_graph/incremental.py
+++ b/code_review_graph/incremental.py
@@ -534,9 +534,29 @@ _SAFE_GIT_REF = re.compile(r"^[A-Za-z0-9_.~^/@{}\-]+$")
 _SAFE_SVN_REV = re.compile(r"^r?\d+(:r?\d+|:HEAD|:BASE|:COMMITTED)?$", re.IGNORECASE)
 
 
-def _decode_nul_paths(output: bytes) -> list[str]:
-    """Decode Git's NUL-delimited path output without quote mangling."""
-    return [os.fsdecode(path) for path in output.split(b"\0") if path]
+def _decode_name_status_paths(output: bytes) -> list[str]:
+    """Decode ``git diff --name-status -z`` output into a list of paths.
+
+    Renames and copies (``R<score>``/``C<score>`` records) carry two paths —
+    the old and the new one.  Both are emitted so the old path flows through
+    the purge loop in :func:`incremental_update`; otherwise a rename leaves
+    the old path's nodes and edges in the graph and the incremental result
+    diverges from a full rebuild.
+    """
+    fields = [os.fsdecode(f) for f in output.split(b"\0") if f]
+    paths: list[str] = []
+    seen: set[str] = set()
+    i = 0
+    while i < len(fields):
+        status = fields[i]
+        takes_two = status[:1] in ("R", "C")
+        entry = fields[i + 1 : i + (3 if takes_two else 2)]
+        i += 3 if takes_two else 2
+        for path in entry:
+            if path not in seen:
+                seen.add(path)
+                paths.append(path)
+    return paths
 
 
 def _store_vcs_metadata(repo_root: Path, store: "GraphStore") -> None:
@@ -571,8 +591,10 @@ def get_changed_files(repo_root: Path, base: str = "HEAD~1") -> list[str]:
         logger.warning("Invalid git ref rejected: %s", base)
         return []
     try:
+        # --name-status (not --name-only): renames/copies must report BOTH
+        # paths, or the old path never reaches the purge loop (issue #684).
         result = subprocess.run(
-            ["git", "diff", "--name-only", "-z", base, "--"],
+            ["git", "diff", "--name-status", "-z", base, "--"],
             capture_output=True,
             cwd=str(repo_root),
             timeout=_GIT_TIMEOUT,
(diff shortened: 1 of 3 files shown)


jondkinney:
I reviewed this alongside #686, which fixes the same issue with a very similar approach — noting it here so the maintainer can compare them easily, since the two would conflict if both landed.

My read is that each PR has one thing the other would benefit from:

**This PR's parser implementation looks the more robust of the two.** Using a slice (`fields[i + 1 : i + 3]`) means a truncated final record clamps harmlessly instead of relying on an index guard, and the `seen` set dedupes paths. #686 uses index arithmetic with an `i + 1 < n` check, where a malformed trailing `R100` carrying only one path falls through and is treated as an ordinary single-path entry. Git shouldn't emit that, so it's a degenerate case rather than a real bug — but the slicing approach here sidesteps it by construction.

**#686's tests are the stronger of the two.** It adds `test_get_changed_files_reports_both_rename_paths`, which asserts the exact expected output for a rename record (`["old/a.py", "new/b.py", "keep.py"]`), plus an assertion that `--name-status` is actually in the git argv. This PR updates the existing fixtures to the new format and imports `_decode_name_status_paths`, but doesn't add a rename-specific case pinning both paths.

Combining the two — this parser with #686's tests — would probably give the best result, if that's helpful.

I checked the assumption both PRs depend on and it holds: `incremental.py:1053` does `if not abs_path.is_file(): store.remove_file_data(...)`, so surfacing the old path is sufficient for it to be purged. Both correctly leave `D` records alone, since `--name-only` already surfaced deletions.

One scope note that applies equally to both: this fixes the git-diff path, while `start_watch_thread()`'s `GraphUpdateHandler` has no `on_moved` handler, so `--watch` users would still accumulate stale nodes after a rename. I opened #742 for that separately — not a gap in this PR, just related.

Thanks for working on this.

PR #733, opened 2026-07-24
fix(graph): scope bare CALLS/INHERITS name fallback to caller's language

### LINKED ISSUE #708: [Bug]: Cross boundry CALLS incorrect
### code-review-graph version

2.3.7

### Operating system

Windows

### Python version

3.14.6

### AI platform

other (describe in "Additional context")

```shell
code-review-graph status
Nodes: 27556
Edges: 159209
Files: 3478
Languages: javascript, typescript, yaml, apex
Last updated: 2026-07-21T15:33:26
```

### Steps to reproduce

A apex class calling .clone on a variable instance in it's class will report "CALLS" to any javascript ES6 which has a method named `clone`. 

detection of calls needs to correctly scope within the correct context

### Additional context

_No response_

vasunandan:
## Summary
- Bare (unqualified) CALLS/INHERITS target names are matched by exact string across the *entire* graph when qualified-name resolution fails — with no language filter, a common short name (e.g. `clone`) can wrongly resolve to a same-named node in a completely unrelated language.
- Adds an optional `language` filter to `search_edges_by_target_name` / `iter_edges_by_target_name` (`graph.py`), and threads the resolved target node's language through the `callers_of` and `inheritors_of` bare-name fallbacks in `query.py`.
- Fixes #708 ("Cross boundary CALLS incorrect" — an Apex `.clone()` call was reported as calling any same-named JS `clone()` method anywhere in the graph).

## Scope note
Left the 3 other `search_edges_by_target_name` call sites in `refactor.py` (rename preview / dead-code detection) untouched to keep this change small and reviewable. Happy to extend the same fix there in a follow-up if wanted.

## Test plan
- [x] Added a regression test (`test_callers_of_bare_fallback_does_not_cross_languages`) reproducing the exact Apex→JS `clone` collision from the issue.
- [x] `pytest tests/test_tools.py tests/test_graph.py tests/test_refactor.py tests/test_hints.py` — 235 passed
- [x] `pytest tests/` — 1959 passed, 1 pre-existing unrelated failure (Windows-only asyncio test failing under Python 3.14 on macOS, unrelated to this change)
- [x] `ruff check` and `mypy` clean on changed files

### DIFF
--- a/code_review_graph/graph.py
+++ b/code_review_graph/graph.py
@@ -398,7 +398,9 @@ class GraphStore:
             ).fetchall())
         return [self._row_to_edge(row) for row in rows]
 
-    def search_edges_by_target_name(self, name: str, kind: str = "CALLS") -> list[GraphEdge]:
+    def search_edges_by_target_name(
+        self, name: str, kind: str = "CALLS", language: str | None = None,
+    ) -> list[GraphEdge]:
         """Search for edges where target_qualified matches an unqualified name.
 
         CALLS edges often store unqualified target names (e.g. ``generateTestCode``)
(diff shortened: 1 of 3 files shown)


tirth8205:
Thanks for addressing the Apex false match. The exact head passed 169 focused tests (1 skipped), but the fix is too strict to merge: exact language equality drops valid unresolved calls across the JavaScript family (javascript, typescript, tsx, and Astro’s TypeScript representation). The rename, dead-code, and inheritance paths in refactor.py also still use unfiltered bare-name lookups, so the same false-positive class remains there. Please use compatible language families across every bare CALLS/INHERITS fallback and add positive JS/TS/TSX/JSX/Astro plus INHERITS coverage.

tirth8205:
The cross-language diagnosis is correct, but exact-language equality breaks valid JavaScript-family edges (JS/JSX/TS/TSX/Astro) and the filter was not applied in every query, rename, and dead-code path. I preserved your commit, resolved the current-main conflict, added six RED→GREEN compatibility/negative regressions, and opened the corrected integration as #754. Exact-head validation: 221 tests passed (1 skipped), Ruff/diff passed, and a 242-file graph build completed with zero errors. I will keep this PR open until #754 passes hosted checks and is merged.

tirth8205:
Validated and incorporated through #754, merged as ce2e394. Your original commit is preserved in the integration history. The original language-equality guard fixed cross-language false links but broke compatible JavaScript/TypeScript-family linking; #754 keeps the intended isolation while allowing valid JS/JSX/TS/TSX links and keeping Apex separate. Validation passed 221 graph/query/refactor tests, six focused regressions, Ruff, diff checks, a full graph build, CodeQL, Python 3.10–3.13, and Windows. Closing this original PR because its corrected form is now on main.

PR #738, opened 2026-07-25
fix(parser): index TS interfaces/type aliases and emit type-position edges

### LINKED ISSUE #737: [Bug]: TypeScript/TSX interfaces and type aliases are never indexed — callers_of/impact on a type returns not_found (Java/C#/PHP index interfaces)
### code-review-graph version

2.3.7 (`pip install -e .` from a `main` checkout at `805c4c6`, 2026-07-25)

### Operating system

macOS 26.6

### Python version

3.14.6

### AI platform

claude-code

```shell
$ code-review-graph status
Nodes: 3
Edges: 2
Files: 2
Languages: typescript
Last updated: 2026-07-25T10:25:28
Built on branch: main
Built at commit: 9fc52404ed38
```

(3 nodes for a 2-file repo that declares an `interface` and a `type` — see below.)

## Summary

For `typescript` and `tsx`, `_CLASS_TYPES` lists only `class_declaration` / `class`. `interface_declaration` and `type_alias_declaration` are absent, so TS interfaces and type aliases are **never indexed as nodes**. Java, C#, PHP, and Solidity all index `interface_declaration`, so the behaviour is asymmetric across languages rather than a deliberate TS design choice.

Consequence: `callers_of` / `tests_for` / `get_impact_radius` on a TypeScript type return `not_found`, and a types-only module (`types.ts`, `*.d.ts` — very common in TS/React codebases) contributes **zero symbol nodes** to the graph. Its blast radius degrades to whole-file `IMPORTS_FROM` fan-out, which is both under-precise (every importer is impacted, no matter which of the 40 interfaces in the file changed) and unqueryable at the symbol level.

## Steps to reproduce

```
repro/
  src/types.ts
  src/use.ts
```

`src/types.ts`:

```ts
export interface Finding {
  id: string;
  severity: string;
}

export type Verdict = "ok" | "bad";
```

`src/use.ts`:

```ts
import { Finding, Verdict } from "./types";

export function summarize(items: Finding[]): Verdict {
  return items.length ? "bad" : "ok";
}
```

```
git init && git add -A && git commit -m init
code-review-graph build
code-review-graph query callers_of "src/types.ts::Finding"
code-review-graph impact --files src/types.ts
```

## Expected vs actual behavior

**Expected:** `Finding` and `Verdict` are nodes (as `interface Finding` is in Java), `callers_of` reports `summarize`, and `impact` names the specific types that changed.

**Actual:** neither type exists in the graph.

```
Full build: 2 files, 3 nodes, 2 edges (postprocess=full)
```

Nodes (`kind, name, qualified_name`):

```
('File',     'src/types.ts',            '.../src/types.ts')
('File',     'src/use.ts',              '.../src/use.ts')
('Function', 'summarize',               '.../src/use.ts::summarize')
```

Edges:

```
('CONTAINS',     '.../src/use.ts', '.../src/use.ts::summarize')
('IMPORTS_FROM', '.../src/use.ts', '.../src/types.ts')
```

`callers_of`:

```json
{ "status": "not_found", "summary": "No node found matching 'src/types.ts::Finding'." }
```

`impact --files src/types.ts` — the changed set is the opaque File node, so there is no way to distinguish "changed `Finding.severity`" from "reformatted a comment":

```
Blast radius for 1 changed file(s):
  - 1 nodes directly changed
  - 2 nodes impacted (within 2 hops)
  - 1 additional files affected
changed:  ['File:src/types.ts']
impacted: ['File:src/use.ts', 'Function:summarize']
```

## Additional context

**Cross-language control.** The same shape in Java indexes fine — `interface Finding` becomes a `Class` node:

```java
// src/Finding.java
public interface Finding { String id(); }
```

```
('Class',    'Finding',    '.../src/Finding.java::Finding')
('Function', 'id',         '.../src/Finding.java::Finding.id')
```

**Root cause** — `code_review_graph/parser.py:806-813`:

```python
_CLASS_TYPES: dict[str, list[str]] = {
    ...
    "typescript": ["class_declaration", "class"],
    "tsx":        ["class_declaration", "class"],
    ...
    "java":       ["class_declaration", "interface_declaration", "enum_declaration"],
```

`grep -rn type_alias_declaration` over the package returns nothing (the one `type_alias` hit at `parser.py:11287` is the Rust branch).

**Not the same as the JSX gap** — `_extract_jsx_component_call` (`parser.py:10843`) already resolves `<Comp />` renders into CALLS edges, so component *values* are covered; only type *declarations* fall through.

**Suggested fix**, in two parts:

1. Add `interface_declaration` and `type_alias_declaration` to `_CLASS_TYPES["typescript"]` and `_CLASS_TYPES["tsx"]`. That alone makes types addressable nodes and is the step already documented in CONTRIBUTING's "Adding Language Support". Property signatures inside an interface may want the `_FUNCTION_TYPES` treatment too (`method_signature`, `property_signature`) to match how Java exposes `Finding.id`.

2. Emit edges from type *positions* so `callers_of` on a type is non-empty: `type_annotation`, type arguments inside `generic_type`, and `implements_clause` / `extends_type_clause`. Worth keeping these out of `CALLS` — a type reference isn't an invocation, and folding it into `CALLS` (weight 1.0 in `IMPACT_EDGE_WEIGHTS`) would over-rank type churn. The existing `REFERENCES` (0.6) looks like the right home, or a dedicated kind.

For what it's worth, we hit exactly this split building a blast-radius feature in a different (Rust) reviewer and ended up keeping three separate buckets — calls + JSX renders, type-position uses, and test references — precisely because collapsing type uses into callers made the impact ranking noisy on TS codebases. Happy to share the classification rules if useful, or to put up a PR for part 1 if you'd like it split that way.

nhatvu148:
Fixes #737.

## Problem

`_CLASS_TYPES` listed only `class_declaration` / `class` for `typescript` and `tsx`, so **interfaces, type aliases, and enums were never indexed**. Java, C#, PHP, and Solidity already index `interface_declaration`, so this was an asymmetry rather than a deliberate TS scope decision.

The sharp edge is a types-only module — `types.ts`, `*.d.ts`, very common in TS/React codebases. It contributed **zero symbol nodes**, so its blast radius collapsed to whole-file `IMPORTS_FROM` fan-out: every importer impacted, no matter which of the file's declarations actually changed, and nothing addressable to query.

## Changes

**1. Index TS type declarations.** `interface_declaration`, `type_alias_declaration`, and `enum_declaration` added to `_CLASS_TYPES` for `typescript`/`tsx`. They become `Class` nodes, matching how Go `type_declaration` and Rust `struct_item`/`trait_item` are already handled — no new node kind.

**2. Emit `REFERENCES` from type positions.** `function summarize(items: Finding[])` is a real dependency on `Finding`, but a type annotation is not call syntax, so neither `_extract_calls` nor `_extract_value_references` ever saw it. New `_extract_ts_type_reference` covers annotations and type arguments, reusing the existing `_emit_reference_if_known` gate (must be a local definition or an import, so `string`/`Map`/`Promise` produce no noise).

`REFERENCES` (0.6 in `IMPACT_EDGE_WEIGHTS`) rather than `CALLS` (1.0) on purpose — a type use is a weaker signal than an invocation, and folding them together would let type churn outrank call churn in impact ranking. Two positions are deliberately skipped: the declaration's own name, and heritage clauses that already produce `INHERITS`.

Edges attribute to the enclosing function, else the enclosing type, else the file — so `interface Wrapper { nested: Verdict }` names `Wrapper` as the dependent instead of collapsing to the module.

**3. Fix `_get_bases` for TS/TSX** — *this one is a pre-existing bug, independent of the above, found while testing.* Class heritage nests under `class_heritage`, but the TS branch scanned only direct children, so `class Impl extends Base implements Findable` emitted **no `INHERITS` edges at all**. Interfaces use `extends_type_clause`, which wasn't listed either. Generic bases (`extends Base<T>`) now resolve to the generic's head.

Happy to split this third change into its own PR if you'd rather review it separately — I bundled it because newly-indexed interfaces would otherwise have no heritage edges, which undercuts change 1.

## Measured on this repo's own `code-review-graph-vscode` extension (17 `.ts` files)

| | before | after |
|---|---:|---:|
| `Class` nodes | 15 | **40** |
| `REFERENCES` edges | 4 | **164** |
| `INHERITS` edges | 0 | **5** |
| `CALLS` edges | 1095 | 1095 (unchanged) |
| full build | 0.93s | 1.04s |

`CALLS` being byte-identical is the point: no existing call extraction is disturbed.

Minimal repro from the issue, before → after:

```
# before
Full build: 2 files, 3 nodes, 2 edges
query callers_of "src/types.ts::Finding"  ->  {"status": "not_found"}
impact --files src/types.ts               ->  changed: ['File:src/types.ts']

# after
Full build: 2 files, 5 nodes, 6 edges
impact --files src/types.ts               ->  changed: ['File:src/types.ts',
                                                        'Class:Finding', 'Class:Verdict']
                                              impacted: ['Function:summarize', 'File:src/use.ts']
inheritors_of "src/base.ts::Base"         ->  ['Impl']   (was empty — no INHERITS edges existed)
```

## Known limit, stated plainly

`callers_of` filters `kind == "CALLS"`, so it still returns 0 results for a type — it now returns `ok` with an empty list rather than `not_found`. The gains here are impact radius granularity, `inheritors_of` working for TS at all, and dead-code accuracy. If you want type users directly queryable, a `references_to` pattern in `query.py` would be the additive way to do it (I'd rather not widen `callers_of`'s documented "functions that call a given function" semantics for every language) — happy to add it here or in a follow-up.

## Tests

10 new tests in `TestTypeScriptTypeDeclarations` (`tests/test_parser.py`), covering nodes for all three declaration forms, annotation and generic-argument references, TSX coverage, interface-member attribution, both heritage forms, and three negative guards: no self-reference from the declaration name, no builtin/unimported noise, no `REFERENCES`/`INHERITS` double-emit on heritage.

7 of the 10 fail on the unpatched parser (the 3 guards pass trivially when no edges exist).

## Checks

- `pytest tests/` — 2075 passed, 5 skipped, 2 xpassed (2065 before this PR)
- `ruff check code_review_graph/` — clean
- `mypy code_review_graph/ --ignore-missing-imports --no-strict-optional` — Success: no issues found in 66 source files
- `dead-code` does not flag the newly indexed type nodes

### DIFF
--- a/code_review_graph/parser.py
+++ b/code_review_graph/parser.py
@@ -804,8 +804,19 @@ _TASK_META_KEYS: frozenset[str] = frozenset({
 _CLASS_TYPES: dict[str, list[str]] = {
     "python": ["class_definition"],
     "javascript": ["class_declaration", "class"],
-    "typescript": ["class_declaration", "class"],
-    "tsx": ["class_declaration", "class"],
+    # TS types are declarations, not just runtime classes: an interface or type
+    # alias is the thing callers depend on, so it needs a node of its own the way
+    # Java/C#/PHP interfaces do. Without them a types-only module (types.ts,
+    # *.d.ts) contributes zero symbol nodes and its blast radius collapses to
+    # whole-file IMPORTS_FROM fan-out. See: #737
+    "typescript": [
+        "class_declaration", "class",
+        "interface_declaration", "type_alias_declaration", "enum_declaration",
+    ],
+    "tsx": [
+        "class_declaration", "class",
+        "interface_declaration", "type_alias_declaration", "enum_declaration",
+    ],
     "go": ["type_declaration"],
     # impl_item is a scope for methods, not a second type definition. It is
     # dispatched separately so repeated impl blocks cannot overwrite structs.
@@ -873,6 +884,20 @@ _CLASS_TYPES: dict[str, list[str]] = {
     # _extract_hcl_constructs.
     "hcl": [],
 }
+
+# TS/TSX heritage clauses. Classes wrap theirs in class_heritage; interfaces use
+# extends_type_clause. A type_identifier inside one is already covered by an
+# INHERITS edge, so it must not also emit REFERENCES.
+_TS_HERITAGE_CLAUSES = frozenset({
+    "extends_clause", "implements_clause", "extends_type_clause",
+})
+
+# TS/TSX declarations whose ``name`` field is a type_identifier. That occurrence
+# is the definition site, not a use of the type.
+_TS_TYPE_DECLARATIONS = frozenset({
+    "class_declaration", "abstract_class_declaration", "interface_declaration",
+    "type_alias_declaration", "enum_declaration",
+})
 
 _FUNCTION_TYPES: dict[str, list[str]] = {
     "python": ["function_definition"],
@@ -5744,6 +5769,14 @@ class CodeParser:
                 and node_type in ("jsx_opening_element", "jsx_self_closing_element")
             ):
                 self._extract_jsx_component_call(
+                    child, language, file_path, edges,
+                    enclosing_class, enclosing_func,
+                    import_map, defined_names,
+                )
+
+            # --- TS type-position references ---
+            if language in ("typescript", "tsx") and node_type == "type_identifier":
+                self._extract_ts_type_reference(
                     child, language, file_path, edges,
                     enclosing_class, enclosing_func,
                     import_map, defined_names,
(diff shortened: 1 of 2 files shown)


tirth8205:
Thanks for the contribution. I replayed it on current main and ran 371 affected tests (1 skipped), Ruff, diff checks, and a full graph build; those passed. I then added three end-to-end regressions for the linked issue, and all three expose blockers: (1) callers_of(src/types.ts::Finding) still returns zero because the public query reads CALLS while this PR emits REFERENCES; (2) generic extends/implements emits imported Box twice, as both INHERITS and REFERENCES; and (3) import type { Finding as ImportedFinding } resolves to a nonexistent types.ts::ImportedFinding node instead of the exported types.ts::Finding symbol. Please add public type-user query behavior (or a references_to API and update the issue contract), preserve original symbols through import aliases, and prevent the duplicate generic-heritage reference. Keeping this open and unmerged until those end-to-end cases pass.

tirth8205:
Thank you, @nhatvu148. This contribution is now corrected, integrated, and merged through #771 at `cbb74ed7`.

Your original TypeScript type-node/reference patch and author attribution are preserved as commit `f28862e`. The maintainer follow-up added alias-origin preservation, generic-heritage deduplication, the public `references_to` query, its end-to-end persisted-graph regression, and the CI type-check correction.

Validation before merge:

- 604 combined local tests passed, 1 skipped, on the exact previous main.
- Python 3.10–3.13, Windows daemon/file-handle tests, lint, type-check, security, schema sync, review, CodeQL, and GitGuardian all passed on the exact merged head.

PR #739, opened 2026-07-25
feat(parser): extract dbt model refs as IMPORTS_FROM edges

adamthuvesen:
### Summary
- SQL files containing `{{ ref() }}` / `{{ source() }}` are dbt models: they carry no DDL (dbt wraps the SELECT at build time), so the tree-sitter DDL walk found no nodes, and the FROM/JOIN regex recorded CTE names as phantom `IMPORTS_FROM` targets while missing every real dependency.
- Detect dbt models by content, the same approach as the Ansible sniff, and extract them directly: one `Class` node per model named after the file stem with `extra["sql_kind"]="dbt_model"`, `{{ ref('m') }}` and `{{ ref('pkg', 'm') }}` become `IMPORTS_FROM` edges targeting `m`, and `{{ source('src', 'tbl') }}` becomes an edge targeting `src.tbl`, kept qualified so external source tables never collide with model names.
- Skip the DDL and FROM/JOIN passes for these files. Plain SQL files are unaffected.
- On a real 1,318-model dbt project this goes from 0 nodes and 4,927 phantom edges to 1,218 model nodes and 1,943 real dependency edges.

### Notes
- Known limitation: a dbt model with no `ref()` or `source()` calls (for example, one selecting only from hardcoded tables) is not detected and falls through to the plain-SQL passes. That accounts for the 100-model gap in the numbers above and is inherent to content sniffing.
- Model names resolve by bare name because dbt enforces project-wide model-name uniqueness.

### Verification
- Command: `pytest tests/` → 2071 passed, 5 skipped; includes 7 new tests in `tests/test_dbt_parser.py` covering ref/source extraction, package-qualified refs, dedup, CTE phantom-edge suppression, plain-SQL non-regression, and an end-to-end `full_build` cross-file resolution test
- Command: `ruff check` → clean
- Command: `mypy code_review_graph/parser.py` → clean

### DIFF
--- a/code_review_graph/parser.py
+++ b/code_review_graph/parser.py
@@ -56,6 +56,16 @@ _SQL_TABLE_RE = re.compile(
     re.IGNORECASE,
 )
 
+# dbt model dependencies: {{ ref('model') }}, {{ ref('package', 'model') }}
+# and {{ source('source_name', 'table') }}. String-literal arguments only —
+# dynamic ref() calls cannot be resolved statically.
+_DBT_REF_RE = re.compile(
+    r"\{\{-?\s*(ref|source)\s*\(\s*"
+    r"['\"]([^'\"]+)['\"]"
+    r"(?:\s*,\s*['\"]([^'\"]+)['\"])?"
+    r"\s*\)",
+)
+
 _PYTHON_STAR_CACHE_MAX = 15_000
 _PYTHON_STAR_EXPORT_CACHE: dict[tuple[str, int, int], dict[str, str]] = {}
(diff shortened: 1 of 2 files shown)


tirth8205:
Validated and incorporated through #763, merged as 809c744. Your contributor commit is preserved in the integration history. The corrected version keeps package identity for two-argument ref(), indexes dbt models without dependencies using dbt_project.yml model paths, and avoids classifying unrelated SQL as dbt. Three end-to-end regressions, 724 affected tests, the full 2,205-test suite, Ruff, diff checks, CodeQL, Python 3.10–3.13, and Windows all passed. Closing this original PR because its corrected form is now on main.

PR #747, opened 2026-07-26
fix(parser): detect PHPUnit test-prefixed methods

### LINKED ISSUE #745: [Bug]: PHP test methods are never promoted to Test nodes, so TESTED_BY and tests_for are always empty for PHPUnit
### code-review-graph version

2.3.7

### Operating system

macOS 15 (arm64)

### Python version

Python 3.14.5

### AI platform

Claude Code

```
Nodes: 6
Edges: 10
Files: 2
Languages: php
Last updated: 2026-07-25T22:48:06
Built on branch: main
Built at commit: 53b8dd0a5160
```

### Steps to reproduce

Two files in a fresh git repo:

`src/Calculator.php`
```php
<?php
namespace App;

class Calculator
{
    public function add(int $a, int $b): int
    {
        return $a + $b;
    }
}
```

`tests/CalculatorTest.php`
```php
<?php
namespace App\Tests;

use App\Calculator;
use PHPUnit\Framework\TestCase;

class CalculatorTest extends TestCase
{
    public function testItAddsTwoNumbers(): void
    {
        $calc = new Calculator();
        $this->assertSame(3, $calc->add(1, 2));
    }
}
```

Then:

```
code-review-graph query tests_for src/Calculator.php
```

`tests_for src/Calculator.php` returns `status: ok` with `result_count: 0`. I expected it to point at `CalculatorTest::testItAddsTwoNumbers`.

Looking in the DB, the parse itself is fine — everything needed is there:

```
Class     is_test=0  Calculator
Class     is_test=0  CalculatorTest
File      is_test=0  Calculator.php
File      is_test=1  CalculatorTest.php   <- test file correctly detected
Function  is_test=0  add
Function  is_test=0  testItAddsTwoNumbers <- parsed, but not marked as a test
```

`Test` nodes: 0. `TESTED_BY` edges: 0. `CALLS` edges: 3.

So the test file is recognised (`is_test=1` on the File node) and the test method is parsed, but the method is stored as a plain `Function` with `is_test=0` and never promoted to a `Test` node. Since `TESTED_BY` appears to be derived from calls resolved inside `Test` node bodies, no `Test` nodes means `TESTED_BY` can never form for PHP.

For contrast, the equivalent TS/vitest repo works. Same two-file shape with a `describe`/`it` test gives 2-3 `Test` nodes (named `it:<name>@L<line>`) and 3+ `TESTED_BY` edges. That suggests `Test` node promotion is keyed on `describe`/`it` **call expressions**, which a PHPUnit test never matches — it's a class with `test`-prefixed methods (or `#[Test]`-attributed ones).

At real scale it's the same picture. On a Symfony/PHPUnit backend of ~10.5k nodes: 640 test files correctly flagged `is_test=1`, 55,375 `CALLS` edges, and **0 `Test` nodes / 0 `TESTED_BY`**. A comparable TypeScript/vitest repo gives 186 test files -> 2,306 `Test` nodes -> 15,651 `TESTED_BY`.

This matters beyond `tests_for`, because anything derived from test coverage silently produces confident false output rather than nothing. `detect-changes --brief` on that PHP repo reported 77 test gaps and `risk 1.00`, naming classes whose unit tests were sitting in the same commit — there is no way for a consumer to tell that apart from a genuine finding.

### Additional context

The fix looks cheap because no new parsing is needed — the methods are already in the graph and the files are already flagged. Promoting a method to a `Test` node when it lives in an `is_test=1` PHP file and either its name matches `^test` or it carries the `#[Test]` attribute would be enough for `TESTED_BY` to populate off the existing call edges.

Worth noting PHPUnit supports both conventions (the `test` prefix, and `#[PHPUnit\Framework\Attributes\Test]` since 10/11), and the older `@test` docblock annotation. Behat is a separate shape again (context classes with `@Given`/`@When`/`@Then`), if that's of interest.

Possibly related to #737, which is also about PHP-adjacent nodes not being indexed the way other languages are.

merlincat11:
# Pull Request

## Linked issue

Closes #745

## What & why

PHPUnit treats class methods whose names start with `test` as tests, including camelCase names such as `testItAddsTwoNumbers`. The parser's shared name patterns only recognized the narrower `test_` prefix, so these methods were stored as ordinary `Function` nodes even when their PHP file was already recognized as a test file.

This change promotes `test*` PHP method declarations to `Test` nodes when they are in recognized test files. It keeps the behavior scoped to PHPUnit-style methods, preserves the existing shared `test_` behavior, and leaves PHP import resolution and graph query behavior unchanged.

A regression test covers both the camelCase PHPUnit method and a top-level camelCase PHP function that must remain a normal `Function`.

## How it was tested

```bash
env -u CRG_DATA_DIR uv run pytest tests/test_multilang.py::TestPHPTestAnnotations -q
env -u CRG_DATA_DIR uv run pytest tests/ --tb=short -q
uv run --with mypy mypy code_review_graph/ --ignore-missing-imports --no-strict-optional
```

Results:

- Focused PHP test classification: 12 passed
- Full suite: 2,077 passed, 5 skipped, 2 xpassed
- Ruff: all checks passed
- Mypy: no issues found in 66 source files

## Checklist

- [x] All tests pass: `env -u CRG_DATA_DIR uv run pytest tests/ --tb=short -q`
- [x] Type checking passes: `uv run --with mypy mypy code_review_graph/ --ignore-missing-imports --no-strict-optional`
- [x] Docs updated where behavior changed (README, `docs/`, docstrings) — not applicable; this is an internal parser-classification fix covered by regression tests

### DIFF
--- a/code_review_graph/parser.py
+++ b/code_review_graph/parser.py
@@ -9602,6 +9602,9 @@ class CodeParser:
             decorators = tuple(deco_list)
 
         is_test = _is_test_function(name, file_path, decorators)
+        # PHPUnit's name convention is ``test*`` (not only ``test_*``).
+        if language == "php" and _is_test_file(file_path) and name.startswith("test"):
+            is_test = True
         kind = "Test" if is_test else "Function"
 
         parent_name = enclosing_class
--- a/code_review_graph/parser.py
+++ b/code_review_graph/parser.py
@@ -9603,7 +9603,12 @@ class CodeParser:
 
         is_test = _is_test_function(name, file_path, decorators)
         # PHPUnit's name convention is ``test*`` (not only ``test_*``).
-        if language == "php" and _is_test_file(file_path) and name.startswith("test"):
+        if (
+            language == "php"
+            and child.type == "method_declaration"
+            and _is_test_file(file_path)
+            and name.startswith("test")
+        ):
             is_test = True
         kind = "Test" if is_test else "Function"
 
--- a/tests/test_multilang.py
+++ b/tests/test_multilang.py
@@ -785,7 +785,8 @@ class TestPHPTestAnnotations:
     """
 
     def _parse(self, tmp_path):
-        p = tmp_path / "ExampleTest.php"
+        p = tmp_path / "tests" / "ExampleTest.php"
+        p.parent.mkdir()
         p.write_text(
             "<?php\n"
             "namespace Tests;\n\n"
@@ -797,6 +798,9 @@ class TestPHPTestAnnotations:
             "class ExampleTest extends TestCase\n"
             "{\n"
             "    public function test_prefixed_method_should_be_detected(): void\n"
+            "    {\n"
+            "    }\n\n"
+            "    public function testItAddsTwoNumbers(): void\n"
             "    {\n"
             "    }\n\n"
             "    /** @test */\n"
@@ -842,6 +846,12 @@ class TestPHPTestAnnotations:
     def test_name_prefix_still_detected(self, tmp_path):
         nodes, _ = self._parse(tmp_path)
         m = next(n for n in nodes if n.name == "test_prefixed_method_should_be_detected")
+        assert m.kind == "Test"
+        assert m.is_test is True
+
+    def test_phpunit_camel_case_name_prefix_detected(self, tmp_path):
+        nodes, _ = self._parse(tmp_path)
+        m = next(n for n in nodes if n.name == "testItAddsTwoNumbers")
         assert m.kind == "Test"
         assert m.is_test is True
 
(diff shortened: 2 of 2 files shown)


tirth8205:
Thank you — the two contributor commits from this PR were preserved with attribution in the corrected integration #764, now merged as 9fed471ccb4071c806c2a1f61cabbce526190a17.

The integration keeps the PHPUnit test-method detection and adds the missing end-to-end resolution for constructed instance calls, import aliases, reassignment invalidation, canonical CALLS/TESTED_BY edges, and public tests_for queries. It passed the focused PHP/graph/multi-language suites, real parser-process parity, and the complete hosted matrix including Windows. I am closing this original PR only because #764 fully incorporates and completes it.

PR #749, opened 2026-07-26
fix(graph): resolve Python imports from indexed files

### LINKED ISSUE #720: tests_for misses non-name-matching test callers in src/ package layouts (callers_of finds them)
## Summary

`query tests_for` misses test callers whose names don't reference the target function, but only in `src/` package layouts. `callers_of` finds all of them correctly in the same graph. In a flat layout, `tests_for` behaves correctly, which made this fun to minimize.

Version: `code-review-graph==2.3.7` (pip; MCP serverInfo reports internal component 3.4.4). Observed on Windows 11 / Python 3.12; the trigger appears to be layout, not OS.

## Minimal reproduction

```
repro/
  src/mypkg/__init__.py        (empty)
  src/mypkg/runner.py
  tests/test_runner.py
```

`src/mypkg/runner.py`:

```python
def render_thing(code: str) -> str:
    return code.upper()
```

`tests/test_runner.py`:

```python
from mypkg.runner import render_thing


def test_render_thing_basic():
    assert render_thing("a") == "A"


def test_pipeline_uses_uppercase():
    assert render_thing("bc") == "BC"
```

```
git init
code-review-graph build
code-review-graph query tests_for  "src/mypkg/runner.py::render_thing"
code-review-graph query callers_of "src/mypkg/runner.py::render_thing"
```

## Observed

- `tests_for` -> `"result_count": 1` - only `test_render_thing_basic` (the test whose *name* references the function)
- `callers_of` -> `"result_count": 2` - both tests, with CALLS edges at confidence 1.0 EXTRACTED

Expected: `tests_for` returns both tests - `test_pipeline_uses_uppercase` calls the target directly.

## Control (does NOT reproduce)

The identical two tests against a flat-layout module (`frob.py` at repo root, `from frob import frobnicate`) give `tests_for` -> 2 results, matching `callers_of`. Adding a same-named function in a second module (ambiguity) also does not degrade the flat case.

## Real-world impact

On a real `src/`-layout codebase (~150 files, ~27k edges), a central function with 20 direct test callers: `callers_of` found all 20 (verified against a manual grep of every call site - zero false positives, including a module-attribute call), while `tests_for` returned 1 - again the single test whose name contains the function name. So in src layouts, `tests_for` effectively degrades to name matching, silently under-reporting by whatever the naming convention hides.

## Workaround

Derive dependent tests from `callers_of` filtered to `"is_test": true` - that path is accurate in every case I tested. Happy to provide more detail or test a fix.

merlincat11:
## Summary

- add a dedicated post-index Python resolver that runs after full builds, incremental `.py` changes, and watched Python batches
- resolve raw Python modules by unique repository-wide suffix before generic bare `CALLS` / `TESTED_BY` qualification
- preserve raw module provenance and reversible unique ↔ ambiguous resolution with bounded candidate metadata
- mark bare caller matches and naming-only test matches as inferred rather than exact graph evidence

## Root cause

Python import resolution only searched the importing file directory and its ancestors. In `src/` layouts, imports such as `from mypkg.runner import render_thing` therefore remained raw module imports. `CALLS` and `TESTED_BY` endpoints stayed bare, so the evidence-gated `tests_for` path rejected valid direct test callers while `callers_of` found them through its bare-name fallback.

## Architecture

The repository-wide lookup now follows the existing language-resolver lifecycle:

1. parse all relevant files
2. resolve Python `IMPORTS_FROM` targets in `python_resolver.py`
3. resolve generic bare `CALLS` and `TESTED_BY` endpoints
4. compute signatures, search, flows, and communities

The Python resolver runs after every full build and after incremental additions, changes, or removals of `.py` files. Watch mode applies the same ordering once per debounced batch, including deletion batches. Generic graph resolution remains language-neutral; no universal resolver framework or Jedi dependency was added.

## Impact

`tests_for` now reports direct callers correctly for unique repository-local modules in `src/`, `lib/`, and nested monorepo layouts. Duplicate module suffixes remain explicitly ambiguous, and full, incremental, and watched changes can move safely between unique and ambiguous states in either direction.

Fixes #720.

## Validation

- `uv run pytest tests/ --tb=short -q` — 2078 passed, 5 skipped, 2 xpassed
- `uv run ruff check code_review_graph/` — passed
- `uv run --with mypy mypy code_review_graph/ --ignore-missing-imports --no-strict-optional` — no issues in 67 source files

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,9 +4,9 @@
 
 ### Fixed
 
-- Resolved repository-local Python imports by unique module suffix after indexing,
-  so `src/` layouts produce canonical `CALLS` and `TESTED_BY` edges while
-  duplicate package candidates remain explicitly unresolved (#720).
+- Added a post-index Python import resolver using unique module suffixes, so
+  `src/` layouts produce canonical `CALLS` and `TESTED_BY` edges while duplicate
+  package candidates remain explicitly unresolved (#720).
 
 ## [2.3.7] - 2026-07-18
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 ## [Unreleased]
 
+### Fixed
+
+- Resolved repository-local Python imports by unique module suffix after indexing,
+  so `src/` layouts produce canonical `CALLS` and `TESTED_BY` edges while
+  duplicate package candidates remain explicitly unresolved (#720).
+
 ## [2.3.7] - 2026-07-18
 
 **Maintainer-reconciliation release.** This release packages the verified work
--- a/code_review_graph/graph.py
+++ b/code_review_graph/graph.py
@@ -608,104 +608,6 @@ class GraphStore:
         ]
         return supported[0] if len(supported) == 1 else None
 
-    def _resolve_python_import_targets(self) -> int:
-        """Resolve raw Python module imports against indexed repository files.
-
-        A module such as ``mypkg.runner`` may live below any source root
-        (``src/``, ``lib/``, or a monorepo package). Match by module suffix
-        after every file has been indexed, and only resolve a unique match.
-        """
-        conn = self._conn
-        python_files = [
-            row["file_path"]
-            for row in conn.execute(
-                "SELECT file_path FROM nodes "
-                "WHERE kind = 'File' AND language = 'python'"
-            ).fetchall()
-        ]
-        if not python_files:
-            return 0
-
-        modules: dict[str, set[str]] = {}
-        for file_path in python_files:
-            parts = [
-                part for part in file_path.replace("\\", "/").split("/")
-                if part
-            ]
-            if not parts:
-                continue
-            filename = parts[-1]
-            if filename == "__init__.py":
-                components = parts[:-1]
-            elif filename.endswith(".py"):
-                components = [*parts[:-1], filename[:-3]]
-            else:
-                continue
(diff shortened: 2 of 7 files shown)


tirth8205:
Thanks for the contribution. I replayed it on current main. 277 affected tests passed (1 skipped, process parity run separately and passed), and Ruff/diff checks passed. New positive end-to-end cases also pass for src/, lib/, monorepo, and Windows paths. One safety blocker remains: when two mypkg/runner.py candidates exist from the initial full build, the import is marked ambiguous but the related CALL/TESTED_BY endpoints are not. The public bare-name fallback then reports the same test as a caller of both production functions. The existing ambiguity test only adds the duplicate after an initially unique build, so it misses this case. Please propagate initial import ambiguity to CALL/TESTED_BY endpoints, or make callers_of reject the bare fallback for an ambiguous import, and add a full-build-in-ambiguous-state regression. Keeping this open and unmerged until that false attribution is removed.

tirth8205:
Validated and incorporated through #762, merged as f9dc6d3. All four MerlinH contributor commits are preserved in the integration history. The corrected version also prevents false attribution when duplicate module suffixes exist from the initial build and preserves the new serialized watch lifecycle. Validation passed 309 affected tests, real process-pool parity, positive src/lib/monorepo/Windows-path cases, Ruff, diff checks, CodeQL, Python 3.10–3.13, and Windows. Closing this original PR because its corrected form is now on main.

PR #767, opened 2026-07-27
fix(graph): make impact traversal directional (#291)

### LINKED ISSUE #291: [bug] Impact radius analysis overestimates affected files for Vue component changes
### Problem Description

When analyzing the impact radius of a Vue single-file component change, the tool reports an overly broad scope.

**Scenario:**
- Modified file: `CustodianAccount.vue` (a child component)
- Expected affected files: Only the parent component that directly imports it (`BankAccount.vue`), plus maybe router config (~1-3 files)

**Actual output:**
- Direct impact: 10 nodes ✅
- Indirect impact: 143 nodes ❌
- Affected files: 83 files ❌

The tool seems to be propagating dependencies in the wrong direction. Modifying a child component should NOT affect unrelated utility components (AmPagination, AmButton), service layer files (api.service.ts), or architecture layer files (App.vue, main.ts).

### Expected vs Actual

| | Expected | Actual |
|---|---|---|
| Affected files | 1-3 | 83 |
| Indirect nodes | 0-5 | 143 |

### Additional Context
- Framework: Vue 3 with TypeScript
- File type: `.vue` single-file component
- Tool version: v2.3.2

### Suggestion
The "blast radius" analysis should distinguish between:
- **Callers** (files that import this component) → should be included ✅
- **Callees** (files this component imports) → should NOT be included ❌

Currently it seems to be counting both directions.

tirth8205:
## Summary

Fixes #291 by making impact-radius traversal follow dependency direction instead of walking every edge both ways.

- dependency-shaped edges propagate from the changed dependency to its dependents;
- `TESTED_BY` propagates from production code to tests;
- `CONTAINS` does not expand impact because changed-file seeding already includes contained nodes;
- unknown edge kinds use the documented dependency-shaped default;
- SQL and NetworkX share the same policy and preserve strongest parallel-edge weights in each permitted direction.

## Why the previous behavior was wrong

For an edge `importer -> imported`, changing the importer incorrectly reported the imported dependency as impacted. The same edge weight was used in both directions, so genuine importers and false dependency results were indistinguishable.

## Validation

- New direction regressions: passed.
- Real three-file full build: changing the middle module includes its importer and excludes its own imported dependency in both engines.
- Broad impact-related local suite: 318 passed, 1 skipped.
- Agent practical graph/tool suite before the final end-to-end addition: 359 passed, 1 skipped.
- Ruff: passed.
- Python compile check: passed.
- `git diff --check`: passed.

Full hosted Python, Windows, type, lint, schema, security, and CodeQL checks are still required before merge.

### DIFF
--- a/code_review_graph/constants.py
+++ b/code_review_graph/constants.py
@@ -66,6 +66,31 @@ IMPACT_EDGE_WEIGHTS: dict[str, float] = {
     "CONTAINS": 0.3,
 }
 IMPACT_DEFAULT_EDGE_WEIGHT = 0.5
+
+# Stored dependency edges point from the dependent to its dependency, so impact
+# normally propagates against the stored edge (target -> source). TESTED_BY is
+# intentionally stored in the opposite orientation (production -> test).
+# CONTAINS is not traversed: changing a file already seeds every node in it, and
+# following containment can bridge into unrelated structure through stale edges.
+IMPACT_DIRECTION_INCOMING = "incoming"
+IMPACT_DIRECTION_OUTGOING = "outgoing"
+IMPACT_DIRECTION_NONE = "none"
+IMPACT_EDGE_DIRECTIONS: dict[str, str] = {
+    "CALLS": IMPACT_DIRECTION_INCOMING,
+    "INHERITS": IMPACT_DIRECTION_INCOMING,
+    "OVERRIDES": IMPACT_DIRECTION_INCOMING,
+    "IMPLEMENTS": IMPACT_DIRECTION_INCOMING,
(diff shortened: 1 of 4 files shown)


tirth8205:
Integrated and merged through #770 at `329dbb91`.

This branch's directional impact fix and real import-direction regression are both present in the merged history. The standalone #767 matrix passed, then the exact combined #770 state passed Python 3.10–3.13, Windows, lint, type-check, security, schema sync, review, CodeQL, and GitGuardian before merge.

Closing this PR only because #770 superseded it.

PR #769, opened 2026-07-27
fix(update): anchor incremental work to last-synced commit (#694)

tirth8205:
## Summary

Corrected integration of contributor PR #694, preserving Utku Özdemir’s patch and attribution.

The default incremental update base is now the commit at which the graph was last successfully built, instead of a fixed `HEAD~1`. This lets one update reconcile every commit since the graph was last in sync, including branch switches and multi-commit pulls.

If a fresh/legacy graph has no usable stored commit, or the commit was removed by a rewrite/shallow clone, the command performs a full rebuild rather than silently diffing from the wrong base. An explicit `--base` still overrides the automatic choice.

## Additional validation

The original contribution had good broad coverage but no exact regression combining the risky paths. The maintainer test splits these changes across three commits:

1. rename `pkg/service.py` to `pkg/core.py`;
2. update an existing importer;
3. add a new importing file.

It proves:

- `HEAD~1` sees only the newest file;
- the default stored-SHA update sees all four old/new paths;
- old nodes and edges are purged;
- complete node and edge snapshots match a fresh full rebuild.

## Results

- Exact multi-commit rename regression: passed.
- All 244 unique affected CLI/incremental/Git/forget/uninstall tests passed after rebasing on #766.
- The one process-pool test blocked by the local sandbox passed with a real two-worker process outside that sandbox.
- Production Ruff: passed.
- Affected test lint: passed.
- `git diff --check`: passed.
- Contributor patch and regression patch IDs are unchanged after rebase.

Full hosted Python, Windows, type, lint, schema, security, and CodeQL checks are required before merge.

### DIFF
--- a/code_review_graph/cli.py
+++ b/code_review_graph/cli.py
@@ -755,7 +755,11 @@ def main() -> None:
 
     # update
     update_cmd = sub.add_parser("update", help="Incremental update (only changed files)")
-    update_cmd.add_argument("--base", default="HEAD~1", help="Git diff base (default: HEAD~1)")
+    update_cmd.add_argument(
+        "--base",
+        default=None,
+        help="Git diff base (default: the commit the graph was last built at)",
+    )
     update_cmd.add_argument("--repo", default=None, help="Repository root (auto-detected)")
     update_cmd.add_argument("-q", "--quiet", action="store_true", help="Suppress output")
     update_cmd.add_argument(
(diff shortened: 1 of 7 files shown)


tirth8205:
Integrated and merged through #770 at `329dbb91`.

utkuozdemir's original #694 patch and authorship are preserved in the merged history. The integration also includes the exact multi-commit rename/import/new-file parity regression.

The standalone #769 matrix passed, then the exact combined #770 state passed Python 3.10–3.13, Windows, lint, type-check, security, schema sync, review, CodeQL, and GitGuardian before merge. Closing this PR only because #770 superseded it.

PR #775, opened 2026-07-28
fix(parser): index C++ and Qt headers correctly

### LINKED ISSUE #463: C++ class method definitions not parsed as Function nodes (scoped `Class::method()` and Qt macros)
## Summary

After a full rebuild with v2.3.2 (which includes the PR #371 scoped-method-name fix), Function nodes
are still not created for C++ class method definitions. This affects both plain C++ and Qt-based projects.
Additionally, Qt macros (`Q_OBJECT`, `QT_BEGIN_NAMESPACE`) are misidentified as Function nodes.

Two distinct bugs are confirmed by running `CodeParser().parse_file()` directly.

---

## Environment

- code-review-graph version: **2.3.2**
- OS: Windows 11 (26200)
- Python: 3.12.10

---

### Files used

**MyWidgetPlain.h** — pure C++, no Qt macros:

```cpp
#pragma once

class MyWidgetPlain {
 public:
  MyWidgetPlain();
  ~MyWidgetPlain();
  void doSomething();
  int calculateValue(int a, int b);

 protected:
  void onButtonClicked();
  void onDataReceived(int value);
  void onReset();
};
```

**MyWidgetPlain.cpp**:

```cpp
#include "MyWidgetPlain.h"

MyWidgetPlain::MyWidgetPlain() {}
MyWidgetPlain::~MyWidgetPlain() {}

void MyWidgetPlain::doSomething() { onReset(); }
int MyWidgetPlain::calculateValue(int a, int b) { return a + b; }
void MyWidgetPlain::onButtonClicked() { int result = calculateValue(1, 2); }
void MyWidgetPlain::onDataReceived(int value) { if (value < 0) return; doSomething(); }
void MyWidgetPlain::onReset() {}
```

**MyWidget.h** — Qt-based:

```cpp
#pragma once
#include <QMainWindow>

QT_BEGIN_NAMESPACE namespace Ui { class MyWidgetClass; };
QT_END_NAMESPACE

class MyWidget : public QMainWindow {
  Q_OBJECT

 public:
  MyWidget(QWidget* parent = nullptr);
  ~MyWidget();
  void doSomething();
  int calculateValue(int a, int b);

 protected Q_SLOTS:
  void onButtonClicked();
  void onDataReceived(int value);

 public Q_SLOTS:
  void onReset();

 Q_SIGNALS:
  void dataReady(int result);
  void errorOccurred(const QString& msg);
};
```

**MyWidget.cpp**:

```cpp
#include "MyWidget.h"

MyWidget::MyWidget(QWidget* parent) : QMainWindow(parent) {}
MyWidget::~MyWidget() {}
void MyWidget::doSomething() { onReset(); }
int MyWidget::calculateValue(int a, int b) { return a + b; }
void MyWidget::onButtonClicked() { Q_EMIT dataReady(calculateValue(1, 2)); }
void MyWidget::onDataReceived(int value) { if (value < 0) { Q_EMIT errorOccurred("err"); return; } doSomething(); }
void MyWidget::onReset() { Q_EMIT dataReady(0); }
```

### Test script

```python
import sys
sys.path.insert(0, r'<path-to-crg-site-packages>')

from pathlib import Path

for fname in ['MyWidgetPlain.cpp', 'MyWidgetPlain.h', 'MyWidget.cpp', 'MyWidget.h']:
    nodes, edges = CodeParser().parse_file(Path(fname))
    print(f'\n=== {fname} -> {len(nodes)} nodes, {len(edges)} edges ===')
    for n in nodes:
        print(f'  [{n.kind:10}] {n.name}')
```

---

## Actual output

```
=== MyWidgetPlain.cpp -> 1 nodes, 1 edges ===
  [File      ] MyWidgetPlain.cpp

=== MyWidgetPlain.h -> 2 nodes, 3 edges ===
  [File      ] MyWidgetPlain.h
  [Function  ] class

=== MyWidget.cpp -> 1 nodes, 1 edges ===
  [File      ] MyWidget.cpp

=== MyWidget.h -> 2 nodes, 2 edges ===
  [File      ] MyWidget.h
  [Function  ] QT_BEGIN_NAMESPACE
```

---

## Expected output

```
=== MyWidgetPlain.cpp -> 8 nodes ===
  [File      ] MyWidgetPlain.cpp
  [Function  ] MyWidgetPlain       (constructor)
  [Function  ] ~MyWidgetPlain      (destructor)
  [Function  ] doSomething
  [Function  ] calculateValue
  [Function  ] onButtonClicked
  [Function  ] onDataReceived
  [Function  ] onReset

=== MyWidgetPlain.h -> 9 nodes ===
  [File      ] MyWidgetPlain.h
  [Class     ] MyWidgetPlain
  [Function  ] MyWidgetPlain
  ... (all declared methods)

=== MyWidget.cpp -> 8 nodes ===
  [File      ] MyWidget.cpp
  [Function  ] MyWidget
  [Function  ] ~MyWidget
  [Function  ] doSomething
  ... (all defined methods)

=== MyWidget.h -> 10 nodes ===
  [File      ] MyWidget.h
  [Class     ] MyWidget
  [Function  ] MyWidget
  ... (all declared methods and signals)
```

---

### Bug 1 — `.cpp` scoped method definitions produce no Function nodes

`MyWidgetPlain.cpp` has **zero Qt dependencies**, but none of its 7 method definitions
(`MyWidgetPlain::MyWidgetPlain`, `void MyWidgetPlain::doSomething`, etc.) are emitted as nodes.
This means the fix in PR #371 (`qualified_identifier` / `destructor_name` / `operator_name` handling)
is either not applied in this version or does not cover this case.

Note: PR #371 was merged 2026-05-07 and v2.3.2 was released after that date.
The scoped method definitions in `.cpp` files are still not parsed.

### Bug 2 — Wrong node emitted for `.h` class definition

`MyWidgetPlain.h` emits a `Function` node named `"class"` instead of a `Class` node named
`"MyWidgetPlain"`. The parser appears to be picking up the `class` keyword token as the node name.

### Bug 3 — Qt macro `QT_BEGIN_NAMESPACE` emitted as a `Function` node

`MyWidget.h` emits a `Function` node named `"QT_BEGIN_NAMESPACE"`.
The following pattern confuses the parser:

```cpp
QT_BEGIN_NAMESPACE namespace Ui { class MyWidgetClass; };
QT_END_NAMESPACE
```

Tree-sitter cannot expand macros, so it treats `QT_BEGIN_NAMESPACE` as a function-like token.

### Bug 4 — Qt class body macros prevent all member parsing

`Q_OBJECT`, `Q_SIGNALS:`, and `protected Q_SLOTS:` inside the class body cause the Tree-sitter
AST for the class body to be malformed. As a result, zero member Function nodes are emitted for
`MyWidget.cpp` — same outcome as `MyWidgetPlain.cpp`, suggesting Bug 1 is the primary cause and
Bug 4 adds further breakage for Qt projects specifically.

---

## Impact

- Any C++ project using out-of-line method definitions (`Class::method()` in `.cpp`) gets zero
  Function nodes — which is the standard pattern for all non-trivial C++ codebases.
- Qt projects (QObject-derived classes) are additionally affected by macro misidentification.
- `detect-changes` always returns `changed_functions: []` and `risk_score: 0.0` for C++ projects,
  making it effectively unusable for C++.

---

## Related

- PR #371 — Fix C++ method name extraction for scoped/destructor/operator definitions (merged 2026-05-07)
  This PR targeted the same area but the fix does not appear to be effective for standard
  out-of-line method definitions.

HsiangNianian:
## Problem

`.h` files currently default to the C parser, so normal C++ class headers lose their classes and member declarations. Qt's structural macros can also push the C++ grammar into error recovery, hiding the real class and surfacing macro names as functions.

## What changed

- promote ambiguous `.h` files only when the C++ tree contains strong language evidence, while keeping ordinary C and C23 headers on the C path
- shield the Qt structural macros from #463 with same-length replacements limited to real code, so comments, literals, directives, and source positions stay intact
- index callable C++ declarations without treating fields, local prototypes, or function-pointer variables as functions
- add the exact four-file plain/Qt fixture as an end-to-end `full_build` regression

## Validation

- `uv run pytest tests/test_cpp_qt_headers.py tests/test_parser.py tests/test_multilang.py tests/test_cpp_overload_identity.py -q` (`627 passed`)
- `uv run pytest tests/ --tb=short -q --cov=code_review_graph --cov-report=term-missing --cov-fail-under=65` (`2319 passed`, `82.23%` coverage)
- `uv run ruff check code_review_graph/ tests/test_cpp_qt_headers.py`
- `uv run --with bandit bandit -r code_review_graph/ -c pyproject.toml -q`

Closes #463

### DIFF
--- a/code_review_graph/parser.py
+++ b/code_review_graph/parser.py
@@ -752,6 +752,21 @@ EXTENSION_TO_LANGUAGE: dict[str, str] = {
     ".yaml": "yaml",
 }
 
+# ``.h`` is shared by C and C++. Keep C as the extension default, then promote
+# a header only when the C++ grammar finds syntax that C cannot express. Weak
+# compatibility markers such as ``__cplusplus`` and ``extern "C"`` are
+# deliberately excluded because they are common in otherwise-C headers.
+_CPP_HEADER_EVIDENCE_TYPES = frozenset({
+    "access_specifier",
+    "alias_declaration",
+    "base_class_clause",
+    "class_specifier",
+    "concept_definition",
+    "namespace_definition",
+    "template_declaration",
+    "using_declaration",
+})
+
 # Shebang interpreter → language mapping for extension-less Unix scripts.
 # Each key is the **basename** of the interpreter path as it appears after
 # ``#!`` (or after ``#!/usr/bin/env``).  Only languages already registered
@@ -2522,6 +2537,17 @@ class CodeParser:
         if not language:
             return [], []
 
+        parser = None
+        tree = None
+        if language == "c" and path.suffix.lower() == ".h":
+            cpp_parser = self._get_parser("cpp")
+            if cpp_parser is not None:
+                cpp_tree = cpp_parser.parse(source)
+                if self._has_cpp_header_evidence(cpp_tree.root_node):
+                    language = "cpp"
+                    parser = cpp_parser
+                    tree = cpp_tree
+
         if language == "blade":
             return self._parse_blade(path, source)
 
@@ -2592,11 +2618,13 @@ class CodeParser:
         if language == "yaml":
             return [], []
 
-        parser = self._get_parser(language)
+        if parser is None:
+            parser = self._get_parser(language)
         if not parser:
             return [], []
 
-        tree = parser.parse(source)
+        if tree is None:
+            tree = parser.parse(source)
         nodes: list[NodeInfo] = []
         edges: list[EdgeInfo] = []
         file_path_str = str(path)
@@ -2683,6 +2711,17 @@ class CodeParser:
                     ))
 
         return nodes, edges
+
+    @staticmethod
+    def _has_cpp_header_evidence(root) -> bool:
+        """Return whether a parsed ``.h`` tree contains C++-only syntax."""
+        pending = [root]
+        while pending:
+            node = pending.pop()
+            if node.type in _CPP_HEADER_EVIDENCE_TYPES:
+                return True
+            pending.extend(node.named_children)
+        return False
 
     @classmethod
     def _mask_blade_comments(cls, text: str) -> str:
--- a/code_review_graph/parser.py
+++ b/code_review_graph/parser.py
@@ -762,10 +762,16 @@ _CPP_HEADER_EVIDENCE_TYPES = frozenset({
     "base_class_clause",
     "class_specifier",
     "concept_definition",
+    "lambda_expression",
     "namespace_definition",
+    "noexcept",
+    "placeholder_type_specifier",
     "template_declaration",
+    "trailing_return_type",
     "using_declaration",
 })
+
+_CPP_HEADER_EVIDENCE_QUALIFIERS = frozenset({b"constexpr", b"consteval", b"constinit"})
 
 _CPP_QT_STRUCTURAL_MACRO_REPLACEMENTS = {
     b"QT_BEGIN_NAMESPACE": b" " * len(b"QT_BEGIN_NAMESPACE"),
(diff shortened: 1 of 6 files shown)


tirth8205:
Verified locally before merge: reproduced #463 on main with the exact four-file fixture from the issue (headers parsed as C, false Function nodes from the class keyword and QT_BEGIN_NAMESPACE, no Class or member nodes), confirmed the fix produces the expected Class and member Function nodes for both headers, and revert-checked the new tests (9 of 14 fail without the fix; the 5 that pass are the stays-C negative guards). Full suite: 2319 passed, 5 skipped, 2 xpassed. Ruff and mypy clean, all CI checks green on the updated branch. Known cost, accepted: parser throughput on C/C++ heavy repos regresses roughly 2-3x from the speculative C++ probe and the masking pass; a follow-up issue will track a cheap prefilter (skip masking when no Q_ or QT_ token is present).

PR #779, opened 2026-07-28
fix(skills): align tool names with exported MCP schema

TheJeje20:
Summary:
- Align generated and bundled skill templates with the MCP tool names exported by the server.
- Keep the skills consistent with the current schema for Codex and Claude Code instructions.
- Add a regression test that fails if the old bare names reappear.

Validation:
- .venv/bin/pytest tests/test_skills.py -q
- 192 passed

Notes:
- This fixes stale skill/template references to bare names such as get_minimal_context, list_graph_stats, list_flows, get_flow, and find_large_functions.
- The MCP server already exported the corresponding *_tool names; the bug was in the instructions/templates layer.

### DIFF
--- a/code_review_graph/skills.py
+++ b/code_review_graph/skills.py
@@ -680,20 +680,20 @@ _SKILLS: dict[str, dict[str, str]] = {
             "## Explore Codebase\n\n"
             "Use the code-review-graph MCP tools to explore and understand the codebase.\n\n"
             "### Steps\n\n"
-            "1. Run `list_graph_stats` to see overall codebase metrics.\n"
+            "1. Run `list_graph_stats_tool` to see overall codebase metrics.\n"
             "2. Run `get_architecture_overview_tool` for high-level community structure.\n"
-            "3. Use `list_communities_tool` to find major modules, then `get_community` "
+            "3. Use `list_communities_tool` to find major modules, then `get_community_tool` "
             "for details.\n"
             "4. Use `semantic_search_nodes_tool` to find specific functions or classes.\n"
             "5. Use `query_graph_tool` with patterns like `callers_of`, `callees_of`, "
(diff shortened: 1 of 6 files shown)


tirth8205:
Thanks, the underlying problem is real and verified: the server exports *_tool names (main.py) while the skill templates referenced bare names, and your renames all match the exported schema. Two gaps before this can merge: (1) skills/debug-issue/SKILL.md (lines 14 and 25) and skills/review-changes/SKILL.md (line 27) still contain the old bare names (get_flow, get_minimal_context). You fixed the skills.py templates for those two skills but not the bundled files, so the generated and bundled copies now disagree, and the bundled files are what the sdist and the Qoder installer ship. (2) The new test only checks explore-codebase and refactor-safely, which is why it passes despite the leftover stale names; please extend expected_tools/legacy checks to all four skills so the guard actually covers the claim in the PR description. Optional follow-up (fine as a separate PR): prompts.py MCP prompt templates and hints.py next_tool_suggestions have the same bare-name drift. Full suite passes locally on your branch (2306 passed) and ruff is clean.

TheJeje20:
Thank you for the review. I missed that. I have pushed the suggested changes:

- Updated bundled SKILL.md files, 
- extended the regression test to cover all four skills,
- validated with `pytest tests/test_skills.py -q`

I also noticed that skills.py currently defines only four skills, while there are additional .md files. I will check whether this is expected; if not, I will open a separate pull request.

TheJeje20:
PR for the prompts.py/hints.py: #821 
PR for the additional .md files: #822

tirth8205:
Checked every renamed reference against the exported schema in main.py and the generated and bundled skills now match byte for byte. Merging. #821 carries these same hunks plus the prompts and hints work, so it will need a rebase down to that scope.

PR #782, opened 2026-07-29
fix(cli): prevent missing graphs from poisoning updates

### LINKED ISSUE #777: status and update create an empty graph.db on a repo with no graph, permanently poisoning it
## Summary

Running `status` or `update` on a repository that has **no graph yet** creates an empty `graph.db` (0 nodes) and exits 0. Subsequent `update` calls never repair it — they incrementally add only the files that changed after that moment. The graph then answers queries confidently and wrongly: `callers_of X` returns `not_found` for symbols that are genuinely called.

Read-only commands behave correctly by contrast (`query`, `search`, `dead-code`, `large-functions` exit 1 and create nothing), so the problem is specific to `status` and `update`.

## Reproduce

```bash
mkdir demo && cd demo && git init
# three files: gamma() calls alpha(), etc.
git add -A && git commit -m init

code-review-graph status        # <-- creates graph.db with Nodes: 0, exits 0
code-review-graph status        # Nodes: 0

# now touch one file and update
echo "# x" >> one_of_them.py
code-review-graph update -q

code-review-graph status        # Files: 1  (should be 3)
code-review-graph query callers_of alpha
# -> not_found, although gamma() calls alpha()
```

Measured on a 3-file fixture: `Files: 1` instead of `3`, `callers_of` returns `not_found` for a symbol with a real caller. Repeated `update` runs do not fix it; only `rm -rf .code-review-graph && code-review-graph build` recovers.

## Why it matters

The failure mode is silent and sticky:

- nothing errors, so a wrapper/hook that checks "does `graph.db` exist?" concludes the graph is ready
- it is easy to hit accidentally — `status` is the natural way to ask "is there a graph here?"
- it is especially likely with **git worktrees**, where each worktree gets its own data dir and starts with no graph

We hit this in production usage across three repos and had to work around it by checking node count (`select 1 from nodes limit 1`) instead of file existence before letting any hook call the tool.

## Suggested fix

Either of these would close it:

1. `update` on a missing/empty graph performs a full `build` instead of an incremental pass
2. `status` and `update` do not create the database as a side effect — `status` reports "no graph, run build" and exits non-zero, matching how the read-only commands already behave

Option 2 seems more consistent with the existing command semantics, and would also make "does a usable graph exist here?" answerable without side effects.

## Environment

code-review-graph 2.3.7, Python 3.13, Windows 11.

HsiangNianian:
## What changed

- make `status` resolve missing graphs without creating a database, data directory, or registry entry
- make `update` fall back to a full build when the graph is missing or has zero nodes
- keep the existing default-path legacy migration while isolating external and explicit data-dir lookups
- use a cheap node-existence query instead of collecting full graph statistics on every update
- cover the original three-file failure end to end, including the `callers_of` result

## Why

An empty database could previously look like a valid graph. Later incremental updates would then parse only changed files, leaving the graph incomplete and queries confidently wrong.

## Validation

- `2314 passed, 5 skipped, 2 xpassed` with 82.20% coverage
- Ruff
- mypy
- Bandit
- focused CLI, graph, and git-integration regressions

Closes #777

### DIFF
--- a/code_review_graph/cli.py
+++ b/code_review_graph/cli.py
@@ -1617,7 +1617,8 @@ def main() -> None:
         legacy_db = repo_root / ".code-review-graph.db"
         default_db = repo_root / ".code-review-graph" / "graph.db"
         if (
-            not db_path.exists()
+            not status_data_dir
+            and not db_path.exists()
             and db_path.resolve() == default_db.resolve()
             and legacy_db.exists()
         ):
--- a/code_review_graph/cli.py
+++ b/code_review_graph/cli.py
@@ -1606,8 +1606,8 @@ def main() -> None:
     if args.command in _data_dir_cmds:
         _handle_data_dir_option(args, repo_root)
 
-    db_path = get_db_path(repo_root)
-    if args.command in ("dead-code", "forget") and not db_path.exists():
+    db_path = get_db_path(repo_root, read_only=args.command == "status")
+    if args.command in ("dead-code", "forget", "status") and not db_path.exists():
         print(
             f"No graph found at {db_path}. Run `code-review-graph build` first.",
             file=sys.stderr,
--- a/code_review_graph/cli.py
+++ b/code_review_graph/cli.py
@@ -1609,7 +1609,12 @@ def main() -> None:
     if args.command == "status":
         db_path = get_db_path(repo_root, read_only=True)
         legacy_db = repo_root / ".code-review-graph.db"
-        if not db_path.exists() and legacy_db.exists():
+        default_db = repo_root / ".code-review-graph" / "graph.db"
+        if (
+            not db_path.exists()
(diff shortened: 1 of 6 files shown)


tirth8205:
Verified locally before merge: reproduced the #777 poisoning on main with a fresh 3-file repo (status created graph.db and exited 0; update --base on a missing graph built a partial graph and callers_of returned not_found). On this branch: status exits 1 with no side effects, update on a missing or empty graph forces a full rebuild, callers_of resolves correctly, and a revert-check confirmed the new tests fail without the fix. Full suite: 2314 passed, 5 skipped, 2 xpassed. Ruff clean, all 11 CI checks green. Follow-up issues will be filed for the other commands that still create an empty graph.db on first run (detect-changes, visualize, wiki, watch) and for healing graphs already poisoned by released versions.

PR #783, opened 2026-07-29
Add Voyage embedding provider

schulle01:
## Summary
- add Voyage AI as an embedding provider
- expose `voyage` in CLI embedding-provider choices
- persist embedding batches incrementally so large repository embeds can resume after provider rate limits
- add optional Voyage request throttling via `CRG_VOYAGE_MIN_INTERVAL_SEC`
- add provider tests for request payloads, response parsing, API key handling, provider identity, batch persistence, and throttling

## Defaults
- model: `voyage-code-3`
- output dimension: `1024`
- output dtype: `float`
- document embeddings use `input_type=document`
- query embeddings use `input_type=query`
- request throttling defaults to disabled unless `CRG_VOYAGE_MIN_INTERVAL_SEC` is set

## Tests
- `python -m pytest tests/test_embeddings.py -q`
- `python -m ruff check code_review_graph/cli.py code_review_graph/embeddings.py tests/test_embeddings.py`
- `python -m compileall -q code_review_graph tests/test_embeddings.py`
- `git diff --check`

## Operational note
This was tested in an internal CRG deployment with a reduced-rate Voyage account. A 21s request interval allowed a full monorepo embedding run to complete without 429s.

### DIFF
--- a/README.md
+++ b/README.md
@@ -542,6 +542,13 @@ pip install "code-review-graph[all]"                 # All optional dependencies
 | `CRG_TOOLS` | Comma-separated allowlist of MCP tools to expose when serving | - |
 | `GOOGLE_API_KEY` | API key for Google Gemini embeddings | - |
 | `MINIMAX_API_KEY` | API key for MiniMax embeddings | - |
+| `VOYAGE_API_KEY` | API key for Voyage embeddings | - |
+| `CRG_VOYAGE_MODEL` | Model name for Voyage embeddings | `voyage-code-3` |
+| `CRG_VOYAGE_OUTPUT_DIMENSION` | Output dimension for Voyage embeddings | `1024` |
+| `CRG_VOYAGE_OUTPUT_DTYPE` | Output dtype for Voyage embeddings | `float` |
+| `CRG_VOYAGE_BASE_URL` | Voyage embeddings endpoint | `https://api.voyageai.com/v1` |
+| `CRG_VOYAGE_BATCH_SIZE` | Batch size for Voyage embedding requests | `100` |
+| `CRG_VOYAGE_MIN_INTERVAL_SEC` | Minimum delay between Voyage requests | `0` |
 | `CRG_OPENAI_BASE_URL` | OpenAI-compatible embeddings endpoint | - |
 | `CRG_OPENAI_API_KEY` | API key for OpenAI-compatible embeddings | - |
 | `CRG_OPENAI_MODEL` | Model name for OpenAI-compatible embeddings | - |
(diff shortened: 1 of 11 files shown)


tirth8205:
Thanks, the Voyage provider itself is in good shape: key comes only from VOYAGE_API_KEY, the cloud egress warning is wired, provider identity includes model, dimension, dtype and endpoint, and the new tests are fully mocked (no network) and genuinely cover the batch persistence change. One blocker before this can merge: the full test suite fails on this branch. tests/test_tools.py::TestEmbedGraphProviderErrors::test_unknown_provider_returns_structured_error still uses "voyage" as an unknown provider name and now receives the missing VOYAGE_API_KEY error instead, so it fails. Please update it the same way you updated test_embeddings.py (for example use "moonbase"), and also adjust test_store_closed_when_provider_unknown in the same class, which uses "voyage" and now exercises a different error path than the one it documents. Two smaller asks: consider dropping the CRG_EMBEDDING_MODEL fallback for the Voyage model, since that variable is documented as the local sentence-transformers model and users who set it for local usage would send that model name to the Voyage API (the OpenAI provider deliberately requires its own model variable); and update the provider lists in the docstrings of main.py (semantic_search_nodes_tool, embed_graph_tool) and tools/docs.py (embed_graph) plus the README to mention voyage.

tirth8205:
Re-reviewed after the update. The blocker is fixed: both TestEmbedGraphProviderErrors tests now use "moonbase" and the class passes; full suite is green on the branch (2319 passed, 5 skipped, 2 xpassed) and on a clean test-merge with current main (2397 passed, 5 skipped, 2 xpassed). ruff and mypy clean. All smaller asks are done: CRG_EMBEDDING_MODEL no longer leaks into the Voyage model (with a regression test), provider lists updated in main.py, tools/docs.py, README, and docs, and eval --embed-provider gained voyage. The throttle is opt-in via CRG_VOYAGE_MIN_INTERVAL_SEC, uses time.monotonic with a single sleep (no busy-wait), and its test mocks both monotonic and sleep, so the suite stays fast. I also verified the new persistence tests fail if embed_nodes is reverted, so they genuinely cover the change. One note for a follow-up, not a blocker: embed_nodes now chunks at 64 for all providers before calling provider.embed, so cloud requests carry at most 64 texts instead of the providers' internal 100, roughly 56% more HTTP requests on large graphs (same total tokens). The per-batch commit is worth it since a failed cloud run now keeps completed batches and re-runs skip up-to-date embeddings, but consider passing the provider's batch size (or 100) as the chunk size to get both benefits. Good to merge.

PR #785, opened 2026-07-30
Add Oracle PL/SQL structural parsing

satwikjambula:
Adds support for Oracle PL/SQL constructs that tree-sitter-sql cannot parse (IS/AS ... BEGIN ... END bodies with no RETURNS/LANGUAGE clause parse as ERROR nodes under the ANSI-oriented grammar), via the same regex-fallback approach already used for CREATE PROCEDURE:

- Route Oracle-specific extensions (.pls, .plb, .pks, .pkb, .pck, .prc, .fnc, .trg) through the existing SQL parser dispatch.
- Extract CREATE PACKAGE / PACKAGE BODY (with nested PROCEDURE/FUNCTION members qualified as pkg_name.member_name), CREATE TRIGGER, and Oracle-style CREATE FUNCTION/PROCEDURE.
- Add regex-based CALLS edge extraction within procedure/function/ trigger/package-member bodies, skipping PL/SQL keywords and scalar type names that also precede '(' (e.g. VARCHAR2(50) declarations).
- Add tests/fixtures/sample_plsql.sql and a TestPLSQLParsing suite covering packages, triggers, nested members, and call edges.
- Update README/docs language lists and CHANGELOG.

# Pull Request

## Linked issue

Closes #

## How it was tested

```bash
```

## Checklist

- [ ] Tests added for new functionality
- [ ] All tests pass: `uv run pytest tests/ --tb=short -q`
- [ ] Linting passes: `uv run ruff check code_review_graph/`
- [ ] Type checking passes: `uv run mypy code_review_graph/ --ignore-missing-imports --no-strict-optional`
- [ ] Lines are at most 100 characters


tirth8205:
Thanks - the problem is real (verified: tree-sitter-sql emits ERROR nodes for IS/AS...BEGIN bodies, and .pkb/.pls files were skipped entirely on main), the happy path works, and the full suite passes locally (2317 passed). Four issues need fixing before merge: 1) Non-Oracle regression: for T-SQL/MySQL procedures with no 'END;' (bare END + GO, or END$$), _plsql_block_end falls back to len(text), so procedures now span to EOF and pick up false CALLS edges across the rest of the file (I observed CALLS to VARCHAR, DATEADD, GETDATE, and a later table from both procedures in a two-proc T-SQL file). Please fall back to the old single-line span (and skip call extraction) when no END terminator is found. 2) Forward declarations in package bodies ('PROCEDURE log_event(p IN VARCHAR2);') create a phantom member whose block-end lands inside the next member, emitting false CALLS edges (log_event -> total_orders in my test). Require IS/AS after the signature before treating a member as a definition. 3) Call noise: 'INSERT INTO t (cols)' emits CALLS to the table, and builtins COUNT/NVL/TO_CHAR/SUBSTR become CALLS targets; plain VARCHAR also leaks (keyword list has varchar2 only). Consider skipping identifiers preceded by INTO and adding common builtins/types to the skip list. 4) Package spec and body both produce Class 'file::pkg' which collide on the nodes UNIQUE(qualified_name) upsert, so one is lost. Also please add tests for bare 'END;' members, a forward declaration, and a T-SQL non-regression case - the current fixture only uses named 'END name;' blocks.

satwikjambula:
> Thanks - the problem is real (verified: tree-sitter-sql emits ERROR nodes for IS/AS...BEGIN bodies, and .pkb/.pls files were skipped entirely on main), the happy path works, and the full suite passes locally (2317 passed). Four issues need fixing before merge: 1) Non-Oracle regression: for T-SQL/MySQL procedures with no 'END;' (bare END + GO, or END$$), _plsql_block_end falls back to len(text), so procedures now span to EOF and pick up false CALLS edges across the rest of the file (I observed CALLS to VARCHAR, DATEADD, GETDATE, and a later table from both procedures in a two-proc T-SQL file). Please fall back to the old single-line span (and skip call extraction) when no END terminator is found. 2) Forward declarations in package bodies ('PROCEDURE log_event(p IN VARCHAR2);') create a phantom member whose block-end lands inside the next member, emitting false CALLS edges (log_event -> total_orders in my test). Require IS/AS after the signature before treating a member as a definition. 3) Call noise: 'INSERT INTO t (cols)' emits CALLS to the table, and builtins COUNT/NVL/TO_CHAR/SUBSTR become CALLS targets; plain VARCHAR also leaks (keyword list has varchar2 only). Consider skipping identifiers preceded by INTO and adding common builtins/types to the skip list. 4) Package spec and body both produce Class 'file::pkg' which collide on the nodes UNIQUE(qualified_name) upsert, so one is lost. Also please add tests for bare 'END;' members, a forward declaration, and a T-SQL non-regression case - the current fixture only uses named 'END name;' blocks.

Thanks for the thorough review. All four issues were valid and reproducible, so I really appreciate you catching them.

I’ve addressed each one in the latest commits:

Non-Oracle regression: I added dialect detection (_is_plsql_dialect) so every PL/SQL-specific pass—body span calculation, CALLS extraction, packages, and triggers—only runs when the file is actually Oracle PL/SQL. The check uses either the file extension or Oracle-specific content signals such as VARCHAR2, %TYPE, or :NEW. As a result, a T-SQL or MySQL CREATE PROCEDURE without an END; now behaves exactly as it did before this PR: it remains a single-line node with no computed span, rather than incorrectly expanding to the end of the file.

Forward declarations: I added _plsql_is_forward_decl to detect signature-only declarations such as PROCEDURE x(...); inside package bodies. These are now skipped, so they’re no longer treated as real members whose block-end search can accidentally attach to the next member’s END.

Call extraction noise: I expanded the filtering logic by excluding VARCHAR (previously only VARCHAR2 was excluded), adding a blocklist for common built-in functions (COUNT, NVL, TO_CHAR, SUBSTR, etc.), filtering out DBMS_* and UTL_* package calls, and excluding the target table in INSERT INTO t (...) statements.

Duplicate nodes: Package bodies now receive a distinct qualified name (pkg_name$body) instead of sharing the same name as the package specification. While validating this against the real GraphStore (not just the parser output), I also found a related bug: GraphStore.upsert_node derives the persisted qualified name from identity_name and parent_name, independently of the parser’s local edge-target strings. The fix therefore had to be applied through those fields, not just within _parse_sql. I added a regression test that exercises the actual store layer so this class of issue can’t slip through unnoticed again.

I also added the requested test coverage for:

bare END; members,

forward declarations, and

the T-SQL non-regression case.

I ran the full suite (2,400+ tests, including the upstream tests), along with ruff and mypy, and everything passes. I also verified that the branch merges cleanly against the current main.

tirth8205:
Re-reviewed after the update. All four blockers are verifiably fixed: the two-proc T-SQL file (bare END + GO) now keeps single-line spans with zero CALLS edges (MySQL END$$ too), the forward-declaration fixture produces exactly one log_event member with no phantom edges, the noise fixture emits only real user calls (VARCHAR, COUNT/NVL/TO_CHAR/SUBSTR, DBMS_*/UTL_*, and INSERT INTO targets all filtered), and spec+body both survive the real GraphStore with distinct qualified names and no dangling CONTAINS targets - the identity_name/parent_name threading matches _make_qualified. I also confirmed the new tests fail against the pre-fix parser (the T-SQL test fails with span 2..11, the exact EOF-bleed regression). Full suite 2400 passed / 5 skipped / 2 xpassed; ruff and mypy clean.

One remaining defect in the new _plsql_is_forward_decl: it tries re.match(r\"(IS|AS)\\b\", text[i:]) at every character position, with no leading word boundary and no comment handling. Two consequences I reproduced: (1) a forward declaration like 'FUNCTION get_alias(p) RETURN t_alias;' (any depth-0 word ending in as/is, including 'RETURN employees.alias%TYPE;') matches AS mid-word and becomes a phantom member with false CALLS edges - the same failure mode as blocker 2, just on a narrower trigger; (2) a real definition with a semicolon inside a trailing comment, e.g. 'PROCEDURE setup(p IN NUMBER) -- initializes state; call once' followed by IS, is classified as a forward declaration and silently dropped from the graph. Suggested fix: only attempt the IS/AS match when the previous character is not a word character, and skip -- and /* */ comments during the scan. That plus a small test for each case and this is good to merge. (Also noted, non-blocking: the named END search in _plsql_block_end is unbounded to EOF, so same-named members across two packages can bleed spans when the first uses a bare END; - pre-existing from the original PR, fine as a follow-up.)

satwikjambula:
> Re-reviewed after the update. All four blockers are verifiably fixed: the two-proc T-SQL file (bare END + GO) now keeps single-line spans with zero CALLS edges (MySQL END$$ too), the forward-declaration fixture produces exactly one log_event member with no phantom edges, the noise fixture emits only real user calls (VARCHAR, COUNT/NVL/TO_CHAR/SUBSTR, DBMS__/UTL__, and INSERT INTO targets all filtered), and spec+body both survive the real GraphStore with distinct qualified names and no dangling CONTAINS targets - the identity_name/parent_name threading matches _make_qualified. I also confirmed the new tests fail against the pre-fix parser (the T-SQL test fails with span 2..11, the exact EOF-bleed regression). Full suite 2400 passed / 5 skipped / 2 xpassed; ruff and mypy clean.
> 
> One remaining defect in the new _plsql_is_forward_decl: it tries re.match(r"(IS|AS)\b", text[i:]) at every character position, with no leading word boundary and no comment handling. Two consequences I reproduced: (1) a forward declaration like 'FUNCTION get_alias(p) RETURN t_alias;' (any depth-0 word ending in as/is, including 'RETURN employees.alias%TYPE;') matches AS mid-word and becomes a phantom member with false CALLS edges - the same failure mode as blocker 2, just on a narrower trigger; (2) a real definition with a semicolon inside a trailing comment, e.g. 'PROCEDURE setup(p IN NUMBER) -- initializes state; call once' followed by IS, is classified as a forward declaration and silently dropped from the graph. Suggested fix: only attempt the IS/AS match when the previous character is not a word character, and skip -- and /* */ comments during the scan. That plus a small test for each case and this is good to merge. (Also noted, non-blocking: the named END search in _plsql_block_end is unbounded to EOF, so same-named members across two packages can bleed spans when the first uses a bare END; - pre-existing from the original PR, fine as a follow-up.)

Both fixed.

Mid-word AS/IS match — re.match(r"(IS|AS)\b", text[i:]) only anchored the end of the match with \b; there was no check on the start. Added a boundary check so the regex is only attempted when the previous character isn't a word character:

`if c.isalpha() and (i == start or not (text[i-1].isalnum() or text[i-1] == "_")):`

Verified with your exact repro — FUNCTION get_alias(p) RETURN t_alias; is now correctly classified as a forward declaration, and the employees.alias%TYPE variant too.

Comment before IS — added --//* */ comment-skipping directly inside _plsql_is_forward_decl itself (not just relying on the clean_text call site upstream), so the function is correct standalone. Your exact repro (PROCEDURE setup(p IN NUMBER) -- initializes state; call once followed by IS) is now correctly classified as a real definition.

Added a unit test for each case (calling _plsql_is_forward_decl directly) plus a parse_file() integration test for the t_alias scenario. While auditing for the same failure pattern elsewhere, found and fixed three more instances: _plsql_block_end matched "ENDfoo;" as "END foo;" (no boundary after END), the CREATE-statement regexes matched "RECREATE PROCEDURE" as "CREATE PROCEDURE" (no leading boundary), and _plsql_member_signature could get confused by an unbalanced paren inside a string-literal default value. All now boundary-anchored and independently comment/string-safe, with tests for each.

Agreed on the unbounded named-END search — noted as a follow-up, not blocking this PR.

Full suite passing (2400+ tests), ruff and mypy clean, pushed.

satwikjambula:
Hi @tirth8205 — this addresses all the feedback from both review rounds (the four original blockers, the mid-word AS/IS boundary issue, and the follow-up hardening pass), plus the branch is now rebased clean against current main with no conflicts. Full suite (2440+ tests), ruff, and mypy all pass. Would appreciate your review and, if it looks good, an approve/merge whenever you have a chance.

tirth8205:
The correctness work from both rounds holds up, and the suite, ruff, and mypy are green on a merge with current main. One performance defect needs fixing before merge: in _extract_plsql_calls (parser.py:5166), the INSERT INTO filter runs re.search over body_text[:m.start()] for every call-site match, rescanning the whole preceding body each time, so call extraction is quadratic. Measured: 8k call sites in a 40KB .pls body takes 1.3s, 32k sites in 160KB takes 25s, and real package bodies in the hundreds of KB would stall a build for minutes. A bounded lookback of a few characters fixes it; the per-target line count at parser.py:5171 has the same shape and could use a running counter. Two smaller notes, non-blocking: _strip_plsql_noise resets string state per line, so a multi-line string literal containing 'END p;' truncates the span and drops later calls, and _PLSQL_IDENT is ASCII-only, so cafe-accented identifiers that main captured in full now truncate at the first non-ASCII char. Happy to merge once the quadratic is fixed.

satwikjambula:
> The correctness work from both rounds holds up, and the suite, ruff, and mypy are green on a merge with current main. One performance defect needs fixing before merge: in _extract_plsql_calls (parser.py:5166), the INSERT INTO filter runs re.search over body_text[:m.start()] for every call-site match, rescanning the whole preceding body each time, so call extraction is quadratic. Measured: 8k call sites in a 40KB .pls body takes 1.3s, 32k sites in 160KB takes 25s, and real package bodies in the hundreds of KB would stall a build for minutes. A bounded lookback of a few characters fixes it; the per-target line count at parser.py:5171 has the same shape and could use a running counter. Two smaller notes, non-blocking: _strip_plsql_noise resets string state per line, so a multi-line string literal containing 'END p;' truncates the span and drops later calls, and _PLSQL_IDENT is ASCII-only, so cafe-accented identifiers that main captured in full now truncate at the first non-ASCII char. Happy to merge once the quadratic is fixed.

Thanks for the detailed measurements  @tirth8205 — confirmed the same quadratic behavior locally.

Root cause: _extract_plsql_calls (parser.py) checked for a preceding INSERT INTO by re-slicing and re-searching body_text[:m.start()] on every call-site match, so the whole preceding body got rescanned on each match — O(n²) overall.

Fix: replaced the full-prefix slice with a fixed-size lookback window (_INTO_LOOKBACK = 60 chars). INTO always appears immediately before its target, so a small bounded window is enough to catch it correctly (including across a couple of formatting/newline variations), without rescanning from the start of the body every time.

Verification (local):

Reproduced the quadratic scaling before the fix: 1,500 calls → 0.54s, 3,000 → 1.81s, 6,000 → 7.13s.
After the fix: 1,500 → 0.15s, 3,000 → 0.13s, 6,000 → 0.43s, 12,000 → 1.34s — roughly linear.
Added two regression tests: one asserting 4,000 calls complete in well under 5s (would fail immediately on a quadratic reintroduction), and one confirming the bounded lookback still correctly filters INSERT INTO across multi-line formatting.
Full suite, ruff, and mypy all pass.

Also merged current main into the branch and resolved the resulting conflicts in README.md and [tests/test_multilang.py](https://github.com/satwikjambula/saltcorn-flow/blob/main/tests/test_multilang.py) — both were purely positional (adjacent/same-line additions from unrelated PRs), not logical conflicts. Re-ran the full suite after the merge to confirm nothing broke.

Scope stayed limited to the reported blocker — I haven't touched the two non-blocking items you flagged (multi-line string literals in _strip_plsql_noise, ASCII-only identifier regex) unless you'd like those addressed in this PR too, happy to follow up.

satwikjambula:
Hi @tirth8205 — following up on this PR. All feedback from the reviews so far has been addressed:

The original 4 blockers (non-Oracle regression, forward-declaration phantom members, call-extraction noise, spec/body node collision)
The forward-declaration IS/AS word-boundary + comment-handling defect from the follow-up review
The quadratic performance regression in _extract_plsql_calls (bounded lookback window, verified linear up to 12k calls)
Merge conflicts against current main (resolved and re-verified)
Full test suite (2440+ tests), ruff, and mypy are all passing on the latest commit. This has been open for a while now — whenever you get a chance, could you take another look and let me know if it's good to merge, or if anything else needs changing?

Thanks for all the detailed review so far, it's helped a lot in getting this right.

tirth8205:
Outside this sweep's mission of demonstrated coding-session token savings: this introduces a substantial Oracle PL/SQL dialect surface. There is also a supported-SQL regression: save `CREATE PROCEDURE café AS SELECT 1;` in `names.sql` and call `CodeParser(repo_root).parse_file(path)`; this branch names the procedure `caf`, while main preserves `café`. That identifier loss needs correcting before the expansion is reconsidered.

satwikjambula:
> Outside this sweep's mission of demonstrated coding-session token savings: this introduces a substantial Oracle PL/SQL dialect surface. There is also a supported-SQL regression: save `CREATE PROCEDURE café AS SELECT 1;` in `names.sql` and call `CodeParser(repo_root).parse_file(path)`; this branch names the procedure `caf`, while main preserves `café`. That identifier loss needs correcting before the expansion is reconsidered.

Good catch — confirmed the regression with your exact repro.

Root cause: _PLSQL_IDENT was [A-Za-z_][A-Za-z0-9_$#]*, an ASCII-only identifier pattern, unlike the \w-based identifier matching used elsewhere in this file for non-Oracle SQL (\w is Unicode-aware by default in Python). So café got silently truncated to caf in the PL/SQL path while main correctly preserved it via \w.

Fix: changed the leading character class from [A-Za-z_] to [^\W\d] (word character that isn't a digit — i.e. Unicode letter or underscore), keeping the rest of the class ([\w$#]*) to preserve the existing $/# mid-identifier handling Oracle allows (e.g. log_event$v2).

Verification:

CREATE PROCEDURE café AS SELECT 1; now correctly produces a node named café, matching main's behavior.
Confirmed $/# handling is untouched — log_event$v2 still parses as a single identifier, not truncated at the $.
Added two regression tests: one for the accented-identifier case, one confirming $/# still can't start an identifier (matching Oracle's own rules).
Full suite (3000+ tests), ruff, and mypy all pass.

This addresses the identifier-loss regression you flagged as blocking further consideration of the PL/SQL expansion. Let me know if there's anything else you'd like adjusted.

tirth8205:
This no longer merges into `staging`. Conflicts in `CHANGELOG.md`, `README.md`, `code_review_graph/docs/LLM-OPTIMIZED-REFERENCE.md`, `docs/FEATURES.md`, `docs/USAGE.md`.

```bash
# resolve, then
uv run pytest tests/ -q
git push
```

When resolving:
- Rebase or merge origin/staging into the PR branch and resolve conflicts in CHANGELOG.md, README.md, code_review_graph/docs/LLM-OPTIMIZED-REFERENCE.md, docs/FEATURES.md, docs/USAGE.md before re-assessment.

satwikjambula:
[tirth8205](https://github.com/tirth8205) resolved conflicts

tirth8205:
The conflicts are resolved properly. README stays at 31 feature rows with no duplicated rows, and CHANGELOG.md, LLM-OPTIMIZED-REFERENCE.md and USAGE.md each take staging's text plus your one line. docs/FEATURES.md came out identical to staging, which is the right result.

Gates on `ea799b4` merged into `staging` 9c13f4c: `3381 passed, 9 skipped, 2 xpassed`, ruff clean, mypy clean. The identifier regression is fixed: `CREATE PROCEDURE café AS SELECT 1;` now parses to `Function 'café'`.

So there is no technical objection left on this branch.

I am not queueing it for merge, and the reason is not a defect. Whether to take on an Oracle PL/SQL dialect surface at all is a scope call that has not been made, and that call belongs to the maintainer, not to this review pass. Nothing for you to fix while it is pending.

One thing to be aware of: only GitGuardian has run on `ea799b4`. The full CI matrix has not, so the numbers above come from running the three gate commands locally on the merge, not from CI.

Removing `needs-rebase`.

tirth8205:
This no longer merges into `staging`. 2 files conflict and the branch is 144 commits behind:

- `CHANGELOG.md`
- `README.md`

```bash
git fetch origin
git merge origin/staging
# resolve, then
uv run pytest tests/ -q
git push
```

PR #787, opened 2026-07-30
fix(swift): emit Function nodes for init/deinit/subscript (#786)

### LINKED ISSUE #786: [Bug]: Swift init/deinit/subscript produce no Function nodes; their body calls are attributed to the File node
### code-review-graph version

2.3.5 (also present on `main` @ 90d760a)

### Operating system

macOS

### Python version

3.14.6

### AI platform

claude-code

```shell
Nodes: 14550
Edges: 86924
Files: 1242
Languages: bash, python, typescript, javascript, swift, ruby
Last updated: 2026-07-30T13:47:30
Built on branch: main
Built at commit: a3bb6f34f6e0
```

### Steps to reproduce

1. `code-review-graph build` on a Swift repository (mine: an iOS app, 1,242 indexed files)
2. Query the graph for initializers:

```bash
sqlite3 .code-review-graph/graph.db \
  "select count(*) from nodes where language='swift' and name in ('init','deinit');"
# => 0

# ground truth in the two indexed source roots
rg -c '^\s*(public |internal |private |required |convenience |override )*init\(' \
  --glob '*.swift' Bitazza Packages | awk -F: '{s+=$2} END {print s}'
# => 712
```

Minimal repro — parse this file with `CodeParser().parse_file()`:

```swift
import Foundation

class Session {
    let token: String
    init(token: String) { self.token = token }
    convenience init() { self.init(token: "anon") }
    deinit { Logger.flush() }
    subscript(key: String) -> String? { return nil }
    func refresh() { Network.send(token) }
}
```

**Expected:** `init`, `deinit` and `subscript` produce `Function` nodes, and the calls in
their bodies are attributed to them — the same way `java`/`csharp` list
`constructor_declaration` in `_FUNCTION_TYPES`.

**Actual:** all three are invisible. Only `refresh` becomes a node, and the calls made
inside the initializer and deinitializer bodies are attributed to the **File** node:

```
# nodes
Class     Session
Function  refresh    parent=Session

# edges
CALLS  probe.swift -> init      # <- source is the File, not the initializer
CALLS  probe.swift -> flush     # <- source is the File, not the deinit
CALLS  probe.swift::Session.refresh -> send
```

Root cause is in `parser.py`:

```python
_FUNCTION_TYPES = {
    ...
    "java":   ["method_declaration", "constructor_declaration"],
    "csharp": ["method_declaration", "constructor_declaration"],
    "swift":  ["function_declaration"],   # <- init/deinit/subscript missing
```

tree-sitter-swift emits `init_declaration`, `deinit_declaration` and
`subscript_declaration` as node types distinct from `function_declaration`, so the generic
walker never matches them.

They also need naming help, which is presumably why they were left out — none of the three
has a usable name field:

| node type | `child_by_field_name("name")` |
|---|---|
| `init_declaration` | the `init` keyword |
| `deinit_declaration` | **absent** → generic loop returns `None`, node dropped |
| `subscript_declaration` | the **return type** (`subscript(i: Int) -> String` names itself `String`) |

### Additional context

The mis-attribution is the part that hurts more than the missing nodes: because
initializer-body calls are sourced from the File node, `get_impact_radius` /
`detect_changes` on a callee can't trace back to the initializer that calls it, and a
change inside one `init` reads as file-wide blast radius.

Measured with a patched parser run over every `.swift` file in my working tree (2,282 files
— wider than the 1,242 the graph indexes, since it includes test and submodule sources):

- **1,816 declarations recovered** (1,784 `init`, 28 `deinit`, 4 `subscript`) — pre-collapse count; the `ON CONFLICT(qualified_name)` upsert still folds overloaded initializers of the same type into one node, same as Java/C# overloads today
- **2,933 CALLS edges** re-attributed from the File node to the declaration that actually makes the call

For an iOS/SwiftUI codebase this is the single largest Swift extraction gap — dependency
injection, `@Dependency` wiring and view-model construction all live in initializers.

PR follows.

Separately, and I'll file it on its own if you'd prefer: 85% of the CALLS edges in my graph
(56,511 of 66,203) point at unresolved names, dominated by SwiftUI/XCTest/Foundation
builtins — `XCTAssertEqual` (3,286), `withLock` (1,847), `eraseToAnyPublisher` (1,426),
`foregroundColor`, `padding`, `frame`. They inflate the store (252 MB `graph.db` for a
93k-line repo) and rank as high-degree nodes. Happy to open a second issue for a Swift
builtin filter if that's useful.

kankunnawat:
# Pull Request

## Linked issue

Closes #786

## What & why

`_FUNCTION_TYPES["swift"]` listed only `function_declaration`, but tree-sitter-swift parses
initializers, deinitializers and subscripts as `init_declaration`, `deinit_declaration` and
`subscript_declaration` — separate node types. All three were invisible to the graph, and
the calls inside their bodies were attributed to the enclosing **File** node:

```swift
class Session {
    let token: String
    init(token: String) { self.token = token }
    convenience init() { self.init(token: "anon") }
    deinit { Logger.flush() }
    subscript(key: String) -> String? { return nil }
    func refresh() { Network.send(token) }
}
```

| | before | after |
|---|---|---|
| nodes | `Class Session`, `Function refresh` | + `Session.init` ×2, `Session.deinit`, `Session.subscript` |
| edges | `CALLS probe.swift -> init`<br>`CALLS probe.swift -> flush` | `CALLS Session.init -> Session.init`<br>`CALLS Session.deinit -> flush` |

The mis-attribution is the more damaging half: a change inside one initializer reads as
file-wide blast radius, and `get_impact_radius` on a callee can't trace back to the
initializer that calls it.

Two changes:

1. **`_FUNCTION_TYPES["swift"]`** — list the three node types alongside `function_declaration`,
   the way `java`/`csharp` already list `constructor_declaration`.
2. **`_get_name`** — name each after its declaration keyword. The grammar gives none of the
   three a usable name field, which is presumably why they were left out:

   | node type | `child_by_field_name("name")` |
   |---|---|
   | `init_declaration` | the `init` keyword |
   | `deinit_declaration` | **absent** → generic loop returns `None`, node dropped |
   | `subscript_declaration` | the **return type** (`subscript(i: Int) -> String` → named `String`) |

Scope notes:

- Overloaded initializers of one type still collapse into a single node under the existing
  `ON CONFLICT(qualified_name) DO UPDATE` upsert — identical to Java/C# constructor and
  method overloads today. Left alone deliberately; it's a language-agnostic identity
  question, not a Swift one.
- No `extra["swift_kind"]` is set on the new nodes; the node name already carries the
  distinction.

Measured over the 2,282 `.swift` files of a production iOS app: **1,816 declarations
recovered** (1,784 `init`, 28 `deinit`, 4 `subscript`) and **2,933 CALLS edges** re-attributed
from the File node to the declaration that actually makes the call.

## How it was tested

```bash
# 2309 passed, 5 skipped, 2 xpassed in 37.54s

uv run ruff check code_review_graph/ tests/test_multilang.py
# All checks passed!

```

Four tests added to `TestSwiftParsing`, over an extended `tests/fixtures/sample.swift`
(`init(seed:)`, `convenience init()`, `deinit`, `subscript(id:)` on `InMemoryRepo`):

- `test_finds_initializers` — both initializers are `Function` nodes parented to the type
- `test_finds_deinitializer`
- `test_finds_subscript` — asserts it's named `subscript`, not `User` (its return type)
- `test_initializer_body_calls_attributed_to_declaration` — the regression that matters:
  `init -> save`, `init -> init` (delegation), `deinit -> removeAll`,
  `subscript -> findById`, and no CALLS edge sourced from the File node

## Checklist

- [x] Docs updated where behavior changed (CHANGELOG entry under `[Unreleased] / Fixed`)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,15 @@
 
 ### Fixed
 
+- Swift initializers, deinitializers and subscripts now emit `Function` nodes
+  (#786). `init_declaration`, `deinit_declaration` and `subscript_declaration`
+  are node types distinct from `function_declaration`, so the walker skipped
+  them and attributed the calls in their bodies to the enclosing *File* node —
+  a change inside one initializer read as file-wide blast radius, and
+  `get_impact_radius` on a callee could not trace back to the initializer that
+  calls it. Each is named after its declaration keyword, since the grammar
+  gives none of the three a usable name field (`subscript`'s would be its
+  return type, `deinit`'s is absent).
 - GitHub Copilot auto-detection now requires the Copilot extension and also
   recognizes the extension bundled with released VS Code installations.
 - Added a post-index Python import resolver using unique module suffixes, so
--- a/code_review_graph/parser.py
+++ b/code_review_graph/parser.py
@@ -949,7 +949,16 @@ _FUNCTION_TYPES: dict[str, list[str]] = {
     "r": ["function_definition"],
     "perl": ["subroutine_declaration_statement", "method_declaration_statement"],
     "kotlin": ["function_declaration"],
-    "swift": ["function_declaration"],
+    # Swift: initializers, deinitializers and subscripts are separate node
+    # types, not `function_declaration`s, so they need listing alongside it —
+    # the same way java/csharp list `constructor_declaration`. Their names come
+    # from the `_get_name` Swift branch (the grammar has no usable name field).
+    "swift": [
+        "function_declaration",
+        "init_declaration",
+        "deinit_declaration",
+        "subscript_declaration",
+    ],
     "php": ["function_definition", "method_declaration"],
     "scala": ["function_definition", "function_declaration"],
     # Solidity: events and modifiers use kind="Function" because the graph
@@ -14147,6 +14156,18 @@ class CodeParser:
             for child in node.children:
                 if child.type == "identifier":
                     return child.text.decode("utf-8", errors="replace")
+        # Swift init/deinit/subscript: the grammar gives none of them a usable
+        # name. `init_declaration`'s name field is the `init` keyword itself,
+        # `deinit_declaration` has no name field at all (so the generic loop
+        # returns None and the node is dropped), and `subscript_declaration`'s
+        # name field is the *return type* (`subscript(i: Int) -> String` would
+        # be named "String"). Name each after its Swift declaration keyword.
+        if language == "swift" and node.type in (
+            "init_declaration",
+            "deinit_declaration",
+            "subscript_declaration",
+        ):
+            return node.type.removesuffix("_declaration")
         # Swift extensions: name is inside user_type > type_identifier
         # (e.g. `extension MyClass: Protocol { ... }`)
         if language == "swift" and node.type == "class_declaration":
--- a/tests/fixtures/sample.swift
+++ b/tests/fixtures/sample.swift
@@ -14,6 +14,24 @@ struct User {
 class InMemoryRepo: UserRepository {
     private var users: [Int: User] = [:]
 
+    init(seed: [User]) {
+        for user in seed {
+            save(user)
+        }
+    }
+
+    convenience init() {
+        self.init(seed: [])
+    }
+
+    deinit {
+        users.removeAll()
+    }
+
+    subscript(id: Int) -> User? {
+        return findById(id)
+    }
+
     func findById(_ id: Int) -> User? {
         return users[id]
     }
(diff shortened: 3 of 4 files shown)


tirth8205:
Verified locally before merge: reproduced #786 on main (init, deinit, and subscript produced no Function nodes and their body calls were attributed to the File node), confirmed the fix routes them through the existing walker with correct parenting and call attribution, and probed additional cases beyond the PR's tests (failable init, required init, static subscript, struct/enum/actor init and deinit, extension-added members, protocol requirements) with correct results. Revert-check confirmed the four new tests fail without the fix. Full suite: 2309 passed, 5 skipped, 2 xpassed. Ruff and mypy clean, all 11 CI checks green.

PR #794, opened 2026-07-30
[Bug]: C# method names extracted as the return type: async Task Foo() indexes as Task, merging disti

### LINKED ISSUE #791: [Bug]: C# method names extracted as the return type: async Task Foo() indexes as Task, merging distinct methods onto one qualified_name
### code-review-graph version

2.3.7

### Operating system

Linux

### Python version

3.13.5

### AI platform

claude-code

```shell
Nodes: 14473
Edges: 158703
Files: 1972
Languages: javascript, sql, bash, python, csharp
Last updated: 2026-07-30T13:44:51
Built on branch: main
Built at commit: cdc530e572dd
```

### Environment

  - code-review-graph 2.3.7 (PyPI), Python 3.13, Linux
  - tree-sitter-language-pack C# grammar
  - Repo: C# / .NET 10, 1 972 files

  ### Root cause

  `_get_name()` has an explicit special case for Java `method_declaration` (added to avoid
  exactly this bug) but none for C#, so C# falls through to the generic
  "first identifier-ish child" loop. The two grammars differ in the one way that matters:

      java   children: [modifiers, type_identifier 'Task', identifier 'foo',   formal_parameters, block]
      csharp children: [modifier, modifier, identifier 'Task', identifier 'Foo', parameter_list,  block]

  C# has no `type_identifier`. A non-generic return type is itself an `identifier`, so the
  generic loop matches it first.

  This also explains why the bug is invisible in many codebases: only **non-generic** return
  types are affected. `Task<T>` is a `generic_name` and `void`/`string`/`int` are
  `predefined_type` — both are skipped by the loop, so those methods come out correct. It is
  specifically the `async Task Name()` shape (idiomatic for xUnit/NUnit/TUnit async tests)
  that breaks.

  ### Reproduction

  ```python
  from pathlib import Path

  Path("T.cs").write_text("""
  class T {
    public async Task Should_do_thing() { }
    public async Task<int> Returns_generic() { return 1; }
    public void PlainVoid() { }
  }
  """)

  nodes = CodeParser().parse_file(Path("T.cs"))
  nodes = nodes[0] if isinstance(nodes, tuple) else nodes
  print([n.name for n in nodes if n.kind in ("Function", "Test")])
  ```

  Actual:   `['Task', 'Returns_generic', 'PlainVoid']`
  Expected: `['Should_do_thing', 'Returns_generic', 'PlainVoid']`

### Summary

  For C#, `_get_name()` returns the **return type** instead of the method name whenever the
  return type is non-generic. `public async Task Foo()` is indexed as `Task`.

  Because the name feeds `qualified_name`, every affected method in a class collapses to the
  same key (`File.cs::Class.Task`) and they are **merged into a single node**. This is not just
  a labelling problem — most of the test graph disappears.

  On a ~2 000-file C#/.NET repo (TUnit tests, so nearly every test is `public async Task Name()`):

  | | v2.3.7 | after the fix below |
  |---|---|---|
  | `Test` nodes | 583 | **3 492** |
  | `Test` nodes named `Task` | 568 | **0** |
  | total nodes | 10 883 | 14 473 |

  Mis-named `method_declaration` / `constructor_declaration` nodes: **4 188/5 327 (78.6 %)**
  under `tests/`, 400/1 444 (27.7 %) under `shared/`, 136/701 (19.4 %) under `src/`.

### Suggested fix

  Note that **adding `"csharp"` to the existing Java branch does not work** — that branch
  takes the first plain `identifier`, which for C# is the return type. Reading the `name`
  field is correct and robust:

  ```python
  if language == "csharp" and kind == "function" and node.type in (
      "method_declaration", "constructor_declaration",
  ):
      name_node = node.child_by_field_name("name")
      if name_node is not None:
          return name_node.text.decode("utf-8", errors="replace")
  ```

  Deliberately no `return None` fallback, so unusual node shapes still fall through to the
  generic loop rather than being dropped.

  Verified locally against the repo above: 4 188/5 327 test-tree methods corrected, zero
  `Task`-named nodes remain after a full rebuild, node count 10 883 → 14 473, and no
  regressions in `callers_of` / `children_of` output.

  `child_by_field_name("name")` would arguably be a safer default for Java too, but I have
  left that branch untouched.

  ### Related

  This is the same defect class already fixed for other languages, which is why C# looks like
  an oversight rather than a new problem:

  - #395 → #403 — C++ scoped definitions with non-primitive return types extracted the return type as the name
  - #371 — C++ scoped/destructor/operator name extraction
  - #622 → #663 — C++ overload identity sharing one `qualified_name` (same collapse mechanism as here)
  - v2.3.3 — "name extraction fixes for Kotlin/Swift (`simple_identifier`), Ruby (`constant`), C/C++ nested `function_declarator`"

loonister1:
`_get_name()` had a Java special case for the return-type-before-name shape but none for C#, where a non-generic return type like `Task` is itself an `identifier` and so was matched first by the generic loop. Added a C# branch that reads the `name` field of `method_declaration`/`constructor_declaration`, which is grammar-accurate and unaffected by generic vs. non-generic return types; it deliberately falls through to the generic loop when no `name` field is present.

Fixes #791

### Testing
[targeted; 3 commit(s) checked individually] $ /Users/m2/Documents/PW/github-bot/work/tirth8205__code-review-graph/791/.venv/bin/python -m pytest -q --no-header tests/test_multilang.py tests/test_parser.py

### Notes for reviewers
Scrutinise the `kind == "function"` guard: the issue reports these methods surfacing as `Test` nodes on a TUnit repo, and the new test only exercises a non-test file (`x.cs`), so all its nodes are `Function`. If `_get_name()` is ever called with a `test`-ish kind for methods in test files, the branch would not fire and the exact case the issue is about would stay broken — I could not verify from the diff whether `kind` here is always "function" for method_declaration regardless of Function/Test classification downstream. That is the first thing I would check, ideally by adding a case under a `Tests/` path or an `[Test]`-attributed method. Secondary, smaller items: (a) whether C# node types beyond `method_declaration`/`constructor_declaration` have the same shape — `local_function_statement`, `operator_declaration`, `destructor_declaration`, `delegate_declaration` — the contributor scoped narrowly, which is defensible but is a scope call that is yours; (b) the reported node-count jump (10,883 → 14,473) means previously-merged nodes now split, so downstream consumers keyed on `qualified_name` (caches, stored graphs, edge dedup) see churn on rebuild — the contributor says `callers_of`/`children_of` showed no regressions, but that was verified on their repo, not in CI.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,9 @@
 
 - GitHub Copilot auto-detection now requires the Copilot extension and also
   recognizes the extension bundled with released VS Code installations.
+- C# methods with a non-generic return type are no longer named after that
+  type: `public async Task Foo()` indexed as `Task`, collapsing every such
+  method in a class onto one qualified name (#791).
 - Added a post-index Python import resolver using unique module suffixes, so
   `src/` layouts produce canonical `CALLS` and `TESTED_BY` edges while duplicate
   package candidates remain explicitly unresolved (#720).
--- a/code_review_graph/parser.py
+++ b/code_review_graph/parser.py
@@ -14051,6 +14051,16 @@ class CodeParser:
                 if child.type == "identifier":
                     return child.text.decode("utf-8", errors="replace")
             return None
+        # C#: unlike Java there is no type_identifier — a non-generic return
+        # type such as ``Task`` is itself an ``identifier``, so the generic
+        # loop below would return it instead of the method name. Read the
+        # ``name`` field; unusual shapes still fall through.
+        if language == "csharp" and kind == "function" and node.type in (
+            "method_declaration", "constructor_declaration",
+        ):
+            name_node = node.child_by_field_name("name")
+            if name_node is not None:
+                return name_node.text.decode("utf-8", errors="replace")
 
         if language == "cpp" and kind == "function":
             declarator = node.child_by_field_name("declarator")
--- a/tests/test_multilang.py
+++ b/tests/test_multilang.py
@@ -460,6 +460,42 @@ class TestCSharpParsing:
 @pytest.mark.skipif(
     not _has_csharp_parser(), reason="csharp tree-sitter grammar not installed",
 )
+class TestCSharpMethodNames:
+    """Regression tests for #791: a non-generic C# return type is itself an
+    ``identifier``, so ``async Task Foo()`` was named ``Task`` and every such
+    method in a class merged onto one ``qualified_name``.
+    """
+
+    def _parse(self, source: str, tmp_path):
+        p = tmp_path / "x.cs"
+        p.write_text(source, encoding="utf-8")
+        return CodeParser().parse_file(p)
+
+    def test_non_generic_return_type_is_not_the_name(self, tmp_path):
+        nodes, _ = self._parse(
+            "public class Suite {\n"
+            "    public async Task Should_do_thing() { }\n"
+            "    public async Task Should_do_other_thing() { }\n"
+            "    public async Task<int> Returns_generic() { return 1; }\n"
+            "    public void PlainVoid() { }\n"
+            "    public Suite() { }\n"
+            "}\n",
+            tmp_path,
+        )
+        funcs = [n for n in nodes if n.kind == "Function"]
+        assert {f.name for f in funcs} == {
+            "Should_do_thing",
+            "Should_do_other_thing",
+            "Returns_generic",
+            "PlainVoid",
+            "Suite",
+        }
+        assert len(funcs) == 5
+
+
+@pytest.mark.skipif(
+    not _has_csharp_parser(), reason="csharp tree-sitter grammar not installed",
+)
 class TestCSharpAttributes:
     """Regression tests for #295 (C# half): C# attributes use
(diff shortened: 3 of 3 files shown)


tirth8205:
Verified locally before merge: reproduced the bug on main (two 'public async Task' methods both indexed as 'Task', matching #791), confirmed the fix extracts names via the grammar's name field for method_declaration and constructor_declaration, and probed edge cases (explicit interface implementations, generic methods, expression-bodied members, constructors) with correct results. Revert-check confirmed the new TestCSharpMethodNames tests fail without the fix. Full suite: 2306 passed, 5 skipped, 2 xpassed. Ruff and mypy clean. The CI matrix never ran here due to first-contribution workflow gating, so verification was done locally by two independent review passes. Note for anyone updating: C# repos will see one-time qualified_name churn on the next rebuild since previously merged nodes split into distinct methods. That is the correctness fix working as intended.

aromal-a:
Fixed edged parser . C-flow . Work name/ $:dirpath 
exit-code : changeling
KR = MD;

Split-resolution/work-flow-tag , pip headline, commit basics and push graphQL

PR #808, opened 2026-08-01
fix(parser): extract Java method names from the name field (#804)

### LINKED ISSUE #804: Java _get_name still uses the first-identifier heuristic instead of the grammar name field
Noted during the #794 review (C# method names). The C# branch of _get_name now reads child_by_field_name('name') for method_declaration and constructor_declaration, which is grammar-accurate. The Java branch still walks for the first identifier, which is the same fragile pattern that produced the C# bug (a non-generic return type parsing as identifier). No known misnaming today because Java return types parse as type_identifier rather than identifier, but the name-field approach is strictly more robust and would make the two branches consistent. Low priority, hardening only.

arimu1:
## Linked issue

Fixes #804

## What & why

During the #794 review, C# `method_declaration` / `constructor_declaration` names were switched to the grammar `name` field via `child_by_field_name("name")`. The Java branch still walked for the first `identifier` child — the same fragile pattern that misnamed C# methods when a non-generic return type is itself an `identifier`.

Java return types usually parse as `type_identifier` today, so there is no known misnaming, but the name-field approach is more robust and keeps the Java and C# branches consistent. Also drops the redundant second Java-only `method_declaration` identifier walk further down `_get_name`.

## How it was tested

```bash
uv run pytest tests/test_multilang.py::TestJavaParsing tests/test_multilang.py::TestJavaMethodNames tests/test_multilang.py::TestCSharpMethodNames -v --tb=short -q
# 10 passed

uv run --extra dev ruff check code_review_graph/parser.py tests/test_multilang.py
# All checks passed
```

## Checklist

- [x] All tests pass: focused Java/C# name tests above
- [x] Linting passes: `uv run --extra dev ruff check code_review_graph/parser.py tests/test_multilang.py`
- [ ] Type checking passes: `uv run mypy code_review_graph/ --ignore-missing-imports --no-strict-optional` (not re-run; change is a short name-field read with no new types)
- [x] Docs updated where behavior changed (CHANGELOG.md)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -71,6 +71,8 @@
 - C# methods with a non-generic return type are no longer named after that
   type: `public async Task Foo()` indexed as `Task`, collapsing every such
   method in a class onto one qualified name (#791).
+- Java method and constructor names now use the grammar `name` field (same
+  approach as C#), instead of walking for the first `identifier` child (#804).
 - Added a post-index Python import resolver using unique module suffixes, so
   `src/` layouts produce canonical `CALLS` and `TESTED_BY` edges while duplicate
   package candidates remain explicitly unresolved (#720).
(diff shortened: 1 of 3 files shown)


tirth8205:
Verified the removed fallback walk was unreachable and that C# output is unchanged. Full suite, ruff and mypy are clean on a merge with current main. Merging.

PR #809, opened 2026-08-01
fix(cli): detect-changes/visualize/wiki/watch no longer create empty graph.db

### LINKED ISSUE #803: Remaining commands still create an empty graph.db on first run; released-version poisoned graphs are not auto-healed (follow-up to #777 / #782)
PR #782 (merged) made status read-only on missing graphs and made update force a full rebuild when the graph is missing or empty. Two verified gaps remain from that review:

1. detect-changes, visualize, wiki, and watch still create an empty graph.db (with full schema) on a repo with no graph, the same side-effect class #777 reported for status. Mitigated today because update self-heals empty graphs, but the commands should resolve the DB path read-only like status now does.

2. Graphs already poisoned by released versions are not auto-healed: the new has_nodes guard only repairs missing or zero-node graphs. A graph that got a partial diff-only build with a synced head (non-empty but incomplete) still updates incrementally forever. Affected users need a documented recovery (build --force) or a heuristic (for example, indexed-file count much lower than tracked-file count) to trigger a rebuild.

arimu1:
## Summary

Follow-up to #777 / #782: `status` already resolved the DB path read-only when the graph was missing, but **`detect-changes`**, **`visualize`**, **`wiki`**, and **`watch`** still called `get_db_path()` (create) + `GraphStore()` and could materialize an empty schema `graph.db`. Later incremental `update`s then only re-parsed changed files, leaving the graph incomplete.

### Changes
- Treat those commands like `status`: resolve DB path with `read_only=True` (or explicit `--data-dir` without registry/mkdir side effects).
- Exit with `No graph found … Run code-review-graph build first.` when the DB is missing (same message as `status` / `dead-code` / `forget`).
- Keep the default-path legacy `.code-review-graph.db` migration for these commands.
- `visualize` / `wiki` still write exports next to an explicit `--data-dir` when a graph is present, without forcing a registry write when the graph is absent.
- Document recovery for already-poisoned graphs: full `code-review-graph build` (see `docs/TROUBLESHOOTING.md`). Auto-heal heuristics for incomplete non-empty graphs left out of scope (issue gap 2).

## Test plan

- [x] `uv run pytest tests/test_cli_reconciliation.py tests/test_cli.py::test_visualize_json_uses_local_export tests/test_cli.py::TestWatchInteraction -k "detect_changes or missing_graph or read_only" -q`
- [x] Parametrized tests: missing graph for status/detect-changes/visualize/wiki/watch does **not** create data dir or empty `graph.db`
- [x] Explicit `--data-dir` for visualize/wiki/watch/status is read-only when missing
- [x] Existing status / visualize / watch regressions still green
- [x] `ruff check` on touched Python files

Fixes #803

### DIFF
--- a/code_review_graph/cli.py
+++ b/code_review_graph/cli.py
@@ -1603,21 +1603,31 @@ def main() -> None:
         "wiki",
         "dead-code",
     )
-    status_data_dir = (
-        args.command == "status" and bool(getattr(args, "data_dir", None))
+    # Read-only consumers must not create graph.db / data dirs / registry
+    # entries when the graph is missing (follow-up to #777 / #782; see #803).
+    _read_only_db_cmds = frozenset({
+        "status",
+        "detect-changes",
+        "visualize",
+        "wiki",
+        "watch",
+    })
+    explicit_data_dir = bool(getattr(args, "data_dir", None))
+    read_only_explicit_data_dir = (
+        args.command in _read_only_db_cmds and explicit_data_dir
     )
-    if args.command in _data_dir_cmds and not status_data_dir:
+    if args.command in _data_dir_cmds and not read_only_explicit_data_dir:
         _handle_data_dir_option(args, repo_root)
 
-    if args.command == "status":
-        if status_data_dir:
+    if args.command in _read_only_db_cmds:
+        if read_only_explicit_data_dir:
             db_path = Path(args.data_dir).expanduser().resolve() / "graph.db"
         else:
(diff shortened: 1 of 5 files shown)


tirth8205:
Verified the new tests fail on main (the watch case hangs there since it used to create the DB and start looping) and that legacy DB migration and --data-dir export placement still behave. Merging. Gap 2 of #803 is not covered here, so I will reopen #803 for that part after the auto close.

PR #818, opened 2026-08-04
feat(parser): parse plain HTML — inline <script> blocks + DocSection anchors (#521)

### LINKED ISSUE #521: Add HTML parser: extract inline <script> functions + id-anchored DocSection nodes
## Motivation

Hand-authored single-page HTML documents — design briefs, HLDs, READMEs rendered to HTML, internal architecture pages — frequently mix substantial prose content with a small inline `<script>` block driving navigation/search/scroll-tracking, plus a `<style>` block for layout.

Today CRG sees these as opaque files (a single `File` node, nothing else). The inline JS functions are invisible to `semantic_search`, and there's no way to ask "where is section §17 in this 6 KLOC doc" without falling back to grep.

This proposal adds first-class HTML support that's symmetric to the existing Vue/Svelte SFC handling.

## Proposal

Two extraction concerns, both already proven by `_parse_vue` / `_parse_svelte`:

### 1. Inline `<script>` blocks → delegate to JS parser
Identical pattern to `_parse_vue`: walk the HTML tree, find `script_element` nodes, parse their `raw_text` with the JavaScript grammar, propagate the resulting `Function` / `Class` / `CALLS` nodes/edges with `line_start` / `line_end` offset to position within the .html file.

### 2. Id-anchored elements → new `DocSection` node kind
For each `<div id="…">` / `<section id="…">` / `<article id="…">`, emit a `DocSection` node with the element's line range and `html_id` in `extra`. This is the navigation primitive a sidebar typically links to — making it searchable lets queries like *"where does the cover section start"* or *"which page has the search-tabs anchor"* resolve in one `semantic_search` call.

### Deliberately out of scope
- `<style>` block extraction. No CSS grammar plumbed in CRG yet; a line-range-only stub would have negligible value. Easy follow-up if/when a CSS extractor lands.
- Heading-anchor sections (`<h2 id="…">`). Common in GitHub-rendered READMEs but pollutes the index when nested inside `<div id="…">` doc sections. Could be added with an opt-in if there's demand.

## Reference implementation

Branch: [`feat/html-parser-support`](https://github.com/dannymn/code-review-graph/tree/feat/html-parser-support) on my fork.

3-file diff, +298 LOC:
- `code_review_graph/parser.py`: 2 lines added to `EXTENSION_TO_LANGUAGE`, 3-line dispatch in `parse_bytes`, ~140-line `_parse_html` method modeled on `_parse_vue`.
- `tests/test_parser.py`: 7 new tests in the existing `TestCodeParser` class.
- `tests/fixtures/sample_html.html`: one fixture exercising every code path.

Tests (all pass; full `test_parser.py` 110/110):
- `test_detect_language_html` (.html and .htm)
- `test_parse_html_file` — File node + Function + Class extracted from `<script>`
- `test_parse_html_doc_sections` — `#cover`, `#s1`, `#s2`, `#appendix` emitted
- `test_parse_html_line_numbers_offset` — script line numbers reflect .html position
- `test_parse_html_nodes_have_html_language` — every node tagged `language="html"`
- `test_parse_html_no_script` — DocSection emitted even without any script
- `test_parse_html_only_id_anchored_sections` — plain `<div>` without id excluded

Field-validated locally: indexed a 6 KLOC HTML (an HLD) and a 3 KLOC reference HTML (a design brief), found 60+ JS functions with accurate line ranges, 39 DocSections, all surfaced by `semantic_search` queries.

## Questions for the maintainer before opening a PR

1. **Node-kind naming.** `DocSection` is intentionally generic — could later cover Markdown headings or other doc formats. Would you prefer `HTMLSection` to keep it format-scoped? Or skip the new kind entirely and reuse `Class` / `File`?
2. **Tag whitelist for sections.** Currently `div` / `section` / `article` with `id`. Worth widening to any element with `id`? Or narrowing to just `section` / `article`?
3. **`<style>` handling.** Drop entirely (current proposal), emit a line-range-only `Stylesheet` stub, or wait until a CSS extractor lands?
4. **Heading-anchor sections** (`<h2 id="x">`). Opt-in flag, always-on, or out of scope?

Happy to adjust the patch to match your preferences before submitting a PR. Local CRG using this patch has been running for ~1 day with zero regressions across 869 indexed files.

joseedson18jc:
## Linked issue

Closes #521

## What & why

Implements plain-HTML parsing per the maintainer's guidance on #521 (credit to @dannymn for the proposal and reference design — this is a fresh implementation rebased on current main, since the reference branch was noted as 341 commits behind).

**Inline `<script>` blocks** — same delegation pattern as `_parse_vue`/`_parse_svelte`:
- Iterative AST walk (script elements sit at arbitrary depth in real documents, and deeply nested pages must not hit the recursion limit)
- `<script src=…>` references and non-JS payloads (`application/json`, `importmap`, `text/x-template`, …) are skipped; an absent/empty `type` and `module`/`text/javascript`-family types parse as JavaScript
- Unquoted (`type=module`) and case-varied attributes handled
- Line numbers offset back to the containing `.html` file
- Relative imports inside `type="module"` blocks resolve from the HTML file's directory (source-side gate only; `.html` is deliberately **not** added as an import *target*, which would not be valid ES semantics)

**`DocSection` anchors** — per your integration requirements on the issue:
- `div`/`section`/`article` with an `id` → `DocSection` node (`extra.html_id`, full element line range)
- `CONTAINS` edge from the File node to each `DocSection`, so `children_of` and structural traversal reach them
- Anchor lookup only — no prose-content search claims; only the id is indexed
- Duplicate ids: first occurrence wins (`getElementById` semantics)
- Style blocks and heading anchors stay out of scope, as requested

Field validation on a real codebase: a 246 KB single-file dashboard (`index.html`, previously a bare `File` node) now yields 169 functions with accurate line ranges; semantic queries like "save the api key to browser storage" resolve to the correct function at rank 1.

## How it was tested

```bash
# 2417 passed, 1 skipped, 2 xpassed in 32.40s

# All checks passed!

```

15 new tests in `tests/test_multilang.py::TestHtmlEmbeddedScripts` cover the regression list from the issue discussion: multiple and nested scripts, external (`src=`) and empty scripts, malformed HTML (unclosed element), duplicate ids, unquoted and case-varied attributes, `.htm` detection, line-number offsets, import/call edges across blocks, and `CONTAINS` reachability of every `DocSection`.

## Checklist

- [x] Docs updated where behavior changed (README language coverage ×2, CHANGELOG)


dannymn:
Original reporter of #521 here. Thanks for picking this up, and for rebasing rather than trying to salvage my branch — that was the right call.

Some field validation, since I've been running an out-of-tree variant of this design against a real repo since June (currently CRG 2.3.7, 2,065 files, 64,747 nodes):

- 16 HTML files produce **165 DocSection nodes + 313 Function nodes** from inline `<script>` blocks.
- The largest single document goes from one opaque `File` node to **45 DocSections + 51 Functions**. Before the parser existed, semantic search over generated HLD documents returned nothing at all; now it resolves to the right section.

On your requirement #1 (`CONTAINS` edges from the file to each `DocSection`) — worth stating that this is not a hypothetical concern. My variant omits those edges, and the consequence is measurable: **0 of 25,957 DocSection nodes in my graph have a single incoming edge.** They are reachable by direct search only; `children_of`, structural context, and architecture traversal all walk straight past them. Emitting `CONTAINS` is the difference between "searchable" and "part of the graph." Good catch, and I'm glad it's in the PR rather than deferred.

Your requirement #2 is also right to insist on. Anchor-`id` indexing is what this delivers; prose-content retrieval is a different feature and shouldn't be implied by it.

One gap, purely as a follow-up and explicitly not a request to widen this PR: this covers `.html` only. My variant also emits one `DocSection` per ATX heading for `.md` (fenced-code aware), and in practice that is where most of the value has been — 447 markdown files produce 25,792 DocSections, versus 165 from HTML. Design docs, specs, and plans are usually markdown, and they're currently invisible to the graph for the same reason HTML was. Happy to open a separate issue for that if it's of interest, so this PR can stay scoped.

joseedson18jc:
Thanks @dannymn for the thorough field validation — glad the `CONTAINS` edges landed as intended.

@tirth8205 — this is ready for your review. Functional checks are green (pytest: 2417 passed, ruff, mypy); only GitGuardian ran automatically because the Actions workflows on a fork PR need maintainer approval to execute. Could you approve the workflow run and take a look when you have a moment? Happy to address anything before merge.

PR #821, opened 2026-08-04
fix(prompts,hints): align MCP tool names and refactor guidance

TheJeje20:
# Align MCP prompts and hints with the exported tool surface

## Summary

This PR updates the guidance layer used by prompts and hints so it matches the public MCP tools exposed by the server.

It also makes the refactor workflow easier to follow by explicitly suggesting symbol search when planning a rename.

## What changed

- Replaced bare tool names in prompts and hint routing with the public MCP names ending in `_tool`.
- Kept `dead_code` as the explicit refactor mode for dead-code checks.
- Added `semantic_search_nodes_tool` as the follow-up step for “find related symbols to also rename”.
- Kept `find_large_functions_tool` as a separate size/complexity check.
- Updated tests to cover the revised prompt text and hint suggestions.

## Scope and intent

This does not change the MCP API itself.

The change is limited to how the workflows are described and which follow-up actions are suggested:
- dead-code detection still uses `refactor_tool(mode="dead_code")`
- rename planning now explicitly points to `semantic_search_nodes_tool`
- large-function checks remain a separate step

## Validation

- `pytest tests/test_prompts.py tests/test_hints.py -q`
- `56 passed` for the refactor/hints path
- `57 passed` after the prompt update


tirth8205:
The alignment itself checks out. I verified every tool name now referenced in hints.py, prompts.py, skills.py and the bundled SKILL.md files against the @mcp.tool() registrations in main.py, and they all exist, including apply_refactor_tool. The reworked refactor hint routing behaves correctly, and your updated tests fail on main without the fix, so the coverage is real.

Two things before this can land:

1. ruff fails on this branch: hints.py lines 21 and 30 are 110 chars, over the 100-char limit, and CI runs ruff check code_review_graph/. Wrap those set literals.
2. tools/review.py line 119: you renamed the first two entries of next_tool_suggestions in get_review_context but left get_impact_radius bare. It should be get_impact_radius_tool.

While you are in there, tests/test_hints.py lines 68 and 74 also exceed 100 chars. Note this branch contains all of #779, so once this lands #779 should be closed, and expect a rebase against #822 which touches the same skills templates.

tirth8205:
Integrated on `integration/token-efficiency-hardening` as `316827a`. Only alignment with registered public MCP tool names is retained, including the three minimal review suggestions and the names recorded for later hints. An AST-based regression checks recommendations against actual registrations; extra workflow steps and already-present skill changes were omitted, while the branch-base propagation from #847 was preserved.

tirth8205:
Merged into `staging` through the integration PR #988 (merge commit 3030ca8). Your change is carried by commit 316827a, which keeps you as the author (TheJeje20). It will be promoted staging → testing → main and released from there. Thank you for the contribution! Closing this PR since its content is now on staging.

PR #826, opened 2026-08-06
fix(cli): render Unicode output with legacy encodings

Bortlesboat:
## Summary

On Windows and redirected streams configured as CP1252, the CLI could raise `UnicodeEncodeError` before showing its banner or graph summaries because those views use box-drawing characters. Reconfigurable standard streams now switch to UTF-8 at startup, while UTF-8, encoding-less, and non-reconfigurable streams keep their existing behavior.

## Validation

- Reproduced the crash on clean `main` with a CP1252 `TextIOWrapper` before applying the fix.
- The real `main()` regression passes on Python 3.10 and 3.13 and verifies both `stdout` and `stderr`.
- `tests/test_cli.py`: 18 passed; the one deselected test is an existing Windows failure reproduced on clean `main`.
- Full Windows suite: 2,353 passed; 37 unrelated platform or fixture failures, with representative cases reproduced on clean `main`.
- Ruff, targeted mypy, and the configured Bandit scan pass.
- Real processes launched with `PYTHONIOENCODING=cp1252` render both the startup banner and graph summaries and exit successfully.

## Post-Deploy Monitoring & Validation

No additional operational monitoring is required because this changes local CLI output and has no deployed service surface. During review and the next Windows release, a healthy run exits successfully and renders readable Unicode when `PYTHONIOENCODING=cp1252`; `UnicodeEncodeError`, `charmap`, or unreadable redirected output are failure signals. Revert this commit to restore the previous stream behavior if those signals appear.

---

[Compound Engineering](https://github.com/EveryInc/compound-engineering-plugin)

### DIFF
--- a/code_review_graph/cli.py
+++ b/code_review_graph/cli.py
@@ -98,6 +98,21 @@ def _supports_color() -> bool:
     return sys.stdout.isatty()
 
 
+def _configure_utf8_stdio() -> None:
+    """Allow Unicode CLI decoration on streams using a legacy encoding."""
+    for stream in (sys.stdout, sys.stderr):
+        encoding = getattr(stream, "encoding", None)
+        if not encoding or encoding.lower().replace("-", "") == "utf8":
+            continue
+        reconfigure = getattr(stream, "reconfigure", None)
+        if reconfigure is None:
+            continue
+        try:
+            reconfigure(encoding="utf-8")
+        except (OSError, ValueError):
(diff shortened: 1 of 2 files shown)


tirth8205:
Reproduced the cp1252 crash on main and verified the guard chain against closed, detached and encoding-less streams. Suite, ruff and mypy clean. Merging.

PR #833, opened 2026-08-08
fix(go): preserve generic receiver method ownership

### LINKED ISSUE #832: [Bug]: (go) generic receiver lose their type
### code-review-graph version

2.3.7

### Operating system

Linux

### Python version

3.12.4

### AI platform

claude-code

```shell
Nodes: 4382
Edges: 66825
Files: 840
Languages: hcl, go
Last updated: 2026-08-07T17:10:57
Built on branch: main
Built at commit: da3a115efef0
```

## Summary

Go methods declared on generic receiver types are parsed as top-level functions instead of methods on their receiver type.

## Reproduction

Terraform uses generic receivers in `DirectedGraph[T]`:

https://github.com/hashicorp/terraform/blob/da3a115efef0de2e817a8c3ef57c5df9297c99ce/internal/addrs/graph.go#L40-L75

For example:

```go
type DirectedGraph[T UniqueKeyer] struct{}

func (g DirectedGraph[T]) Add(addr T) {}

## Expected
DirectedGraph.Add
DirectedGraph.Has
DirectedGraph.Remove
DirectedGraph.TopologicalOrder

## Actual
Add: parent=None
Has: parent=None
Remove: parent=None
TopologicalOrder: parent=None

### Additional context

Impact

callers_of and callees_of cannot address generic methods by receiver 

type.Receiver method identity is lost for Go 1.18+ generic code.

In Terraform's graph.go, g.g.Add(...) can become a false self-edge:

merlincat11:
## Summary

Fixes #832.

Go methods declared on generic receiver types were emitted as top-level
functions because the receiver parser did not unwrap `generic_type` nodes.
The fix unwraps generic and pointer receiver nodes to their base type
identifier, preserving method ownership for both `T[P]` and `*T[P]` receivers.

## Validation

- `uv run --extra dev pytest -q` — 2399 passed, 5 skipped, 2 xpassed
- `uv run --extra dev ruff check code_review_graph/parser.py tests/test_multilang.py`
- Validated against Terraform's `internal/addrs/graph.go`: all 11
  `DirectedGraph[T]` methods now have `parent_name='DirectedGraph'`.

Repository-wide Ruff still reports unrelated pre-existing violations in other
test files.

### DIFF
--- a/code_review_graph/parser.py
+++ b/code_review_graph/parser.py
@@ -14697,31 +14697,25 @@ class CodeParser:
         """Extract the receiver type from a Go method_declaration.
 
         For ``func (s *T) Foo() {...}`` returns ``"T"``. For ``func (T) Foo()``
-        also returns ``"T"``. Returns None if no receiver is present.
-
-        The receiver is always the first ``parameter_list`` child of a
-        Go ``method_declaration`` and contains a single ``parameter_declaration``
-        whose type is either a ``type_identifier`` or a ``pointer_type``
-        wrapping one. See: #190
+        also returns ``"T"``. Generic receivers ``T[P]`` and ``*T[P]`` return
+        ``"T"``. Returns None if no receiver is present. See: #190, #832
         """
-        for child in node.children:
-            if child.type != "parameter_list":
-                continue
-            for param in child.children:
-                if param.type != "parameter_declaration":
-                    continue
-                for sub in param.children:
-                    if sub.type == "type_identifier":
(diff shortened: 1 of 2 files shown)


tirth8205:
Confirmed the new test fails on main and that T[P] and *T[P] receiver methods attach correctly, including multi param and unnamed receivers. Merging. Small follow-up: the unwrap in _get_go_receiver_type can land on a comment node, so func (b * /*c*/ T) F() loses its parent; filtering comment nodes there would cover it.

PR #835, opened 2026-08-08
fix(go): model and resolve embedded type relationships

### LINKED ISSUE #834: [Bug]: Go struct and interface embeddings never emit `INHERITS` edges
### code-review-graph version

2.3.7

### Operating system

Linux

### Python version

3.12.4

### AI platform

claude-code

```shell
## Summary

Go type embedding is currently invisible to the graph. Although `INHERITS` is CRG’s cross-language representation for Go embedding/composition, the parser emits no such edges for either embedded struct fields or embedded interfaces.

This causes false negatives in `inheritors_of`, impact-radius analysis, and other relationship traversals.
```

### Steps to reproduce

From the code-review-graph repository, run:

```bash
uv run python - <<'PY'
from pathlib import Path

source = b"""
package sample

type StructBase struct{}

type EmbeddedStruct struct {
    StructBase
}

type InterfaceBase interface {
    Method()
}

type EmbeddedInterface interface {
    InterfaceBase
}
"""

_, edges = CodeParser().parse_bytes(Path("sample.go"), source)

print([
    (edge.source, edge.target)
    for edge in edges
    if edge.kind == "INHERITS"
])
PY
```

### Actual result

```text
[]
```

### Expected result

Equivalent to:

```text
[
    ("sample.go::EmbeddedStruct", "StructBase"),
    ("sample.go::EmbeddedInterface", "InterfaceBase"),
]
```

### Additional context

A full graph was built for HashiCorp Terraform at commit `da3a115efef0de2e817a8c3ef57c5df9297c99ce`, covering all 1,993 Go files.

The corpus contains:

- 258 embedded struct fields
- 82 embedded interfaces
- 340 total embedding sites across 307 type declarations
- 306 embedding sites in non-test code

The current Go `_get_bases` implementation extracted zero bases from these declarations.

Examples include:

- [`GraphNodeConfigResource` embeds `dag.Vertex`](https://github.com/hashicorp/terraform/blob/da3a115efef0de2e817a8c3ef57c5df9297c99ce/internal/terraform/node_resource_abstract.go#L30-L33)
- [`GraphNodeAttachResourceConfig` embeds `GraphNodeConfigResource`](https://github.com/hashicorp/terraform/blob/da3a115efef0de2e817a8c3ef57c5df9297c99ce/internal/terraform/transform_attach_config_resource.go#L14-L18)
- [`TolerantVertex` embeds `Vertex`](https://github.com/hashicorp/terraform/blob/da3a115efef0de2e817a8c3ef57c5df9297c99ce/internal/dag/graph.go#L28-L34)

Despite those declarations, `inheritors_of` returns zero results for both `Vertex` and `GraphNodeConfigResource`.

## Root cause

The Go branch of `_get_bases` expects embedded types to be direct children of `field_declaration_list`:

```python
for f in field_node.children:
    if f.type == "type_identifier":
        bases.append(...)
```

Tree-sitter’s Go AST uses different shapes:

```text
struct_type
└── field_declaration_list
    └── field_declaration
        └── type_identifier / qualified_type / generic_type
```

Interfaces do not use `field_declaration_list`:

```text
interface_type
└── type_elem
    └── type_identifier / qualified_type / generic_type
```

Therefore, the current condition cannot match either form.

merlincat11:
Closes #834

## Summary

- model each Go `type_spec` or alias as its own graph node, including grouped declarations
- extract struct and interface embedding candidates from the actual tree-sitter Go AST
- resolve same-package and imported embedding targets to canonical graph nodes using package/import and `go.mod` evidence
- exclude `_test.go` declarations from imported-package resolution so external test packages cannot shadow production types
- keep interface `type_elem` relationships as `INHERITS` only when they resolve to interfaces; represent concrete or unresolved constraint terms as `REFERENCES`
- scope import-edge loading to files with embedding candidates and batch edge reconciliation
- remove unused file-level Go package metadata and document the expanded `REFERENCES` semantics
- force a one-time full rebuild for graphs created with the legacy Go type model

## Root cause

The parser treated a whole `type_declaration` as one type and expected embedded identifiers at the wrong AST depth. Its raw relationship targets also lacked enough package evidence for graph queries, while a single interface `type_elem` is syntactically ambiguous between interface composition and a concrete type-set term. Imported resolution additionally treated declarations from external `_test.go` packages as normal package exports, creating false ambiguity when production and test packages defined the same type name.

## Impact

Go embedding and composition now participate in canonical graph traversal without making false inheritance claims for resolved concrete constraints. Grouped declarations and aliases are no longer silently omitted. Qualified imports resolve to production declarations even when a same-directory external test package shadows the type name.

On Terraform main at `da3a115efef0de2e817a8c3ef57c5df9297c99ce`:

- parsed all 1,993 Go files
- `inheritors_of Vertex` returns all 12 embedding sites, including the nine formerly disconnected `dag.Vertex` relationships
- `inheritors_of GraphNodeConfigResource` returns all five embedding sites
- all remaining raw qualified embedding targets belong to standard-library or third-party packages absent from the repository graph

## Validation

- `uv run pytest -q` — 2,402 passed, 5 skipped, 2 xpassed
- focused Go parser/resolver tests — 11 passed
- `uv run ruff check code_review_graph/ tests/test_multilang.py`
- full Terraform graph build with minimal post-processing — 4,314 files, 28,918 parsed nodes, 314,844 parsed edges
- Terraform queries after rebuild — 12 `Vertex` inheritors and 5 `GraphNodeConfigResource` inheritors

### DIFF
--- a/code_review_graph/graph.py
+++ b/code_review_graph/graph.py
@@ -743,6 +743,235 @@ class GraphStore:
         """
         return self._resolve_bare_endpoints("CALLS", "target_qualified")
 
+    def resolve_go_embedding_targets(self) -> int:
+        """Resolve Go embeddings and reject concrete interface type terms.
+
+        Struct fields are unambiguous embeddings. An interface ``type_elem``
+        can instead be a concrete type-set term, so it remains inheritance only
+        when its target resolves to an interface. Repository-local imports are
+        mapped through their module's ``go.mod`` to canonical graph nodes.
+        """
+        edges = self._conn.execute(
+            "SELECT e.id, e.kind, e.source_qualified, e.target_qualified, "
+            "e.file_path, e.extra "
+            "FROM edges e JOIN nodes s ON s.qualified_name = e.source_qualified "
+            "WHERE s.language = 'go' AND e.kind IN ('INHERITS', 'REFERENCES') "
+            "AND e.extra LIKE '%\"go_embedding_target\"%'",
+        ).fetchall()
+        if not edges:
+            return 0
+
+        def load_extra(raw: str | None) -> dict:
+            try:
+                value = json.loads(raw or "{}")
+            except (TypeError, json.JSONDecodeError):
+                return {}
+            return value if isinstance(value, dict) else {}
+
+        by_qualified: dict[str, dict[str, Any]] = {}
+        by_location: dict[tuple[str, str, str], list[dict[str, Any]]] = {}
+        by_directory_name: dict[tuple[str, str], list[dict[str, Any]]] = {}
+        candidate_dirs: set[str] = set()
+        for row in self._conn.execute(
+            "SELECT name, qualified_name, file_path, extra FROM nodes "
+            "WHERE language = 'go' AND kind = 'Class'",
+        ).fetchall():
+            extra = load_extra(row["extra"])
+            directory = normalize_file_path(str(Path(row["file_path"]).parent.resolve()))
+            candidate = {
+                "name": row["name"],
+                "qualified_name": row["qualified_name"],
+                "file_path": row["file_path"],
+                "directory": directory,
+                "package": extra.get("go_package", ""),
+                "extra": extra,
+            }
+            by_qualified[row["qualified_name"]] = candidate
+            by_location.setdefault(
+                (directory, candidate["package"], row["name"]), [],
+            ).append(candidate)
+            by_directory_name.setdefault(
+                (directory, row["name"]), [],
+            ).append(candidate)
+            candidate_dirs.add(directory)
+
+        imports_by_file: dict[str, list[str]] = {}
+        for row in self._conn.execute(
+            "SELECT file_path, target_qualified FROM edges "
+            "WHERE kind = 'IMPORTS_FROM'",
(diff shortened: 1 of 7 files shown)


tirth8205:
Solid modeling work and the new tests are good, but the directory-suffix fallback in import_directory (graph.py:847-866) resolves imports that go.mod proves are external. Repro: module example.com/project, a local internal/dag defining Vertex, and a consumer importing github.com/external/lib/internal/dag as dag with type Embeds interface { dag.Vertex }. The fallback matches the internal/dag suffix and emits a canonical INHERITS edge to the local Vertex. That is the same false-inheritance class this PR sets out to remove, and it will bite repos with layouts like pkg/errors or pkg/api. Please gate the fallback so it runs only when the go.mod walk found no module directive at all. The walk also reads go.mod files above the repo root; bound it.

Minor: at parser.py:13455 the alias not in ("_", ".") guard never fires, since blank and dot imports parse as blank_identifier and dot nodes and the alias has already fallen back to the path's last segment. Check the name node instead.

Suite, ruff, and mypy are green on the merge, and all four new tests fail on main as expected.

PR #837, opened 2026-08-10
Fix affected flow path normalization

MrMfer:
## Problem

`analyze_changes()` passes `changed_files` directly to
`get_affected_flows()`.

Git reports changed files relative to `repo_root`, while graph nodes
store normalized repository-rooted paths. This causes affected-flow
analysis to miss impacted flows when the changed file paths are
repository-relative.

## Fix

Normalize `changed_files` against `repo_root` before passing them to
`get_affected_flows()`.

Absolute paths continue to work because `Path` preserves an already
absolute path when joined with the repository root.

## Regression test

Added a regression test covering the case where:

- Git reports `src/services.py`
- graph nodes store `/fake/repo/src/services.py`

The test verifies that the affected `handler` flow is correctly
identified.

## Validation

- `tests/test_changes.py` + `tests/test_flows.py`: 73 passed
- `git diff upstream/main...HEAD --check`: clean

The broader test suite currently contains unrelated platform /
environment-dependent failures, including Windows symlink privilege
requirements and missing async pytest support.

### DIFF
--- a/code_review_graph/changes.py
+++ b/code_review_graph/changes.py
@@ -464,7 +464,18 @@ def analyze_changes(
     overall_risk = max((nr["risk_score"] for nr in node_risks), default=0.0)
 
     # Affected flows.
-    affected = get_affected_flows(store, changed_files)
+    # Git reports changed files relative to repo_root, while graph nodes
+    # store normalized paths rooted at the repository. Normalize the paths
+    # here so flow impact analysis uses the same identity as the graph.
+    flow_changed_files = changed_files
+    if repo_root is not None:
+        root_path = Path(repo_root)
+        flow_changed_files = [
+            normalize_file_path(root_path / path)
+            for path in changed_files
+        ]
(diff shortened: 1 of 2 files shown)


tirth8205:
Closing this as a duplicate of #852, which fixes the same bug (#848) more completely.

The difference is scope. This patch normalizes paths only for the get_affected_flows call, but analyze_changes also matches changed_files against nodes.file_path in the no-ranges fallback (changes.py, the store.get_nodes_by_file loop). With an empty ranges dict, say a binary-only diff, relative CLI paths still miss every node there: flows come back, but changed functions, test gaps, and risk all stay at zero. #852 remaps changed_files once at the top, mirroring the existing changed_ranges remap, so both call sites see the same identity. I verified your regression test also passes under #852.

The test here is good and fails on main as expected. If you see a case #852 misses, please comment on that PR.

PR #838, opened 2026-08-11
fix(parser): handle IF NOT EXISTS in SQL table references

### LINKED ISSUE #820: SQL: CREATE TABLE IF NOT EXISTS indexes the table as "IF" — real table name lost, and dedup drops every later CREATE in the file
### Summary

`_SQL_TABLE_RE` does not allow for `IF NOT EXISTS` between `CREATE TABLE` / `CREATE VIEW` and the object name:

https://github.com/tirth8205/code-review-graph/blob/main/code_review_graph/parser.py#L53-L57

```python
_SQL_TABLE_RE = re.compile(
    r"(?:FROM|JOIN|INTO|CREATE\s+(?:OR\s+REPLACE\s+)?(?:TABLE|VIEW)|INSERT\s+OVERWRITE)"
    r"\s+((?:`[^`]+`|\w+)(?:\.(?:`[^`]+`|\w+))*)",
    re.IGNORECASE,
)
```

So `CREATE TABLE IF NOT EXISTS clientes (...)` captures the literal `IF` as the table name. `IF` isn't in `_SQL_KEYWORDS` ([L441-L447](https://github.com/tirth8205/code-review-graph/blob/main/code_review_graph/parser.py#L441-L447)), so it passes the filter and becomes an `IMPORTS_FROM` target.

There is a second-order effect that makes it worse than one bogus edge: the pass dedups on `seen_refs` ([L4438](https://github.com/tirth8205/code-review-graph/blob/main/code_review_graph/parser.py#L4438)), so within a file only the *first* `IF` is recorded. Every subsequent `CREATE TABLE IF NOT EXISTS` is silently dropped. A schema file written entirely in the idempotent `IF NOT EXISTS` style therefore contributes exactly one edge — pointing at `IF` — and none of its actual tables.

### Reproduction

Regex in isolation:

```python
import re
R = re.compile(
    r"(?:FROM|JOIN|INTO|CREATE\s+(?:OR\s+REPLACE\s+)?(?:TABLE|VIEW)|INSERT\s+OVERWRITE)"
    r"\s+((?:`[^`]+`|\w+)(?:\.(?:`[^`]+`|\w+))*)",
    re.IGNORECASE,
)

R.findall("CREATE TABLE IF NOT EXISTS clientes (id INT);")
# ['IF']            expected: ['clientes']

R.findall("CREATE OR REPLACE VIEW IF NOT EXISTS v_x AS SELECT 1;")
# ['IF']            expected: ['v_x']

R.findall("CREATE TABLE clientes (id INT);")
# ['clientes']      correct
```

End to end, on a real repo (crg 2.3.7, Python 3.12) whose `.sql` schema files use `CREATE TABLE IF NOT EXISTS`, the stored edges look like:

```
('.../01_base_schema.sql',      'clientes')
('.../02_auditoria.sql',        'IF')
('.../03_planos_templates.sql', 'IF')
('.../04_advanced_management.sql', 'IF')
```

`IF` also becomes a shared edge target across files, so unrelated schema files end up looking structurally coupled through a node that doesn't exist.

### Expected

`CREATE TABLE IF NOT EXISTS clientes` should yield `clientes`, same as `CREATE TABLE clientes`.

### Suggested fix

Make the clause optional in the pattern:

```python
_SQL_TABLE_RE = re.compile(
    r"(?:FROM|JOIN|INTO|CREATE\s+(?:OR\s+REPLACE\s+)?(?:TABLE|VIEW)"
    r"(?:\s+IF\s+NOT\s+EXISTS)?|INSERT\s+OVERWRITE)"
    r"\s+((?:`[^`]+`|\w+)(?:\.(?:`[^`]+`|\w+))*)",
    re.IGNORECASE,
)
```

Adding `IF`, `NOT` and `EXISTS` to `_SQL_KEYWORDS` is worth doing regardless, as defence in depth — it would at least turn this into a missing edge rather than a wrong one, and it also covers `DROP TABLE IF EXISTS` style statements if the pattern ever grows to match `DROP`.

Two adjacent gaps in the same pattern, if you want them in the same change (happy to split them out):

- `CREATE TEMPORARY TABLE` / `CREATE TEMP TABLE` / `CREATE UNLOGGED TABLE` / `CREATE GLOBAL TEMPORARY TABLE` aren't matched.
- `CREATE MATERIALIZED VIEW` isn't matched (only `TABLE` and `VIEW`).

### Notes

Found while investigating #819 — same repo, unrelated code path. Both surfaced because the language mix in that project (procedural PHP + raw `.sql` schema files) isn't represented in the benchmark corpus.

### Environment

- code-review-graph 2.3.7 (PyPI)
- Python 3.12.3, Linux x86_64

yzxcj797:
# Pull Request

## Linked issue

Closes #820

## What & why

`_SQL_TABLE_RE` currently captures the literal `IF` from `CREATE TABLE IF NOT EXISTS ...`. Because references are deduplicated, later idempotent creates in the same file can then disappear as well.

This change:

- allows an optional `IF NOT EXISTS` clause when extracting `CREATE TABLE` and `CREATE VIEW` targets
- classifies `IF`, `NOT`, and `EXISTS` as SQL keywords as a defensive fallback
- adds regex and end-to-end regressions proving ordinary creates still work and multiple idempotent creates retain distinct edges

The adjacent TEMP/UNLOGGED/MATERIALIZED gaps described in the issue are intentionally left out to keep this PR atomic.

## How it was tested

```bash
uv run pytest tests/test_notebook.py::TestSqlTableExtraction tests/test_multilang.py::TestSQLParsing --tb=short -q
# 21 passed

# passed

uv run mypy code_review_graph/parser.py --ignore-missing-imports --no-strict-optional
# passed
```

Broader local runs also completed 481 tests in the two touched test modules; the remaining failures were existing Windows path-separator assertions. Whole-project mypy reached an existing Windows-only `SIGKILL` error in `daemon_cli.py`.

## AI assistance

OpenAI Codex was used for repository inspection, implementation and test drafting, local verification, and review of the resulting diff.

## Checklist

- [ ] All tests pass: `uv run pytest tests/ --tb=short -q` (targeted tests pass; full supported-platform suite was not available on Windows)
- [ ] Type checking passes: `uv run mypy code_review_graph/ --ignore-missing-imports --no-strict-optional` (the touched parser passes; see existing Windows `SIGKILL` failure above)
- [ ] Docs updated where behavior changed (no user-facing documentation change required)

### DIFF
--- a/code_review_graph/parser.py
+++ b/code_review_graph/parser.py
@@ -51,7 +51,8 @@ class CellInfo(NamedTuple):
 
 
 _SQL_TABLE_RE = re.compile(
-    r"(?:FROM|JOIN|INTO|CREATE\s+(?:OR\s+REPLACE\s+)?(?:TABLE|VIEW)|INSERT\s+OVERWRITE)"
+    r"(?:FROM|JOIN|INTO|CREATE\s+(?:OR\s+REPLACE\s+)?(?:TABLE|VIEW)"
+    r"(?:\s+IF\s+NOT\s+EXISTS)?|INSERT\s+OVERWRITE)"
     r"\s+((?:`[^`]+`|\w+)(?:\.(?:`[^`]+`|\w+))*)",
     re.IGNORECASE,
 )
@@ -443,7 +444,7 @@ _SQL_KEYWORDS: frozenset[str] = frozenset({
     "UNION", "INTERSECT", "EXCEPT", "AS", "ON", "USING", "SET",
     "VALUES", "DEFAULT", "NULL", "TRUE", "FALSE",
     "INNER", "OUTER", "LEFT", "RIGHT", "FULL", "CROSS", "NATURAL",
-    "LATERAL", "RECURSIVE", "ONLY", "WITH",
+    "LATERAL", "RECURSIVE", "ONLY", "WITH", "IF", "NOT", "EXISTS",
 })
 
 logger = logging.getLogger(__name__)
--- a/tests/test_multilang.py
+++ b/tests/test_multilang.py
@@ -3635,6 +3635,17 @@ class TestSQLParsing:
         targets = {e.target for e in imports}
         # active_orders view and archive procedure both reference orders/users
         assert "orders" in targets or "users" in targets
+
+    def test_multiple_if_not_exists_creates_keep_distinct_references(self):
+        _, edges = self.parser.parse_bytes(
+            Path("idempotent_schema.sql"),
+            b"CREATE TABLE IF NOT EXISTS customers (id INT);\n"
+            b"CREATE TABLE IF NOT EXISTS invoices (id INT);\n",
+        )
+        targets = [e.target for e in edges if e.kind == "IMPORTS_FROM"]
+        assert targets == ["customers", "invoices"]
+
+
 class TestZigParsing:
     def setup_method(self):
         self.parser = CodeParser()
--- a/tests/test_notebook.py
+++ b/tests/test_notebook.py
@@ -131,13 +131,25 @@ class TestSqlTableExtraction:
 
     def test_create_table(self):
         matches = _SQL_TABLE_RE.findall("CREATE TABLE my_db.new_table (id INT)")
-        assert "my_db.new_table" in matches
+        assert matches == ["my_db.new_table"]
+
+    def test_create_table_if_not_exists(self):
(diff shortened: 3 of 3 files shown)


tirth8205:
Regex fix and keyword fallback both check out, the new tests fail without the fix, and the full suite is green on current main. Merging.

PR #844, opened 2026-08-12
fix(daemon): stop cleanly on Windows

### LINKED ISSUE #843: [Bug][Windows] crg-daemon stop always ends in AttributeError (signal.SIGKILL missing): stale PID file, and restart never restarts
## Summary

On Windows, `crg-daemon stop` kills the daemon but then always ends in an uncaught
`AttributeError: module 'signal' has no attribute 'SIGKILL'`. Because the exception escapes
`_handle_stop`, the `clear_pid()` call on `daemon_cli.py:97` never runs and a stale PID file is
left in `~/.code-review-graph/daemon.pid`. Exit code is 1 on every stop.

`crg-daemon restart` is broken as a consequence: `_handle_restart` calls `_handle_stop` first
(`daemon_cli.py:106`), the `AttributeError` propagates out, and `_handle_start` on
`daemon_cli.py:110` is never reached. The daemon stops and does not come back.

Affected: 2.3.7 (verified against the sdist) and 2.3.6. Windows only.

## Root cause

`_handle_stop` sends SIGTERM (`daemon_cli.py:73`), then polls for the process to exit:

```python
# daemon_cli.py:83-95 (2.3.7)
for _ in range(50):
    try:
        os.kill(pid, 0)
    except ProcessLookupError:
        break
    time.sleep(0.1)
else:
    # Still alive after 5s - send SIGKILL
    print("Daemon did not stop gracefully, sending SIGKILL...")
    try:
        os.kill(pid, signal.SIGKILL)
    except ProcessLookupError:
        pass

clear_pid()
```

Two POSIX assumptions break on Windows:

1. **`os.kill(pid, 0)` never raises `ProcessLookupError` on Windows.** Signal `0` is equal to
   `signal.CTRL_C_EVENT`, so `os.kill` takes the `GenerateConsoleCtrlEvent` branch instead of an
   existence check. For an already-exited PID it returns `None`, so the loop never breaks and
   always runs the full 5 seconds into the `else:`.
2. **`signal.SIGKILL` does not exist on Windows**, so line 93 raises `AttributeError`. Only
   `ProcessLookupError` is caught there, so it escapes `_handle_stop` and `main()`, taking
   `clear_pid()` on line 97 with it.

This is the same `os.kill(pid, 0)` pitfall that #511 reported for `daemon status`. The fix for
that landed as `pid_alive()` in `daemon.py:400`, whose docstring documents this exact behaviour,
but `_handle_stop` still calls `os.kill` directly and never picked it up.

## Reproduction

`crg-daemon start`, then `crg-daemon stop` on Windows.

- Observed: daemon process dies, 5s pause, `Daemon did not stop gracefully, sending SIGKILL...`,
  then the `AttributeError` traceback. `daemon.pid` still on disk, exit code 1.
- Expected: `Daemon stopped.`, PID file removed, exit code 0.

The two platform facts in isolation (Python 3.13.2, Windows 11):

```
>>> hasattr(signal, "SIGKILL")
False
>>> signal.CTRL_C_EVENT
0
# child process spawned, killed, and reaped first:
>>> os.kill(dead_pid, 0)
None                      # returns, does not raise ProcessLookupError
```

Replaying the `daemon_cli.py:83-95` loop verbatim against that same dead PID:

```
loop ran to exhaustion in 5.02s -> else: branch
now: os.kill(pid, signal.SIGKILL)
ESCAPED with AttributeError: module 'signal' has no attribute 'SIGKILL'
-> clear_pid() on daemon_cli.py:97 is never reached
```

For contrast, the shipped helper answers correctly on the same PID:
`daemon.pid_alive(dead_pid)` -> `False`, `daemon.pid_alive(os.getpid())` -> `True`.

## Suggested fix

In `_handle_stop`:

- use the existing `pid_alive()` from `daemon.py` for the wait loop instead of
  `os.kill(pid, 0)` (it already handles Windows via `OpenProcess`/`WaitForSingleObject`);
- guard the escalation, e.g. `getattr(signal, "SIGKILL", signal.SIGTERM)`. On Windows `os.kill`
  with any signal other than `CTRL_C_EVENT`/`CTRL_BREAK_EVENT` routes to `TerminateProcess`
  anyway, so SIGTERM is a correct escalation there;
- run `clear_pid()` on the failure paths too (`try/finally`), so a failed escalation cannot leave
  the PID file behind.

Happy to test a patch on Windows if useful.

## Environment

- OS: Windows 11
- Python 3.13.2
- code-review-graph 2.3.7 (source verified) / 2.3.6 (installed)

y4ho0:
## Linked issue

Closes #843

## What & why

`crg-daemon stop` polled process state with `os.kill(pid, 0)` and escalated with `signal.SIGKILL`. Both assumptions are POSIX-specific: on Windows the liveness probe does not reliably report an exited process, and `SIGKILL` is unavailable.

The resulting `AttributeError` also skipped PID cleanup and prevented `crg-daemon restart` from starting the daemon again.

This change:

- reuses the existing cross-platform `pid_alive()` helper while waiting for shutdown
- uses `SIGKILL` when available and falls back to `SIGTERM` on platforms without it
- clears the daemon PID in a `finally` block so a failed forced stop cannot leave stale state
- adds mocked Windows regressions for forced stop, restart after exit, and PID cleanup after an escalation failure

## How it was tested

```bash
uv run --python 3.13 pytest tests/ --tb=short -q
# 2401 passed, 5 skipped, 2 xpassed

uv run pytest tests/test_daemon.py --tb=short -q
# 69 passed

uv run ruff check code_review_graph/ tests/test_daemon.py
# All checks passed

bandit -r code_review_graph/ -c pyproject.toml
# No issues identified
```

## Checklist

- [x] Tests added for the corrected behavior
- [x] All tests pass
- [x] Linting passes
- [x] Type checking passes
- [x] Documentation is not required because the CLI interface and expected behavior are unchanged

### DIFF
--- a/code_review_graph/daemon_cli.py
+++ b/code_review_graph/daemon_cli.py
@@ -57,7 +57,7 @@ def _handle_start(args: argparse.Namespace) -> None:
 
 def _handle_stop(_args: argparse.Namespace) -> None:
     """Stop the running daemon process."""
-    from .daemon import clear_pid, is_daemon_running, read_pid
+    from .daemon import clear_pid, is_daemon_running, pid_alive, read_pid
 
     if not is_daemon_running():
         print("Daemon is not running.")
@@ -79,22 +79,22 @@ def _handle_stop(_args: argparse.Namespace) -> None:
         print(f"Error: Permission denied sending signal to PID {pid}.")
         sys.exit(1)
 
-    # Wait up to 5 seconds for process to die
-    for _ in range(50):
-        try:
-            os.kill(pid, 0)
-        except ProcessLookupError:
-            break
-        time.sleep(0.1)
-    else:
-        # Still alive after 5s — send SIGKILL
-        print("Daemon did not stop gracefully, sending SIGKILL...")
-        try:
-            os.kill(pid, signal.SIGKILL)
-        except ProcessLookupError:
-            pass
+    try:
+        # Wait up to 5 seconds for process to die.
+        for _ in range(50):
+            if not pid_alive(pid):
+                break
+            time.sleep(0.1)
+        else:
+            print("Daemon did not stop gracefully, force-stopping...")
+            force_signal = getattr(signal, "SIGKILL", signal.SIGTERM)
+            try:
+                os.kill(pid, force_signal)
+            except ProcessLookupError:
+                pass
+    finally:
+        clear_pid()
 
-    clear_pid()
     print("Daemon stopped.")
 
 
--- a/tests/test_daemon.py
+++ b/tests/test_daemon.py
@@ -1029,6 +1029,83 @@ class TestDaemonCLI:
 
         assert exc_info.value.code == 1
 
+    def test_handle_stop_windows_uses_sigterm_for_forced_stop(self):
+        """Windows falls back to SIGTERM when SIGKILL is unavailable."""
+        from code_review_graph.daemon_cli import _handle_stop
+
+        args = MagicMock()
+        pid = 4242
+        windows_signal = MagicMock(spec=["SIGTERM"])
+        windows_signal.SIGTERM = signal.SIGTERM
+
+        with (
+            patch("code_review_graph.daemon.is_daemon_running", return_value=True),
+            patch("code_review_graph.daemon.read_pid", return_value=pid),
+            patch("code_review_graph.daemon.pid_alive", return_value=True) as mock_alive,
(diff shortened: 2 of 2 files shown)


tirth8205:
All gates pass and the three new tests fail on main without the fix. pid_alive keeps POSIX semantics and the finally block covers the stale PID case from #843. Merging. This supersedes #859.

PR #847, opened 2026-08-13
fix(review): use merge-base for branch diffs

### LINKED ISSUE #842: [Bug]: detect-changes --base <branch> uses a two-dot diff, misattributing base-branch commits to the reviewed branch
>Note from me (human): 
>- Issue generated with Opus 5 in Claude Code via T3 Code
>- We created our custom code-review skill, and faced with this problem, so had to adjust our skill (see screenshot in the end, it might help with understanding the problem)

### Summary

`detect-changes --base <ref>` compares the working tree against `<ref>` with a **two-dot** diff (`git diff <ref> HEAD`) rather than a three-dot one (`git diff <ref>...HEAD`). When the checked-out branch has not been rebased onto the tip of `<ref>`, every commit `<ref>` gained since the branch point is reported as part of the change set.

The docs recommend `--base origin/<base-branch>` as the primary invocation (`docs/GITHUB_ACTION.md:15` and `:185`), which makes this the default way users hit the problem.

### Why the GitHub Action doesn't show it

On `pull_request`, `actions/checkout` checks out `refs/pull/N/merge` — a synthetic merge of the PR into the current base. That commit already contains the base branch, so two-dot and three-dot diffs coincide and the results look correct.

Any workflow that checks out `refs/pull/N/head` (the author's branch tip, as GitHub shows under "Files changed"), or reviews a long-lived local branch, diverges.

### Reproduction

In a repo whose base branch has moved on since the branch point:

```bash
git diff --name-only origin/main...HEAD | wc -l   # 6   — actual branch scope
git diff --name-only origin/main HEAD | wc -l     # 194 — what detect-changes sees

code-review-graph detect-changes --base origin/main
#   194 files, risk 0.85, symbols from unrelated branches

code-review-graph detect-changes --base "$(git merge-base origin/main HEAD)"
#   6 files, 30 functions, risk 0.50 — correct
```

Observed with v2.3.7.

### Impact

Risk scores are inflated (0.50 → 0.85 above) and the reported symbols include code the branch never touched, so downstream consumers (`get_review_context`, `get_affected_flows`, agent-driven reviews) analyse and comment on unrelated changes.

### Suggested fix

When `--base` resolves to a ref rather than an explicit commit, diff against `git merge-base <ref> HEAD` — matching the three-dot semantics GitHub itself uses for "Files changed". Options, roughly in order of preference:

1. Use merge-base by default; add `--no-merge-base` for the current behaviour if anyone depends on it.
2. Keep the default and add an opt-in flag (e.g. `--three-dot` / `--merge-base`).
3. At minimum, document the two-dot semantics in `docs/GITHUB_ACTION.md` alongside the `--base "$(git merge-base origin/main HEAD)"` workaround, and note that it matters when checking out `refs/pull/N/head`.

> Note from me (human):
>- The skill diff:
> Image
> ...
> Image
> ...
> Image

yzxcj797:
Fixes #842.

## Summary

- resolve local and remote branch bases to their merge base with `HEAD`, matching pull-request review semantics
- preserve explicit commit IDs, tags, and revision expressions as exact comparison points, with the previous behavior retained when a merge base cannot be resolved
- apply the resolved base consistently across CLI, MCP tools, minimal context, prompts, changed-file discovery, and line-range analysis
- add a real divergent Git history regression plus caller and fallback coverage

## Testing

- `uv run pytest tests/test_integration_git.py tests/test_changes.py tests/test_cli.py tests/test_prompts.py tests/test_tools.py -q` - 108 passed, 1 skipped
- `uv run mypy code_review_graph/cli.py code_review_graph/incremental.py code_review_graph/prompts.py code_review_graph/tools/context.py code_review_graph/tools/query.py code_review_graph/tools/review.py --ignore-missing-imports --no-strict-optional --follow-imports=skip`
- `uv run bandit -q -r code_review_graph/`
- `git diff --check`

The complete Windows suite also ran: 2,355 passed, 14 skipped, 2 xpassed, with 37 pre-existing Windows path/symlink/D3-fixture failures unrelated to this change.

## AI assistance

OpenAI Codex assisted with issue analysis, implementation, tests, and review. I reviewed the final diff and verified the commands and results above locally.

### DIFF
--- a/code_review_graph/cli.py
+++ b/code_review_graph/cli.py
@@ -985,7 +985,11 @@ def main() -> None:
         help="Analyze change impact against the existing graph (read-only). "
              "Does NOT re-parse files — for that, use 'update --brief'.",
     )
-    detect_cmd.add_argument("--base", default="HEAD~1", help="Git diff base (default: HEAD~1)")
+    detect_cmd.add_argument(
+        "--base",
+        default="HEAD~1",
+        help="Git diff base (branch refs use their merge base with HEAD; default: HEAD~1)",
+    )
     detect_cmd.add_argument(
         "--brief",
         action="store_true",
@@ -1977,9 +1981,9 @@ def main() -> None:
                 attach_context_savings,
                 estimate_file_tokens,
             )
-            from .incremental import get_changed_files, get_staged_and_unstaged
+            from .incremental import get_changed_files, get_staged_and_unstaged, resolve_review_base
 
-            base = args.base
+            base = resolve_review_base(repo_root, args.base)
             changed = get_changed_files(repo_root, base)
             if not changed:
                 changed = get_staged_and_unstaged(repo_root)
--- a/code_review_graph/incremental.py
+++ b/code_review_graph/incremental.py
@@ -667,6 +667,64 @@ def resolve_incremental_base(repo_root: Path, store: "GraphStore") -> str | None
     return None
 
 
+def resolve_review_base(repo_root: Path, base: str) -> str:
+    """Resolve a branch-like Git review base to its common ancestor with HEAD.
+
+    ``git diff <branch>`` compares the two tips and therefore includes commits
+    made only on the base branch after the reviewed branch diverged.  GitHub's
+    "Files changed" view instead uses the merge base.  Preserve explicit
+    commit/revision inputs, but translate local and remote-tracking branch
+    refs to that common ancestor.
+
+    If ref detection or merge-base resolution fails (for example in a shallow
+    clone), return *base* unchanged so callers retain the existing diff
+    behaviour rather than silently reporting no changes.
+    """
+    if (
+        detect_vcs(repo_root) != "git"
+        or not base
+        or base.startswith("-")
+        or not _SAFE_GIT_REF.fullmatch(base)
+    ):
+        return base
+
+    try:
+        symbolic = subprocess.run(
+            ["git", "rev-parse", "--symbolic-full-name", "--verify", base],
+            capture_output=True,
+            text=True,
+            encoding="utf-8",
+            errors="replace",
+            cwd=str(repo_root),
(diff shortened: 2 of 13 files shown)


tirth8205:
The approach is right and matches the three-dot semantics asked for in #842. Tags, SHAs, and rev expressions stay exact, branch and remote-tracking refs get the merge base, and the fallbacks for shallow clones, unrelated histories, and missing refs behave sensibly. I also ran extra edge cases locally (detached HEAD, base equal to HEAD, base strictly ahead of HEAD, ambiguous tag plus branch names, dirty working tree) and the implementation held up on all of them.

One blocker: tests/test_tools.py test_review_base_is_shared_by_probe_discovery_and_analysis fails on macOS. tempfile.mkdtemp() returns /var/folders/..., a symlink to /private/var/..., and _get_store resolves the root, so resolve_review_base gets the resolved path and the assert against self.root does not match. Please change the three mock asserts in that test (resolve, has_changes, get_changed) to compare against self.root.resolve().

With that fixed the suite is green here, and ruff and mypy are clean. Happy to merge after that one change.

tirth8205:
Integrated on `integration/token-efficiency-hardening` as `f3bdb72`, `61300d8`. Branch and remote-branch reviews now use their merge base consistently, while tags, commit SHAs and revision expressions remain exact. A real divergent-Git regression confirms that `get_review_context` excludes a base-only file and retains the feature file; the rebase preserves current read-only behaviour and corrects canonical-path expectations on macOS.

tirth8205:
Merged into `staging` through the integration PR #988 (merge commit 3030ca8). Your change is carried by commit f3bdb72, which keeps you as the author (yzxcj797). It will be promoted staging → testing → main and released from there. Thank you for the contribution! Closing this PR since its content is now on staging.

PR #852, opened 2026-08-13
Remap relative changed_files to absolute paths in analyze_changes (fixes #848)

### LINKED ISSUE #848: [Bug]: CLI detect-changes --brief prints 0 affected flows where detect_changes_tool reports 208 on identical input
## Version

code-review-graph 2.3.7, macOS, Python 3.13, PHP/Laravel repository (~1,050 indexed files, graph built with default settings plus a .code-review-graphignore for vendored assets).

## What happens

On the same graph, same repo, same base ref:

CLI:

```
$ code-review-graph detect-changes --base <ref> --brief
Analyzed 86 changed file(s):
  - 342 changed function(s)/class(es)
  - 0 affected flow(s)
  ...
```

MCP tool, identical input:

```python
detect_changes_tool(base="<ref>", detail_level="minimal")
# summary: "Analyzed 86 changed file(s): ... 208 affected flow(s) ..."
```

Cross-check: `get_affected_flows_tool(changed_files=[the same 75 PHP files])` returns 206 flows, and `flows` lists 50 flows for the repo, so flow detection itself works. Only the CLI `--brief` path loses them.

Reproduced on both a 2-file diff (CLI: 0, MCP: 2) and an 86-file diff (CLI: 0, MCP: 208).

## Why it matters

The brief panel is what the pre-commit hook and the daily review flow print, so flow impact silently disappears exactly where most users read it.

## Expected

CLI `detect-changes --brief` and `detect_changes_tool` report the same affected-flow count for identical input.

asemraza:
Fixes #848

## Problem

`analyze_changes` remaps `changed_ranges` keys to absolute paths, but passes `changed_files` through raw. The affected-flows lookup (`get_node_ids_by_files`, exact `IN (...)` match) and the no-ranges node fallback both compare those repo-relative CLI paths against the absolute `nodes.file_path`, so CLI `detect-changes` printed `0 affected flow(s)` while `detect_changes_tool` (which remaps before calling) reported 208 flows on identical input.

## Fix

Remap `changed_files` in `analyze_changes` exactly the way `changed_ranges` keys are already remapped. Already-absolute inputs pass through pathlib joining unchanged, so the MCP path stays untouched.

## Test

New regression test stores nodes under absolute paths (as real builds do), passes relative `changed_files` plus `repo_root` (as the CLI does), and asserts the flow is found. Verified it fails on main and passes with the fix. Full suite: 2,399 passed, ruff and mypy clean.

### DIFF
--- a/code_review_graph/changes.py
+++ b/code_review_graph/changes.py
@@ -420,6 +420,19 @@ def analyze_changes(
             for key, ranges in parse_diff_ranges(repo_root, base).items()
         }
 
+    # The affected-flows lookup and the no-ranges fallback match
+    # changed_files against nodes.file_path, which stores absolute
+    # normalized paths. CLI callers pass repo-relative diff paths, so an
+    # exact IN (...) match finds no nodes and detect-changes reports
+    # "0 affected flow(s)" even when the MCP tool reports hundreds on the
+    # same input (#848). Remap the same way as changed_ranges keys;
+    # already-absolute inputs (MCP) pass through pathlib joining unchanged.
+    if repo_root is not None:
+        _root = Path(repo_root)
+        changed_files = [
+            normalize_file_path(_root / fp) for fp in changed_files
+        ]
+
     # Map changes to nodes.
     if changed_ranges:
         changed_nodes = map_changes_to_nodes(store, changed_ranges)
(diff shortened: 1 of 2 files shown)


tirth8205:
Verified the regression test fails on main and that the early remap also covers the no-ranges node fallback that #837 misses. Full suite, ruff and mypy clean on a merge with current main. Merging.

PR #853, opened 2026-08-13
Add detail_level and max_flows to get_affected_flows (fixes #849)

### LINKED ISSUE #849: get_affected_flows_tool has no detail_level or limit parameter and returns 247k+ tokens, breaking the documented token budget
## Version

code-review-graph 2.3.7, PHP/Laravel repository (~1,050 indexed files, 50 detected flows).

## What happens

`get_affected_flows_tool(changed_files=[75 files])` returns a response measuring 247,479 tokens (cl100k_base). Its signature only accepts `changed_files`, `base` and `repo_root`; it is the only review tool in the chain without `detail_level`, and it has no max-results parameter either.

## Why it matters

`LLM-OPTIMIZED-REFERENCE.md` prescribes the review-pr workflow as `detect_changes_tool -> get_affected_flows_tool -> structured review` with a target of "<=5 tool calls, <=800 tokens total context". Step 2 of that prescribed workflow can alone exceed the budget by 300x. Every other tool in the chain honors `detail_level="minimal"`, so agents following the documented convention get an unbounded response exactly once, in the middle of the recommended flow.

## Expected

- `detail_level="minimal"` support returning flow names, criticality and counts only, like the other review tools, and/or
- a `max_flows` parameter with a sane default.

## Workaround

We removed the tool from our MCP `--tools` allowlist and rely on the affected-flow count from `detect_changes_tool` plus `list_flows_tool` for identity.

asemraza:
Fixes #849

## Problem

`get_affected_flows_tool` is step 2 of the documented review-pr workflow with its <=800-token target, but it was the only tool in the chain without `detail_level` and had no result bound. Each flow embeds a full `steps` list, so a 75-file change set produced a 247k-token response.

## Change

- `detail_level="minimal"`: per-flow metadata only (id, name, criticality, depth, node_count, file_count), same convention as the sibling tools.
- `max_flows` (default 50, 0 disables): bounds the returned list. `total` always reports the untruncated count and `truncated` flags the cut, mirroring the impact-radius truncation contract. The summary appends "showing N" when truncated.

Both parameters are wired through the MCP tool wrapper with updated docstrings.

## Tests

Three new tests: minimal drops steps/path, max_flows truncates with correct total/truncated/summary, max_flows=0 disables the limit. Full suite 2,401 passed, ruff and mypy clean.

### DIFF
--- a/code_review_graph/main.py
+++ b/code_review_graph/main.py
@@ -530,6 +530,8 @@ def get_affected_flows_tool(
     changed_files: Optional[list[str]] = None,
     base: str = "HEAD~1",
     repo_root: Optional[str] = None,
+    detail_level: str = "standard",
+    max_flows: int = 50,
 ) -> dict:
     """Find execution flows affected by changed files.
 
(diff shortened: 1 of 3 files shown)


tirth8205:
Matches the sibling tool conventions and the truncation contract, and the 60 flow default cap case checks out. Merging. Follow-up noted in #866: the no changed files early return in tools/review.py omits the new truncated key.

PR #855, opened 2026-08-14
fix(parser): track keyword argument callback references

### LINKED ISSUE #840: dead_code false positive: keyword-argument callback refs (argparse set_defaults(func=...)) — gap left open by #363/#424
## Description

`dead_code`/`suggest` still reports false positives for functions passed as **keyword-argument** callbacks — the exact gap [PR #424](https://github.com/tirth8205/code-review-graph/pull/424) (closing #363) explicitly scoped out:

> **What this does not fix**
> - Keyword-argument callbacks (`executor.submit(callable=h)`) — `keyword_argument` is a different child shape; out of scope for this PR

The most common real-world instance of this shape is Python's `argparse` CLI-dispatch idiom, which is extremely widespread:

```python
def cmd_telemetry(args):
    ...

def build_parser():
    sp = subparsers.add_parser("telemetry")
    sp.set_defaults(func=cmd_telemetry)   # <-- keyword_argument, never gets a REFERENCES edge
```

`cmd_telemetry` is genuinely alive (invoked later via `args.func(args)`), but because the reference is `func=cmd_telemetry` (a `keyword_argument` node) rather than a bare positional identifier or `argument_list` member, `_ref_from_arguments`'s PR #424 fix never fires for it, and it gets flagged dead.

## Impact

On CLI-heavy Python codebases (any repo using `argparse.set_defaults(func=...)`, or dict/kwargs-based command-dispatch tables in general — FastAPI dependency wiring, click-style registries, etc.), this produces a very high false-positive rate: essentially every CLI subcommand handler function is flagged, since its only reference is a keyword argument in its own file's parser-construction code. This matches what's reported independently in #363's original description and its comment thread (the Golang "registry pattern" comment there is the same shape in a different language).

## Suggested fix

Mirror PR #424's own approach one level further: extend `_ref_from_arguments` (or add a sibling `_ref_from_keyword_argument`) to walk a `keyword_argument` node's **value** child the same way `_ref_from_arguments` already walks bare identifiers in `argument_list`/`arguments` — gated through the same `_emit_reference_if_known` check PR #424 uses, so it stays scoped to locally-defined/imported names and doesn't add noise.

## Environment

- `code-review-graph` 2.3.6 (installed via `uv`)
- Python 3.12
- MCP stdio transport

## Note

Ran into this independently: fixing a similar false-positive class in a small homegrown Python dead-code scanner (same root cause — only counting cross-file references, missing same-file dispatch-table call sites) is what led me to check whether `code-review-graph` had the same gap. It does, and PR #424's own PR description already named it as the next piece of scope, so this should be a fairly contained follow-up to that work.

lamchun1110:
## Summary

- extract Python callback references from `keyword_argument` values
- reuse the existing known-definition/import guard to avoid noisy edges
- add regression coverage for `set_defaults(func=handler)` and dead-code detection

Fixes #840

## Testing

- `uv run pytest tests/test_parser.py -q --disable-warnings --maxfail=1` (163 passed)
- `uv run pytest tests/ --tb=short -q` (2399 passed, 5 skipped, 2 xpassed)
- `uv run ruff check code_review_graph/parser.py`

### DIFF
--- a/code_review_graph/parser.py
+++ b/code_review_graph/parser.py
@@ -12193,13 +12193,17 @@ class CodeParser:
     ) -> None:
         """Extract REFERENCES from identifier arguments (callbacks)."""
         for ch in args_node.children:
-            if ch.type == "identifier":
-                name = ch.text.decode("utf-8", errors="replace")
-                self._emit_reference_if_known(
-                    name, language, file_path, caller, edges,
-                    import_map, defined_names,
-                    line=ch.start_point[0] + 1,
-                )
+            reference_node = ch
+            if ch.type == "keyword_argument" and language == "python":
+                reference_node = ch.child_by_field_name("value")
+            if reference_node is None or reference_node.type != "identifier":
+                continue
+            name = reference_node.text.decode("utf-8", errors="replace")
+            self._emit_reference_if_known(
+                name, language, file_path, caller, edges,
+                import_map, defined_names,
+                line=reference_node.start_point[0] + 1,
+            )
 
     def _extract_solidity_constructs(
         self,
--- a/tests/fixtures/sample_callback_refs.py
+++ b/tests/fixtures/sample_callback_refs.py
@@ -1,8 +1,8 @@
 """Fixture for issue #363: function references in callback positions.
 
-Each `*_callback` function is passed as a bare-identifier argument to
-another call. They are never invoked with parens, so without REFERENCES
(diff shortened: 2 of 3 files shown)


tirth8205:
Verified the new tests fail without the parser change and that builtins and keyword names get no spurious edges. Merging. Bound method callbacks like cb=self.on_event are still not covered, consistent with the scope cut in #424.

PR #860, opened 2026-08-18
fix(incremental): reconcile reverted file content

### LINKED ISSUE #817: Incremental update leaves stale nodes when a file is edited then reverted to match base (empty diff, never reparsed)
## Summary

`incremental_update()` discovers changed files via `git diff <base> --name-only` (default base `HEAD~1`). If a file is edited and then reverted to content identical to `<base>` within the same base window (no intervening commit), the diff for that file is empty, so it is never re-parsed. Any nodes/edges created by the intermediate (now-reverted) state are never purged — they persist as permanent ghost nodes until a `full_rebuild`, silently diverging the incremental graph from a full rebuild.

- **Version:** 2.3.6 (also present in 2.3.7 source, same `get_changed_files`/purge logic as far as I can tell from the changelog)
- **OS:** Windows 11
- **Found via:** live repro during normal Claude Code agent use (PostToolUse hook calling `code-review-graph update --skip-flows` after every Edit/Write/Bash)

## Root cause

`incremental_update()`'s purge/reparse set is built from `git diff <base> --name-only` (or `--name-status`) against `<base>`. When working-tree content for a file is byte-identical to `<base>` — because it was edited and then reverted — that file does not appear in the diff at all. It is therefore excluded from both the reparse set and the purge set, so nodes/edges added during the intermediate edited state are never revisited or removed.

This is adjacent to #684 / #742 (rename leaves stale nodes — diff reports only one side) and #812 (mid-update failure leaves orphan nodes no full-rebuild purges) — same family of bug (incremental's changed-file discovery misses a case, purge loop only acts on what's in that set), different trigger. Neither issue's fix covers this case: #684's fix (emit both old+new path for renames) doesn't help because there's no rename here, and the closed/unmerged #694/#769 (anchor base to last-synced-commit) doesn't help either because the problem isn't which commit is the base — it's that content matches whatever the base is, so the diff is empty regardless of which base is chosen.

## Steps to reproduce

1. Build/update the graph for a repo containing a tracked file, e.g. `foo.py` with no function `bar`.
2. Edit `foo.py`, add `def bar(): return 1`. Run `code-review-graph update` (or let a per-edit hook do it). Confirm `bar` is now a node (`semantic_search_nodes` / `file_summary`).
3. Edit `foo.py` again, removing the `bar` function you just added — file content now matches `<base>` (e.g. `HEAD~1`) exactly. Run `code-review-graph update` again.
4. Query the graph for `bar` (`semantic_search_nodes_tool` or `file_summary` on `foo.py`).

## Expected vs actual

- **Expected:** `bar` node is gone — `update` converges to the same state as a `full_rebuild`.
- **Actual:** `bar` node still present. `git diff <base> -- foo.py` is empty (file matches base again), so the file is skipped entirely by the reparse/purge logic. Node stays until an unrelated later change makes that file's diff non-empty again, or until a manual `full_rebuild`.

## Suggested fix

`incremental_update()`'s changed-file discovery should not be the sole signal for "does this file need reparsing." Either:
- Track a per-file content hash (already partially done per #797 "store the hash of the bytes actually parsed") at last-parse time, and compare current on-disk hash against it independent of git diff, so a file that round-trips back to a previously-seen-but-graph-still-has-old-node state is still caught; or
- Track a per-file "graph state generation" separate from git history, so any file whose current on-disk content doesn't match what's currently represented in the graph (regardless of git diff outcome) gets reparsed.

Confirmed the ghost node clears with a manual `full_rebuild`, so this is a purge-completeness bug in the incremental path specifically, not a parser bug.

## Impact

Silent — no error, no warning. Any MCP/CLI query (`semantic_search_nodes`, `callers_of`, `get_impact_radius`, `file_summary`) can report phantom functions/classes that no longer exist in the source, which is the exact failure mode this tool is meant to prevent when used for review/impact-radius decisions. Hit within minutes of normal agent use with the standard PostToolUse hook pattern (`update` after every Edit/Write/Bash) — this is not an edge case, it's a routine edit-then-undo cycle.

yzxcj797:
## Problem

An incremental update discovers work solely from the VCS diff. If a file is edited, indexed, and then reverted to bytes identical to the diff base, that file disappears from the next diff. The graph keeps nodes and edges from the intermediate content until an unrelated event causes that path to be reparsed.

This also happens when another file keeps the diff non-empty: the reverted file is simply absent from the changed-path set and is never considered.

## Fix

- Expose the file hashes already stored on `File` nodes.
- Before dispatching incremental work, compare those hashes with current file bytes and merge mismatched paths into the changed-file set.
- The existing per-file hash check and atomic replacement then purge the intermediate nodes and replace them with the reverted content.

Hash comparison is independent of git history, so it works regardless of which valid diff base the caller selects.

## Tests

- Added a direct edit/revert regression: an intermediate function is indexed, then removed after an otherwise-empty update.
- Added a mixed-path regression: a reverted file is reconciled even when another changed file keeps the update non-empty.
- Verified the first regression fails on `main` and passes with this change.

Closes #817

### DIFF
--- a/code_review_graph/graph.py
+++ b/code_review_graph/graph.py
@@ -1190,6 +1190,16 @@ class GraphStore:
         ).fetchall()
         return [r["file_path"] for r in rows]
 
+    def get_file_hashes(self) -> dict[str, str]:
+        """Return indexed file paths mapped to the hash last stored for each file."""
+        rows = self._conn.execute(
+            "SELECT file_path, file_hash FROM nodes WHERE kind = 'File'"
+        ).fetchall()
+        return {
+            normalize_file_path(row["file_path"]): row["file_hash"] or ""
+            for row in rows
+        }
+
     def search_nodes(self, query: str, limit: int = 20) -> list[GraphNode]:
         """Keyword search across node names.
 
--- a/code_review_graph/incremental.py
+++ b/code_review_graph/incremental.py
@@ -708,7 +708,10 @@ def get_changed_files(repo_root: Path, base: str = "HEAD~1") -> list[str]:
         return []
 
 
-def _find_content_mismatches(repo_root: Path, store: "GraphStore") -> list[str]:
+def _find_content_mismatches(
+    repo_root: Path,
+    store: "GraphStore",
+) -> tuple[list[str], dict[str, str]]:
     """Find indexed files whose bytes differ from the graph's last parsed hash.
 
     Git diffs compare the working tree with a base commit.  If a file is
@@ -719,6 +722,7 @@ def _find_content_mismatches(repo_root: Path, store: "GraphStore") -> list[str]:
     git diff non-empty.
     """
     mismatched_files: list[str] = []
+    current_hashes: dict[str, str] = {}
 
     for stored_path, stored_hash in store.get_file_hashes().items():
         path = Path(stored_path)
@@ -726,7 +730,9 @@ def _find_content_mismatches(repo_root: Path, store: "GraphStore") -> list[str]:
             path = repo_root / path
         try:
             relative_path = path.relative_to(repo_root).as_posix()
-            current_hash = hashlib.sha256(path.read_bytes()).hexdigest()
+            raw = path.read_bytes()
+            current_hash = hashlib.sha256(raw).hexdigest()
+            current_hashes[relative_path] = current_hash
         except (OSError, ValueError):
             # Missing files are handled by stale-file reconciliation.  Paths
             # outside the repository cannot be represented as relative update
@@ -735,7 +741,7 @@ def _find_content_mismatches(repo_root: Path, store: "GraphStore") -> list[str]:
         if current_hash != stored_hash:
             mismatched_files.append(relative_path)
 
-    return mismatched_files
+    return mismatched_files, current_hashes
 
 
 def _get_svn_changed_files(repo_root: Path, rev_range: str | None = None) -> list[str]:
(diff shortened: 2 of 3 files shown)


tirth8205:
The fix itself is sound and the regressions are valid; I confirmed both new tests fail on main. My concern is where the scan runs. _find_content_mismatches (incremental.py:711) is invoked unconditionally at the top of incremental_update, so it stats, reads, and sha256s every indexed file on every call. On the CLI path _reconcile_stale_files already does a full scan, so there it roughly doubles read cost (103ms to 170ms no-op on a 2000-file repo here). Watch mode is the real problem: batches call incremental_update with reconcile_stale=False precisely to stay proportional to events, and this adds a full-repo read to every debounced batch (0ms to 62ms at 2000 files, linear in repo bytes). Watch does not need the scan: a revert generates its own event, so the file is already in the batch and the existing hash check covers it. Please gate the scan on reconcile_stale or a dedicated flag. Optional: pass the computed hashes forward so mismatched files are not read and hashed again in the quick-check loop.

yzxcj797:
Thanks for the benchmark. I gated the content scan on `reconcile_stale`, so watch batches keep their existing event-bounded behavior. The hashes computed by that scan are now reused by the quick-check loop, avoiding a second read/hash for indexed files. I also added a regression test proving the scan is skipped when reconciliation is disabled.

tirth8205:
Integrated on `integration/token-efficiency-hardening` as `3510d45`. CLI reconciliation now detects edit-and-revert content that disappears from the VCS diff, while event-driven watch updates retain their bounded scan behaviour. The integration preserves main's file-marker inventory, and the edit/revert and watch-scan regressions passed.

tirth8205:
Merged into `staging` through the integration PR #988 (merge commit 3030ca8). Your change is carried by commit 3510d45, which keeps you as the author (yzxcj797). It will be promoted staging → testing → main and released from there. Thank you for the contribution! Closing this PR since its content is now on staging.

PR #861, opened 2026-08-18
fix(graph): purge orphan rows during reconciliation

### LINKED ISSUE #812: Files that fail mid-update leave permanent orphan nodes/edges — full rebuild never purges them
## Summary

When a `watch`-triggered update for a file errors out partway through, that file can end up with content nodes (`Function`, etc.) and/or `edges` rows in the graph DB, but **no corresponding `File`-kind node**. Because the stale-file purge in `full_build()` determines "does this file still exist" by querying only `kind='File'` rows, a file in this broken state becomes permanently invisible to cleanup — it survives every subsequent `build` (full rebuild) forever, even a clean rebuild with correct ignore rules.

## Root cause

`graph.py`:
```python
def get_all_files(self) -> list[str]:
    rows = self._conn.execute(
        "SELECT DISTINCT file_path FROM nodes WHERE kind = 'File'"
    ).fetchall()
    return [r["file_path"] for r in rows]
```

`incremental.py`'s `full_build()`:
```python
existing_files = set(store.get_all_files())          # only file_paths with a File-kind node
current_abs = {str(repo_root / f) for f in files}     # files from collect_all_files() (ignore-aware)
stale_files = existing_files - current_abs
for stale in stale_files:
    store.remove_file_data(stale)
```

If a file's `File`-kind node is missing (e.g. because an earlier watcher-triggered update for it errored mid-write), that file's path is absent from `existing_files` too — so it can never appear in `stale_files`, and `remove_file_data()` is never called for it, no matter how many times `build`/`update` runs afterward.

## How we hit it

We run `code-review-graph daemon` watching several worktrees of the same repo. One watcher's log shows a file being repeatedly ingested/removed as a generated `coverage/` directory got regenerated by test runs, then eventually failing mid-update:

```
INFO: Updated: intranet-frontend/coverage/sorter.js (16 nodes, 69 edges)
...
ERROR: Error updating /home/.../intranet-frontend/coverage/sorter.js: bad parameter or other API misuse
```

After that error, the graph DB for that repo had:

```sql
SELECT kind, name, file_path FROM nodes WHERE file_path LIKE '%coverage/sorter.js%';
-- Function | loadColumns  | .../coverage/sorter.js
-- Function | loadRowData  | .../coverage/sorter.js
-- Function | loadData     | .../coverage/sorter.js
-- (zero File-kind rows for this path)
```

We ran `code-review-graph build --repo <path>` **twice** afterward (a real full rebuild each time, confirmed by `updated_at` advancing and `collect_all_files()` correctly excluding this path both times, verified via `git ls-files` and direct Python testing). Both rebuilds preserved the exact same 3 orphaned `Function` nodes and their edges — the graph's `total_nodes`/`total_edges` were byte-identical across both rebuilds, differing from four sibling repos (same commit, same content) built at different times, which all converged to one identical, smaller number. This ruled out "parser non-determinism" as a possible cause and pointed at persistent DB state.

Only deleting the graph DB files outright and rebuilding from scratch fixed it:
```bash
rm .code-review-graph/graph.db .code-review-graph/graph.db-wal .code-review-graph/graph.db-shm
code-review-graph build --repo <path>
```
After that, `total_nodes` dropped to match the other repos exactly, and:
```sql
SELECT file_path FROM nodes WHERE file_path NOT IN (SELECT file_path FROM nodes WHERE kind='File');
-- (empty)
```

We found the same orphan pattern (this time in `edges`, not `nodes`) independently in two more of our worktrees, with `edges.file_path` values that had no matching `File`-kind node either — same root cause, same fix required (hard reset).

## Suggested fix

`get_all_files()` (or wherever `full_build()` sources its "does this file still exist in the graph" comparison set) should union distinct `file_path` values across **all** tables that carry one (`nodes`, `edges`, and anything else keyed by `file_path`), not just `nodes WHERE kind='File'`. That way any partially-written file — regardless of whether its `File` node specifically made it in — is still eligible for the stale-file purge on the next full build.

A secondary/defensive fix: make per-file updates (in both `watch()`'s handler and `full_build()`'s insertion path) transactional per-file, so a mid-write failure can't leave a file half-inserted (some node/edge rows committed, others — including the `File` node — not) in the first place.

## Environment

- `code-review-graph` 2.3.7
- Python 3.12.3, WSL2/Linux

Happy to share the exact `graph.db` (pre-hard-reset) if that'd help reproduce, or run additional queries against our setup.

yzxcj797:
## Linked issue

Fixes #812

## What changed

- Reconciliation now inventories file paths from all non-virtual node rows and edge rows, not only rows with `kind = 'File'`.
- Partially failed updates that leave Function/Class nodes or edge rows without their File marker are now visible as stale paths and can be purged by the next full build.
- Resolver-owned virtual nodes (notably Spring Event nodes) remain excluded from the filesystem inventory and continue to be managed by their resolver.
- Added a regression covering both an orphan Function node and an edge-only orphan path.

## Testing

- `python -m pytest tests/test_incremental.py -q -k without_a_file_node`
- `python -m pytest tests/test_graph.py tests/test_incremental.py tests/test_status_stats.py tests/test_windows_path_identity.py -q --deselect tests/test_incremental.py::TestWatchReconciliation::test_watch_symlink_events_are_rejected --deselect tests/test_incremental.py::TestWatchReconciliation::test_watch_rejects_file_below_symlinked_ancestor_outside_repo`
- `python -m ruff check code_review_graph/graph.py tests/test_incremental.py`
- `python -m mypy --ignore-missing-imports code_review_graph/graph.py tests/test_incremental.py`

The two deselected watch tests require Windows symlink privileges that are unavailable in this environment; the remaining 194 selected tests pass.

### DIFF
--- a/code_review_graph/graph.py
+++ b/code_review_graph/graph.py
@@ -1185,8 +1185,17 @@ class GraphStore:
         return resolved
 
     def get_all_files(self) -> list[str]:
+        # A partially failed file update can leave Function/Class nodes or
+        # edges behind after their File marker is gone. Reconciliation must see
+        # every represented path so those orphans can be purged, while explicit
+        # virtual nodes (such as Spring Event markers) remain outside the file
+        # inventory and are managed by their owning resolver.
         rows = self._conn.execute(
-            "SELECT DISTINCT file_path FROM nodes WHERE kind = 'File'"
+            "SELECT file_path FROM nodes "
+            "WHERE json_extract(extra, '$.virtual') IS NULL "
+            "UNION "
+            "SELECT file_path FROM edges "
+            "ORDER BY file_path"
         ).fetchall()
         return [r["file_path"] for r in rows]
 
--- a/tests/test_incremental.py
+++ b/tests/test_incremental.py
@@ -748,6 +748,50 @@ class TestFullBuild:
         finally:
             store.close()
 
+    def test_full_build_removes_nodes_left_without_a_file_node(self, tmp_path):
+        from code_review_graph.parser import EdgeInfo, NodeInfo
+
+        current = tmp_path / "sample.py"
+        current.write_text("def hello():\n    pass\n")
+        orphan_path = str(tmp_path / "generated" / "orphan.js")
+        orphan_edge_path = str(tmp_path / "generated" / "orphan-edge.js")
+        store = GraphStore(tmp_path / "test.db")
+        try:
+            store.upsert_node(
+                NodeInfo(
+                    kind="Function",
+                    name="loadData",
+                    file_path=orphan_path,
+                    line_start=1,
+                    line_end=2,
+                    language="javascript",
+                )
+            )
+            store.upsert_edge(
+                EdgeInfo(
+                    kind="CALLS",
+                    source=f"{orphan_edge_path}::caller",
+                    target=f"{orphan_edge_path}::target",
+                    file_path=orphan_edge_path,
+                )
+            )
+            store.commit()
+
+            with patch(
+                "code_review_graph.incremental.get_all_tracked_files",
+                return_value=["sample.py"],
+            ):
+                result = full_build(tmp_path, store)
+
+            assert result["stale_files_removed"] == 2
+            assert store.get_nodes_by_file(orphan_path) == []
+            assert store._conn.execute(
+                "SELECT COUNT(*) FROM edges WHERE file_path = ?",
+                (orphan_edge_path,),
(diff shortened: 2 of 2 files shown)


tirth8205:
Verified the new test fails without the graph.py change, virtual Spring Event nodes survive both reconciliation modes, and edge only orphans purge on full build and reconciling updates. Merging. One hardening follow-up filed as #864: a nodes row with malformed JSON in extra makes json_extract raise inside get_all_files.

PR #870, opened 2026-08-19
fix(graph): survive malformed extra JSON in get_all_files

### LINKED ISSUE #864: get_all_files: malformed JSON in nodes.extra crashes reconciliation after #861
Hardening follow-up noted while merging #861. get_all_files (code_review_graph/graph.py) now runs json_extract(extra, '$.virtual') over every non-File node row. A row whose extra column holds invalid JSON (empty string, truncated text from a killed writer) makes SQLite raise OperationalError 'malformed JSON', which crashes every get_all_files caller: build, status, visualize, watch.

Fix shape: guard with a json_valid(extra) check in the WHERE clause (or wrap per-row), and add a regression that inserts a node row with extra='' and asserts get_all_files still returns.

Second smaller point from the same review: the virtual exclusion keys on presence of the key, not truthiness, so extra='{"virtual": false}' is also hidden from the file inventory and becomes immortal to reconciliation. Compare with json_extract(...) = 1 instead.

yzxcj797:
## Summary

Follow-up to #861 (noted while merging it, #864): a killed writer can leave `extra=''` in a node row. `get_all_files` ran `json_extract(extra, '$.virtual')` over every non-File row, so SQLite raised `malformed JSON` and crashed **every** caller — build, status, visualize, watch.

Two changes in the WHERE clause:

1. **Malformed rows survive**: `extra IS NULL OR json_valid(extra) = 0` keeps a row with empty/truncated `extra` in the file inventory instead of aborting the query.
2. **Truthiness, not key presence**: `COALESCE(json_extract(extra, '$.virtual'), 0) != 1`. Previously the presence of the key alone hid a row, so `extra='{"virtual": false}'` was invisible to reconciliation and immortal; now only `virtual: true` (resolver-managed, e.g. Spring Event markers) stays out of the inventory.

## Testing

- New regression `test_get_all_files_survives_malformed_extra_and_false_virtual`: inserts rows with `extra=''`, `{"virtual": false}`, and `{"virtual": true}`, then asserts the split — malformed and false-virtual paths are returned, true-virtual stays excluded. **Fails against the previous query** (verified by checkout mutation: `OperationalError: malformed JSON`), passes with the fix.
- `pytest tests/test_graph.py tests/test_incremental.py` — **176 passed** (2 symlink-privilege tests deselected, a pre-existing Windows-environment limitation, same as #860/#861).
- `ruff check code_review_graph/graph.py tests/test_incremental.py` — all checks passed.

Fixes #864

### DIFF
--- a/code_review_graph/graph.py
+++ b/code_review_graph/graph.py
@@ -1190,9 +1190,16 @@ class GraphStore:
         # every represented path so those orphans can be purged, while explicit
         # virtual nodes (such as Spring Event markers) remain outside the file
         # inventory and are managed by their owning resolver.
+        # A killed writer can leave extra='' (or truncated text), which makes
+        # json_extract raise SQLite 'malformed JSON' and crash every caller of
+        # get_all_files. json_valid gates the extraction, and the truthiness
+        # comparison keeps extra='{"virtual": false}' visible instead of
+        # making the node immortal to reconciliation (#864).
         rows = self._conn.execute(
             "SELECT file_path FROM nodes "
-            "WHERE json_extract(extra, '$.virtual') IS NULL "
+            "WHERE extra IS NULL "
+            "OR json_valid(extra) = 0 "
+            "OR COALESCE(json_extract(extra, '$.virtual'), 0) != 1 "
             "UNION "
             "SELECT file_path FROM edges "
             "ORDER BY file_path"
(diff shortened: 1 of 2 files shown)


tirth8205:
Integrated on `integration/token-efficiency-hardening` as `a1a8daa`. File reconciliation remains available when node metadata contains malformed JSON and keeps `virtual=false` rows purgeable, while retaining resolver-owned `virtual=true` nodes. The targeted graph and incremental regressions passed.

tirth8205:
Merged into `staging` through the integration PR #988 (merge commit 3030ca8). Your change is carried by commit a1a8daa, which keeps you as the author (yzxcj797). It will be promoted staging → testing → main and released from there. Thank you for the contribution! Closing this PR since its content is now on staging.

PR #871, opened 2026-08-19
Apply lifecycle test-gap exemptions to review guidance

### LINKED ISSUE #865: Review guidance paths do not apply the lifecycle test-gap exemption from #854
Follow-up to #854, which exempted setUp/tearDown/__construct and friends from analyze_changes test gaps.

get_review_context_tool and _generate_review_guidance (code_review_graph/tools/review.py) compute their own untested lists without consulting _TEST_GAP_EXEMPT_NAMES, so review guidance still names lifecycle methods as lacking coverage while detect_changes no longer counts them. The two surfaces should agree.

Also from the same review: exempt nodes still receive the 0.30 untested penalty in risk scoring, so a diff touching only lifecycle methods reports 0 test gaps next to a nonzero risk score partly driven by untested. Worth aligning while in there.

yzxcj797:
## Summary

`detect_changes` exempts lifecycle and construction methods from test-gap reporting, but the review surfaces still computed their own untested lists. This could report `setUp`, `tearDown`, `__construct`, and similar methods as coverage gaps even when change analysis correctly reported zero gaps.

This change centralizes the lifecycle predicate and applies it consistently to:

- minimal review-context test-gap counts;
- generated review guidance;
- risk scoring, so exempt lifecycle methods no longer retain the 0.30 untested penalty.

Fixes #865.

## Testing

- `pytest tests/test_changes.py tests/test_pr854_edges.py tests/test_pr865_review_guidance.py -q` — 47 passed
- Added regressions for:
  - review guidance omitting exempt lifecycle methods while retaining real gaps;
  - lifecycle risk having no untested penalty.
- `ruff check` on all changed files — passes
- `git diff --check` — passes

A full Windows test run was also executed: 2,625 tests passed and 75 failed. The failures are in pre-existing Windows/path/encoding-sensitive areas unrelated to these files; the targeted change-related suites pass.

### DIFF
--- a/code_review_graph/changes.py
+++ b/code_review_graph/changes.py
@@ -31,6 +31,11 @@ _TEST_GAP_EXEMPT_NAMES = frozenset({
     "__construct", "__init__", "__destruct",
 })
 
+
+def is_test_gap_exempt(node: GraphNode) -> bool:
+    """Return whether a lifecycle node is exempt from test-gap reporting."""
+    return node.kind == "Function" and node.name in _TEST_GAP_EXEMPT_NAMES
+
 _GIT_TIMEOUT = int(os.environ.get("CRG_GIT_TIMEOUT", "30"))  # seconds, configurable
 
 _SAFE_GIT_REF = re.compile(r"^[A-Za-z0-9_.~^/@{}\-]+$")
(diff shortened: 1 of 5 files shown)


tirth8205:
Changes required: a name-only lifecycle exemption suppresses coverage risk for production code. Run `python -m pytest tests/test_changes.py -k lifecycle_method_has_no_untested_penalty`; the new test explicitly creates an untested `setUp` function in `a.py` and requires `compute_risk_score` to return zero, and the same policy also exempts production constructors. Restrict noise suppression to verified test lifecycle nodes and retain coverage risk for substantive production logic.

tirth8205:

- A test for get_review_context(..., detail_level="minimal") asserting result["test_gaps"] excludes lifecycle methods (both mixed exempt+real -> 1 and all-exempt -> 0), using a seeded .code-review-graph/graph.db like tests/test_tools.py::_seed_repo_relative_graph.
- A test for _generate_review_guidance when all changed functions are exempt: no 'lack test coverage' line is emitted (falls through to 'Changes appear well-contained').
- An end-to-end get_review_context (standard detail_level) test asserting context["review_guidance"] omits lifecycle names and still names real gaps, with real GraphNode objects rather than SimpleNamespace.
- (Optional) A test pinning that a covered lifecycle method now scores 0.0 for the coverage term (previously 0.05), so the semantic change is deliberate and documented.

tirth8205:

- `code_review_graph/changes.py`
- `code_review_graph/tools/review.py`

```bash
git fetch origin
git merge origin/staging
# resolve, then
uv run pytest tests/ -q
git push
```

PR #872, opened 2026-08-19
Handle comments in Go receiver types

### LINKED ISSUE #863: Go: receiver type unwrap can land on a comment node and drop method ownership
Follow-up noted while merging #833. In _get_go_receiver_type (code_review_graph/parser.py around line 14710), the unwrap loop takes next(iter(receiver_type.named_children), None). Tree-sitter comments are named extras, so for

    func (b * /*c*/ T) F() {}

the unwrap lands on the comment node, fails the type_identifier check, and F loses its parent where it previously attached to T.

Fix shape: skip comment nodes when picking the child to unwrap. While in there, parenthesized receivers (func (b (T)) and func (b (Box[T]))) are legal Go and still return None; adding parenthesized_type to the unwrap set would cover them. Rare in gofmt-formatted code, so low priority.

yzxcj797:
## Summary

Go receiver type unwrapping previously took the first named child of `generic_type` and `pointer_type`. Tree-sitter comments are named extras, so a receiver such as:

```go
func (b * /*receiver comment*/ Box[T]) Commented() {}
```

landed on the comment node and lost method ownership. Parenthesized receiver types were also outside the unwrap set.

The unwrap loop now skips comment nodes and includes `parenthesized_type`.

Fixes #863.

## Testing

- Added a regression with a comment inside a generic pointer receiver and a parenthesized generic receiver.
- The regression fails on clean `main` with both methods losing their `Box` parent and passes with the fix.
- `pytest tests/test_multilang.py::TestGoParsing -q` — 28 passed
- `ruff check code_review_graph/parser.py tests/test_multilang.py` — passes
- `git diff --check` — passes

### DIFF
--- a/code_review_graph/parser.py
+++ b/code_review_graph/parser.py
@@ -14796,9 +14796,15 @@ class CodeParser:
         )
         receiver_type = parameter.child_by_field_name("type") if parameter else None
         while receiver_type is not None and receiver_type.type in {
-            "generic_type", "pointer_type",
+            "generic_type", "pointer_type", "parenthesized_type",
         }:
-            receiver_type = next(iter(receiver_type.named_children), None)
+            receiver_type = next(
+                (
+                    child for child in receiver_type.named_children
+                    if child.type != "comment"
+                ),
+                None,
+            )
         if receiver_type is None or receiver_type.type != "type_identifier":
             return None
         return receiver_type.text.decode("utf-8", errors="replace")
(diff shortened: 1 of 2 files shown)


tirth8205:
Integrated on `integration/token-efficiency-hardening` as `f1a76f7`. Go receiver extraction now preserves method ownership when receiver syntax contains comments or parentheses. The new receiver regression fails on main, passes with the change, and the targeted Go tests passed.

tirth8205:
Merged into `staging` through the integration PR #988 (merge commit 3030ca8). Your change is carried by commit f1a76f7, which keeps you as the author (yzxcj797). It will be promoted staging → testing → main and released from there. Thank you for the contribution! Closing this PR since its content is now on staging.

PR #873, opened 2026-08-19
Read Kotlin import names from AST children

### LINKED ISSUE #862: Kotlin: _collect_import_names still folds trailing comments into per-name import attribution
Follow-up to #824, which fixed _extract_import to read the import_header's identifier child instead of raw node text.

The same folding remains in _collect_import_names (code_review_graph/parser.py, kotlin branch around lines 13406-13416), which still parses raw node text. For:

    import javax.inject.Inject

    /** doc */
    class D

the import_map entry becomes {'Inject\n\n/** doc */': 'javax.inject.Inject\n\n/** doc */'}, so per-name call attribution for the last import with a trailing comment is broken.

Fix shape: same as #824, read the identifier child of import_header instead of node text, plus a regression that puts a KDoc block after the final import and asserts the import_map keys are clean.

yzxcj797:
## Summary

`_extract_import` was fixed to read a Kotlin `import_header` identifier child instead of folded node text, but `_collect_import_names` still parsed raw text. When tree-sitter folded a trailing KDoc into the final import, the per-name import map stored a key containing the comment, so calls to that imported function remained unresolved.

The Kotlin branch now builds import bindings from AST children:

- identifier → module;
- `import_alias` → local alias;
- wildcard imports remain excluded from per-name bindings.

Java import handling remains unchanged.

Fixes #862.

## Testing

- Added an end-to-end regression with `import dep.helper`, a trailing KDoc, and a `helper()` call.
- On clean `main`, the call target stayed unresolved as `helper`; with the fix it resolves to `dep/Helper.kt::helper`.
- `pytest tests/test_pr862_edges.py tests/test_kotlin_imports.py tests/test_pr824_edges.py -q` — 23 passed
- `git diff --check` — passes

### DIFF
--- a/code_review_graph/parser.py
+++ b/code_review_graph/parser.py
@@ -13488,7 +13488,29 @@ class CodeParser:
                         if module_name and real_name and alias:
                             import_map[alias] = f"{module_name}.{real_name}"
 
-        elif language in ("java", "kotlin"):
+        elif language == "kotlin":
+            module = None
+            alias = None
+            wildcard = False
+            for child in node.children:
+                if child.type == "identifier":
+                    module = child.text.decode("utf-8", errors="replace")
+                elif child.type == "wildcard_import":
+                    wildcard = True
+                elif child.type == "import_alias":
+                    alias = next(
+                        (
+                            part.text.decode("utf-8", errors="replace")
+                            for part in child.children
+                            if part.type == "type_identifier"
+                        ),
+                        None,
+                    )
+            if module and not wildcard:
+                local_name = alias or module.rsplit(".", 1)[-1]
+                import_map[local_name] = module
+
+        elif language == "java":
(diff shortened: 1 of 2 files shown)


tirth8205:
Integrated on `integration/token-efficiency-hardening` as `84a5501`. Kotlin imports with trailing KDoc are read from syntax children so their calls retain the correct import attribution. The fixture was corrected to use the supported portable `helper.kt` naming; the focused import and Kotlin tests passed without expanding symbol-to-file resolution.

tirth8205:
Merged into `staging` through the integration PR #988 (merge commit 3030ca8). Your change is carried by commit 84a5501, which keeps you as the author (yzxcj797). It will be promoted staging → testing → main and released from there. Thank you for the contribution! Closing this PR since its content is now on staging.

PR #874, opened 2026-08-19
Address August sweep follow-ups

### LINKED ISSUE #866: Small follow-ups from the August PR sweep
Collecting the small non-blocking items noted while reviewing and merging the August batch, so they are not lost:

- tools/review.py: the 'No changed files detected.' early return in get_affected_flows omits the truncated key added by #853, so clients indexing result['truncated'] hit KeyError on that branch (noted merging #853).
- parser.py ~13591: the NodeNext/ESM suffix-substitution fallback only tries .ts and .tsx, so a ./foo.js specifier no longer resolves to an on-disk foo.jsx (allowJs/Babel layouts), and .mts/.cts sources never resolve (noted merging #845).
- cli.py ~1638: dead-code and forget still resolve the DB path in create mode, so a missing graph leaves an empty .code-review-graph/ dir behind before exiting 1 (noted merging #809; same class as #803 gap 1).
- daemon_cli.py: the UTF-8 stdio reconfigure from #826 covers only the code-review-graph entry point; crg-daemon would need the same call if it ever prints non-ASCII.
- daemon_cli.py: the forced-stop escalation catches only ProcessLookupError; on Windows a vanished PID raises OSError WinError 87, which would surface a raw traceback in a narrow race (noted merging #844).

yzxcj797:
## Summary

Addresses the small follow-ups collected in #866:

- `get_affected_flows` now includes `truncated: false` on its no-changed-files early return.
- JS/TS module suffix substitution now supports:
  - `./foo.js` → `foo.jsx`;
  - `./foo.mjs` → `foo.mts`;
  - `./foo.cjs` → `foo.cts`;
  - direct `.mts` / `.cts` source resolution.
- `dead-code` and `forget` resolve their graph path in read-only mode so a missing graph no longer creates an empty `.code-review-graph/` directory.
- `crg-daemon` configures UTF-8 stdio like the main CLI.
- Daemon forced-stop handles the Windows race where a vanished PID raises `OSError`/WinError 87.

Fixes #866.

## Testing

- Added focused regressions for each code-level follow-up.
- Extended the missing-graph CLI test to assert that no data directory is materialized.
- `pytest tests/test_pr866_followups.py tests/test_cli_reconciliation.py::test_dead_code_missing_graph_exits_nonzero -q` — 7 passed
- `ruff check` on all changed files — passes
- `git diff --check` — passes

### DIFF
--- a/code_review_graph/cli.py
+++ b/code_review_graph/cli.py
@@ -1658,7 +1658,10 @@ def main() -> None:
             # materialize graph state when neither database exists.
             db_path = get_db_path(repo_root)
     else:
-        db_path = get_db_path(repo_root)
+        db_path = get_db_path(
+            repo_root,
+            read_only=args.command in ("dead-code", "forget"),
+        )
     if (
         args.command in ("dead-code", "forget", *_read_only_db_cmds)
         and not db_path.exists()
--- a/code_review_graph/daemon_cli.py
+++ b/code_review_graph/daemon_cli.py
@@ -23,6 +23,21 @@ import time
 logger = logging.getLogger(__name__)
 
 
+def _configure_utf8_stdio() -> None:
+    """Allow Unicode daemon output on streams using a legacy encoding."""
+    for stream in (sys.stdout, sys.stderr):
+        encoding = getattr(stream, "encoding", None)
+        if not encoding or encoding.lower().replace("-", "") == "utf8":
+            continue
+        reconfigure = getattr(stream, "reconfigure", None)
+        if reconfigure is None:
+            continue
+        try:
+            reconfigure(encoding="utf-8")
+        except (OSError, ValueError):
+            pass
+
+
 # ---------------------------------------------------------------------------
 # Subcommand handlers
 # ---------------------------------------------------------------------------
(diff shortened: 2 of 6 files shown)


tirth8205:
Integrated on `integration/token-efficiency-hardening` as `928fb81`, `4f3c1ea`, `24b5a18`. The retained changes repair read-only CLI behaviour on a missing graph, JSX/MTS/CTS source resolution and the empty-flow `truncated` field. A later correction maps `.mts` and `.cts` to the existing TypeScript grammar so full builds retain their definitions and caller, callee and review context; regressions preserve native-extension priority and the existing `.ts` fallbacks. No daemon changes were retained: treating a generic `OSError` from signalling as process disappearance can clear the PID record while the daemon remains alive.

tirth8205:
Merged into `staging` through the integration PR #988 (merge commit 3030ca8). Your change is carried by commit 928fb81, which keeps you as the author (yzxcj797). It will be promoted staging → testing → main and released from there. Thank you for the contribution! Closing this PR since its content is now on staging.

PR #875, opened 2026-08-19
Import PHP include and require targets

### LINKED ISSUE #819: PHP: include/require produce no IMPORTS_FROM edges — procedural (non-Composer) PHP repos get an empty cross-file graph
### Summary

For PHP, `_IMPORT_NODE_TYPES` maps only `namespace_use_declaration`:

https://github.com/tirth8205/code-review-graph/blob/main/code_review_graph/parser.py#L1087

```python
"php": ["namespace_use_declaration"],
```

`include` / `include_once` / `require` / `require_once` are never mapped, so a PHP codebase that wires files together with `require_once` instead of Composer/PSR-4 autoloading produces **zero** `IMPORTS_FROM` edges. Impact radius, `importers_of` and `detect_changes` then have no cross-file structure to walk on those repos.

This is different from the (closed) #574, which was about `use` imports being stored as raw statement text — namespaced code. This one is about non-namespaced, procedural PHP, which is still very common in legacy apps and in a lot of self-hosted PHP software.

### Reproduction

Real project, 332 files, plain procedural PHP (PDO, no Composer autoload), crg 2.3.7 from PyPI, Python 3.12:

```bash
grep -c "require_once\|include_once\|require \|include " *.php | awk -F: '{s+=$2} END{print s}'
# 622

code-review-graph build --repo <project>
# Full build: 332 files, 1314 nodes, 25835 edges
```

Querying the resulting graph:

```
edges by kind: CALLS 24227, CONTAINS 982, IMPORTS_FROM 67, INHERITS 1
```

Those 67 `IMPORTS_FROM` edges do **not** come from PHP — they come from the `.sql` files in the repo, and they are junk:

```
('.../01_base_schema.sql', 'clientes')
('.../02_auditoria.sql',  'IF')
('.../03_planos_templates.sql', 'IF')
```

(`IF` as an import target looks like a separate SQL parsing bug — happy to file that separately if useful.)

So: 622 include/require statements in the source, 0 edges from PHP.

Because there is no import structure, `detect-changes` on a 20-file diff reported `0 affected flow(s)`, and the `CALLS` edges are dominated by PHP builtins rather than project structure:

```
execute 1799, prepare 1786, json_encode 1097, empty 1054,
htmlspecialchars 985, trim 740, isset 567
```

### Expected

`require_once 'includes/conexao.php'` should create an `IMPORTS_FROM` edge from the including file to `includes/conexao.php`, so impact radius reaches dependents.

### Suggested fix

The four node types exist in the upstream grammar ([`tree-sitter-php/common/define-grammar.js#L1619-L1623`](https://github.com/tree-sitter/tree-sitter-php/blob/master/common/define-grammar.js#L1619-L1623)):

```js
include_expression:      $ => seq(keyword('include'),      $.expression),
include_once_expression: $ => seq(keyword('include_once'), $.expression),
require_expression:      $ => seq(keyword('require'),      $.expression),
require_once_expression: $ => seq(keyword('require_once'), $.expression),
```

so the mapping side is a one-liner:

```python
"php": [
    "namespace_use_declaration",
    "include_expression",
    "include_once_expression",
    "require_expression",
    "require_once_expression",
],
```

The target extraction needs a bit more than that, since the argument is an expression rather than a plain string. Two cases worth covering:

1. **Literal path** — `require_once 'includes/conexao.php';` → resolve relative to the including file's directory, bounded to the repository (same containment rule the Composer PSR-4 resolution already uses).
2. **Concatenated path** — `require_once __DIR__ . '/includes/conexao.php';` is extremely common. Resolving `__DIR__` to the including file's directory would cover most real-world usage. Fully dynamic paths (`require $var;`) are not statically resolvable and should just be skipped, or emitted at a lower confidence tier.

Given the existing three-tier confidence model, literal and `__DIR__`-prefixed paths that resolve to a file inside the repo seem like a natural `EXTRACTED`; anything partially dynamic could be `AMBIGUOUS` or dropped.

### Notes

Worth flagging that this is invisible in the benchmark suite — none of the 6 evaluation repos is a procedural PHP codebase, so a repo in this shape reports a healthy-looking node/edge count while the cross-file graph is effectively empty. A build-time warning when a language contributes many files but ~0 import edges might be a cheap way to surface this class of problem generally.

Also, `.code-review-graph/languages.toml` can't be used as a workaround here, since built-in languages can't be overridden — which is the right call, just noting there's no user-side escape hatch.

### Environment

- code-review-graph 2.3.7 (PyPI)
- Python 3.12.3, Linux x86_64
- No embeddings/igraph extras installed

yzxcj797:
## Summary

PHP `use` declarations produced import edges, but procedural PHP files wired together with `include`, `include_once`, `require`, and `require_once` had no `IMPORTS_FROM` edges. Impact radius and importer queries therefore saw no cross-file structure for non-Composer codebases.

The parser now treats all four include/require expression forms as import nodes, extracts their string argument, and resolves relative paths ending in `.php` without appending a second extension.

Fixes #819.

## Testing

- Added end-to-end regressions for:
  - `require_once 'includes/conexao.php'`;
  - `include_once 'helper.php'`.
- Both tests fail on clean `main` with zero import edges and pass with the fix.
- `pytest tests/test_pr819_php_includes.py -q` — 2 passed
- `pytest tests/test_php_scoped_calls.py tests/test_php_laravel.py tests/test_pr819_php_includes.py -q` — 56 passed, 2 skipped, plus one pre-existing Windows path-separator assertion unchanged by this patch
- `git diff --check` — passes

### DIFF
--- a/code_review_graph/parser.py
+++ b/code_review_graph/parser.py
@@ -1086,7 +1086,13 @@ _IMPORT_TYPES: dict[str, list[str]] = {
     "perl": ["use_statement", "require_expression"],
     "kotlin": ["import_header"],
     "swift": ["import_declaration"],
-    "php": ["namespace_use_declaration"],
+    "php": [
+        "namespace_use_declaration",
+        "include_expression",
+        "include_once_expression",
+        "require_expression",
+        "require_once_expression",
+    ],
     "scala": ["import_declaration"],
     "solidity": ["import_directive"],
     # Dart: import_or_export wraps library_import > import_specification > configurable_uri
@@ -13788,7 +13794,12 @@ class CodeParser:
             # ``.../App/Foo``) resolve; vendor/global classes (``\Exception``)
             # and ``use function`` / ``use const`` targets with no matching
             # file stay unresolved and keep the bare FQN, like JDK imports.
-            rel_path = module.replace("\\", "/").lstrip("/") + ".php"
+            normalized_module = module.replace("\\", "/").lstrip("/")
+            rel_path = (
+                normalized_module
+                if normalized_module.endswith(".php")
+                else normalized_module + ".php"
+            )
             try:
                 boundary = self._php_repository_boundary(caller_dir)
                 current = caller_dir.resolve()
@@ -15708,6 +15719,22 @@ class CodeParser:
                     if txt and txt != "extends":
                         imports.append(txt)
         elif language == "php":
+            include_types = {
+                "include_expression",
+                "include_once_expression",
+                "require_expression",
+                "require_once_expression",
+            }
+            if node.type in include_types:
+                for child in node.children:
+                    if child.type == "string":
+                        value = child.text.decode(
+                            "utf-8", errors="replace",
+                        ).strip("'\"")
+                        if value:
+                            imports.append(value)
+                return imports
+
             # ``namespace_use_declaration`` covers several shapes:
             #   use A\B\C;            use A\B\C as D;
             #   use function A\b;     use const A\B;
--- a/tests/test_pr819_php_includes.py
+++ b/tests/test_pr819_php_includes.py
@@ -1,0 +1,39 @@
+"""PHP include/require import edge regressions (#819)."""
+
+from pathlib import Path
+
+from code_review_graph.parser import CodeParser
+
+
+def test_require_once_resolves_relative_php_file(tmp_path: Path) -> None:
+    includes = tmp_path / "includes"
+    includes.mkdir()
+    dependency = includes / "conexao.php"
+    dependency.write_text("<?php\nfunction connect() {}\n", encoding="utf-8")
+    source = tmp_path / "index.php"
+    source.write_text(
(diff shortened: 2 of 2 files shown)


tirth8205:
Integrated on `integration/token-efficiency-hardening` as `97bc2fc`, `f08ebee`. Direct plain single- and double-quoted PHP include/require paths now supply importer context, with corrections preserving absolute paths and repository containment. Dynamic or compound expressions, interpolated strings and double-quoted escapes remain unsupported; the empty-query caveat now describes dynamic include/require gaps. Literal, interpolation, escape and absolute-path regressions passed.

tirth8205:
Merged into `staging` through the integration PR #988 (merge commit 3030ca8). Your change is carried by commit 97bc2fc, which keeps you as the author (yzxcj797). It will be promoted staging → testing → main and released from there. Thank you for the contribution! Closing this PR since its content is now on staging.

PR #876, opened 2026-08-19
Prefilter plain C headers before C++ parsing

### LINKED ISSUE #802: C/C++ parse throughput regression from header promotion and Qt masking (follow-up to #775)
PR #775 (merged) fixed #463 but measurably slows C/C++ parsing: the Qt macro masking pass is a per-byte Python loop over every C++ file (~49ms per 290KB measured), and pure C headers pay a speculative tree-sitter-cpp probe parse (measured 69ms to 189ms on a 175KB header, about 2.7x; an independent verification measured about 3x on a synthetic 152KB header). Both costs apply at build/update time only.

Proposed mitigations, from the review that measured this:
1. Skip the masking pass entirely when the source contains no Q_ or QT_ token (cheap substring prefilter).
2. Skip or cheapen the speculative C++ probe for .h files with no C++-evidence tokens (class, namespace, template, ::) in a fast pre-scan.

Both preserve the merged behavior on Qt/C++ headers while restoring near-original throughput for plain C repositories.

yzxcj797:
## Summary

Header promotion introduced two throughput costs for plain C repositories:

- every `.h` file paid a speculative tree-sitter-cpp parse;
- every C++ file paid the per-byte Qt macro masking walk even when no Qt macros existed.

This change adds cheap source-level prefilters:

- `.h` files are only speculatively parsed as C++ when they contain textual evidence such as `class`, `namespace`, `template`, `::`, `constexpr`, `noexcept`, or a trailing return arrow;
- Qt macro masking returns immediately when the source contains neither `Q_` nor `QT_`.

Qt/C++ behavior is unchanged when those signals are present.

Fixes #802.

## Testing

- Added a plain C header regression that fails if the C++ parser is probed.
- Added a C++ header regression confirming speculative promotion still happens.
- Added a direct Qt-mask fast-path regression.
- Existing Qt header behavior remains covered.
- `pytest tests/test_pr802_cpp_prefilter.py tests/test_cpp_qt_headers.py -q` — 17 passed
- `git diff --check` — passes

### DIFF
--- a/code_review_graph/parser.py
+++ b/code_review_graph/parser.py
@@ -2614,15 +2614,16 @@ class CodeParser:
         tree = None
         parse_source = source
         if language == "c" and path.suffix.lower() == ".h":
-            cpp_parser = self._get_parser("cpp")
-            if cpp_parser is not None:
-                cpp_source = self._mask_cpp_qt_macros(source)
-                cpp_tree = cpp_parser.parse(cpp_source)
-                if self._has_cpp_header_evidence(cpp_tree.root_node):
-                    language = "cpp"
-                    parser = cpp_parser
-                    tree = cpp_tree
-                    parse_source = cpp_source
+            if self._has_cpp_header_text_evidence(source):
+                cpp_parser = self._get_parser("cpp")
+                if cpp_parser is not None:
+                    cpp_source = self._mask_cpp_qt_macros(source)
+                    cpp_tree = cpp_parser.parse(cpp_source)
+                    if self._has_cpp_header_evidence(cpp_tree.root_node):
+                        language = "cpp"
+                        parser = cpp_parser
+                        tree = cpp_tree
+                        parse_source = cpp_source
         elif language == "cpp":
             parse_source = self._mask_cpp_qt_macros(source)
 
@@ -2824,8 +2825,21 @@ class CodeParser:
         return False
 
     @staticmethod
+    def _has_cpp_header_text_evidence(source: bytes) -> bool:
+        """Cheaply decide whether a C header merits a speculative C++ parse."""
+        return any(
+            token in source
+            for token in (
+                b"class", b"namespace", b"template", b"::",
+                b"constexpr", b"noexcept", b"->",
+            )
+        )
+
+    @staticmethod
     def _mask_cpp_qt_macros(source: bytes) -> bytes:
         """Shield structural Qt macros without changing byte or line offsets."""
+        if b"Q_" not in source and b"QT_" not in source:
+            return source
         masked = bytearray(source)
         length = len(source)
         index = 0
--- a/tests/test_pr802_cpp_prefilter.py
+++ b/tests/test_pr802_cpp_prefilter.py
@@ -1,0 +1,36 @@
+"""C/C++ parse-throughput prefilters (#802)."""
+
+from pathlib import Path
+
+from code_review_graph.parser import CodeParser
+
+
+def test_plain_c_header_skips_speculative_cpp_parser(monkeypatch):
+    source = b"struct point { int x; int y; };\nint distance(struct point p);\n"
+
+    original_parser = CodeParser._get_parser
+
+    def guarded_parser(parser_self, language):
+        if language == "cpp":
+            raise AssertionError("plain C header must not probe the C++ parser")
+        return original_parser(parser_self, language)
+
+    monkeypatch.setattr(CodeParser, "_get_parser", guarded_parser)
+
+    nodes, _ = CodeParser().parse_bytes(Path("point.h"), source)
+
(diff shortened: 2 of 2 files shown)


tirth8205:
Changes required: the C++ header prefilter drops structural context that main already records. Save `struct Parent {}; struct Child : Parent { int run() { return 1; } };` in `derived.h` and call `CodeParser(repo_root).parse_file(path)`; this branch selects C parsing and loses the inheritance edge and `Child` ownership of `run`. The fast path must conservatively cover every existing C++ promotion signal before it can be retained.

tirth8205:
Holding this one: the problem is the change, not the tests.

All three gates pass and the merge is clean, but the change is wrong, not merely under-tested. The new _has_cpp_header_text_evidence prefilter gates the existing AST-based _has_cpp_header_evidence check, yet its token list (class, namespace, template, ::, constexpr, noexcept, ->) is not a superset of the evidence the AST check recognises (_CPP_HEADER_EVIDENCE_TYPES: access_specifier, alias_declaration, base_class_clause, class_specifier, concept_definition, lambda_expression, namespace_definition, noexcept, template_declaration, trailing_return_type, using_declaration; enum class/struct; consteval/constinit). I ran the same 11 headers through a second worktree at plain origin/staging and the merged worktree: 7 shapes flip from cpp to c after the PR (struct with public:, using X = int;, struct Derived : Base {};, enum struct, consteval, constinit, inline lambda).

What would change our mind:
- Regression tests in tests/test_pr802_cpp_prefilter.py (or test_cpp_qt_headers.py) proving a .h is still promoted to cpp when its ONLY evidence is: access_specifier, alias_declaration, base_class_clause, enum struct, consteval, constinit, lambda_expression.
- Fix to _has_cpp_header_text_evidence so the token list is a superset of what _has_cpp_header_evidence can match: add at least b'using', b'public', b'private', b'protected', b'consteval', b'constinit', plus a cheap signal for base_class_clause / enum struct / lambdas (e.g.
- A test asserting every key of _CPP_QT_STRUCTURAL_MACRO_REPLACEMENTS starts with Q_ or QT_, so the new fast path in _mask_cpp_qt_macros stays exact if the table grows.
- Optional: an end-to-end regression for empty .h input and for false-positive tokens (-> in a C macro, 'subclass' identifier) confirming they still index as C.

Checks pass, so this only showed up when we probed the behaviour directly. Push back here if you think that is wrong.

tirth8205:
The changes from the last review were never pushed. The only content commit on this branch is 1c801981 from 19 August, and 3fbd23c5 is a staging merge on top. The token list is unchanged, `tests/test_pr802_cpp_prefilter.py` still has its original three tests, and the Qt table assertion is missing.

I built a real C++ tree on both branches to price the downgrade. Corpus is bulletphysics/bullet3, `src` plus `examples`, 2603 files, 1280 `.h` headers, same corpus both sides, paths normalised.

```
                staging   this branch   delta
nodes             52476         52407     -69
edges            224880        224816     -64
  Class            8924          8922      -2
  Function        40851         40784     -67
  CONTAINS        55280         55214     -66
  INHERITS          943           935      -8
```

All 8 INHERITS deltas are losses, none are gains. `btDbvtBroadphase` from `btBroadphaseInterface`, `SimpleCamera` from `CommonCameraInterface`, `btContactSolverInfo` from `btContactSolverInfoData` and five more stop existing.

The `Class -2` is silent corruption rather than missing data. `nodes.qualified_name` is `TEXT NOT NULL UNIQUE` at graph.py:200. When a C++ class is parsed as C, its destructor `~Foo()` is read as a function named `Foo`, which claims the same key and overwrites the class. Three real bullet3 classes are now stored as Function nodes under their own names: `b3BroadphaseAabbCallback`, `b3BroadphaseRayCallback`, `btContactSolverInfoData`. Queries for those classes return the wrong kind of node.

Two more identities lose their signature. `b3GrahamScanConvexHull2D(b3AlignedObjectArray<b3GrahamVector3>&,b3AlignedObjectArray<b3GrahamVector3>&,const b3Vector3&)` becomes plain `b3GrahamScanConvexHull2D`, so edges in an existing database that point at the old spelling no longer resolve. `setVelocityPrediction` and `SimpleCameraInternalData` lose their parent class and reparent to the file.

The speedup is real: 34% on libuv's 38 pure C headers, 31% on bullet3's 1411 headers, best of three. On fmtlib/fmt the graph diff is exactly zero, because every one of its headers contains `template`, `namespace` or `::`. That is why the three tests here pass.

To merge: widen the token list to a real superset of `_CPP_HEADER_EVIDENCE_TYPES` as listed in my earlier comment, add one regression test per uncovered evidence type, add the Qt key prefix assertion, and re-run a build diff on a C++ repo showing zero node and edge loss.

PR #877, opened 2026-08-19
Add Windows overrides for Codex hooks

### LINKED ISSUE #620: Bug: Codex installer generates Unix-only hooks on Windows

yzxcj797:
## Summary

The generated Codex hooks previously contained only Unix shell commands. On Windows, Codex attempted to run `cat`, `/dev/null`, and Unix shell chaining, causing PostToolUse hook failures.

Both generated hooks now include `commandWindows` overrides using PowerShell:

- drain stdin with `[Console]::In.ReadToEnd()`;
- check whether the current directory is a Git repository;
- run the graph update or status command;
- always exit successfully, matching the Unix `|| true` behavior;
- print `Not a git repo, skipping` for the startup hook outside Git repositories.

Fixes #620.

## Testing

- Added regressions asserting both Windows overrides drain stdin, perform the Git check, invoke the correct graph command, and end successfully.
- `pytest tests/test_skills.py::TestGenerateCodexHooksConfig tests/test_skills.py::TestInstallCodexHooks -q` — 12 passed
- `ruff check code_review_graph/skills.py tests/test_skills.py` — passes
- `git diff --check` — passes

### DIFF
--- a/code_review_graph/skills.py
+++ b/code_review_graph/skills.py
@@ -1094,6 +1094,12 @@ def generate_codex_hooks_config(repo_root: Path) -> dict[str, Any]:
                                 " && code-review-graph update --skip-flows"
                                 " || true"
                             ),
+                            "commandWindows": (
+                                'powershell.exe -NoProfile -NonInteractive -Command "'
+                                "[Console]::In.ReadToEnd() | Out-Null; "
+                                "if ((git rev-parse --git-dir) -eq $null) { exit 0 }; "
+                                'code-review-graph update --skip-flows; exit 0"'
+                            ),
                             "timeout": 30,
                             "statusMessage": "Updating code-review-graph",
                         },
@@ -1111,6 +1117,13 @@ def generate_codex_hooks_config(repo_root: Path) -> dict[str, Any]:
                                 "git rev-parse --git-dir >/dev/null 2>&1"
                                 " && code-review-graph status"
                                 " || echo 'Not a git repo, skipping'"
+                            ),
+                            "commandWindows": (
+                                'powershell.exe -NoProfile -NonInteractive -Command "'
+                                "[Console]::In.ReadToEnd() | Out-Null; "
+                                "if ((git rev-parse --git-dir) -eq $null) { "
+                                "'Not a git repo, skipping'; exit 0 }; "
+                                'code-review-graph status; exit 0"'
                             ),
                             "timeout": 10,
                             "statusMessage": "Checking code-review-graph status",
--- a/tests/test_skills.py
+++ b/tests/test_skills.py
@@ -584,6 +584,23 @@ class TestGenerateCodexHooksConfig:
         assert "code-review-graph update --skip-flows" in post_cmd
         assert "code-review-graph status" in session_cmd
 
+    def test_windows_commands_drain_stdin_and_exit_successfully(self, tmp_path):
+        config = generate_codex_hooks_config(tmp_path)
+        post = config["hooks"]["PostToolUse"][0]["hooks"][0]
+        session = config["hooks"]["SessionStart"][0]["hooks"][0]
+
+        for command in (post["commandWindows"], session["commandWindows"]):
+            assert command.startswith(
+                'powershell.exe -NoProfile -NonInteractive -Command "'
+            )
+            assert "[Console]::In.ReadToEnd() | Out-Null" in command
+            assert "git rev-parse --git-dir" in command
+            assert command.endswith('exit 0"')
+
+        assert "code-review-graph update --skip-flows" in post["commandWindows"]
+        assert "code-review-graph status" in session["commandWindows"]
+        assert "Not a git repo, skipping" in session["commandWindows"]
+
 
 class TestInstallCodexHooks:
     def test_creates_hooks_file(self, tmp_path, monkeypatch):


tirth8205:
Changes required: reinstall still fails to add Windows overrides to existing Codex hooks. Generate a config with `generate_codex_hooks_config`, remove its inner `commandWindows` fields, write it to the hooks.json location used by `install_codex_hooks`, and call that installer again; the fields remain absent because the existing Unix commands match. Consolidate the migration with the related hook PRs and verify actual execution in the released Windows client.

tirth8205:

- Upgrade path: install_codex_hooks must add (or refresh) commandWindows on existing entries whose `command` already matches, plus a test that writes an older Unix-only hooks.json, reinstalls, and asserts both hooks gain commandWindows and no duplicate entries are appended.
- Windows execution tests behind @pytest.mark.skipif(sys.platform != 'win32') that run each commandWindows string through os.environ['COMSPEC'] /C exactly as Codex does: (a) in a `git init` tmp repo with a multi-MB piped stdin , returncode 0, no hang/BrokenPipe; (b) outside a git repo , returncode 0 a...
- Add tests/test_skills.py (or the new Windows tests) to the `windows-native` job in .github/workflows/ci.yml so the Windows-only tests actually execute in CI.
- CHANGELOG.md [Unreleased] entry for the fix referencing #620 (repo convention; the PR touches no changelog).

tirth8205:
The upgrade path works because of `_merge_hook_entries`, which landed on staging in d39f9fa on 2026-09-16 from a different PR, not because of anything on this branch. The only content commit here is still c883e50 from 2026-08-19. Installing with staging and then reinstalling with this branch into the same HOME does add `commandWindows` to both existing Codex entries with no duplicate appended, but nothing pins that coupling. If `_is_generated_hook_command` or the merge rule changes, Windows users keep a Unix-only hooks.json forever and no test notices.

Four items from my 2026-09-15 comment are still open and neither review got a reply: the reinstall test over an older Unix-only hooks.json, the Windows-only execution tests behind `skipif(sys.platform != 'win32')`, the CI wiring, and the CHANGELOG entry for #620.

The new test never runs on a Windows runner. It sits in `TestGenerateCodexHooksConfig`, and the `windows-native` job enumerates only:

```
tests/test_skills.py::TestInstallCodexHooks
tests/test_skills.py::TestInstallCursorHooks
```

It runs only in the cross-platform `test` job, where it checks string shape and proves nothing about Windows.

Nothing here shows Codex honours a `commandWindows` key at all. The installer writes that key into every user's `~/.codex/hooks.json` on every OS. If Codex ignores unknown keys it is harmless. If it validates the hooks schema strictly, every Codex hook breaks for everyone.

The two forms also diverge on the failure branches, untested. The Unix form sends git's stderr to `/dev/null` and the PowerShell form does not, so outside a git repo it prints a red error record on every hook fire. If git is missing from PATH, PowerShell raises `CommandNotFoundException`, skips the `if` body, and runs code-review-graph anyway, where the Unix form takes the skip branch.

The claim does reproduce with the real CLI. Staging writes Codex hooks with no `commandWindows`, this branch writes both. The new test fails on staging with `KeyError: 'commandWindows'` and passes here, so it is a real guard. Gates are clean, and uninstall is unaffected because `uninstall._commands()` matches on `command` only.

To merge: the reinstall test, the Windows execution tests, `TestGenerateCodexHooksConfig` added to `windows-native`, the CHANGELOG entry, and evidence that a released Codex build actually runs `commandWindows`.

PR #878, opened 2026-08-19
Index Liquid templates

### LINKED ISSUE #348: not picking .liquid files
not picking .liquid files

yzxcj797:
## Summary

`.liquid` files were not recognized by `detect_language`, so full builds excluded them and file summaries returned zero nodes.

This adds a structural, regex-based Liquid parser that indexes:

- a `File` node for every template;
- `Variable` nodes for `assign` declarations, including whitespace `liquid` blocks;
- `Output` nodes for non-empty `{{ ... }}` expressions;
- `REFERENCES` edges for static `render` and `include` targets.

Liquid comments are masked while preserving byte offsets and line numbers, so directives inside comments are not indexed.

Fixes #348.

## Testing

- Added regressions for:
  - ordinary `assign`, output, `render`, and comment masking;
  - whitespace `{% liquid ... %}` blocks.
- `pytest tests/test_liquid.py -q` — 2 passed
- `ruff check code_review_graph/parser.py tests/test_liquid.py` — passes
- `git diff --check` — passes

### DIFF
--- a/code_review_graph/parser.py
+++ b/code_review_graph/parser.py
@@ -782,6 +782,7 @@ EXTENSION_TO_LANGUAGE: dict[str, str] = {
     ".properties": "properties",
     ".yml": "yaml",
     ".yaml": "yaml",
+    ".liquid": "liquid",
 }
 
 # ``.h`` is shared by C and C++. Keep C as the extension default, then promote
@@ -2410,6 +2411,21 @@ class CodeParser:
     _BLADE_DIRECTIVE_RE = re.compile(
         r"""(?<!@)@(extends|include|component|livewire)\s*\(\s*(['"])([^'"]+)\2\s*\)""",
     )
+    _LIQUID_COMMENT_RE = re.compile(
+        r"\{%-?\s*comment\s*-?%\}.*?\{%-?\s*endcomment\s*-?%\}",
+        re.DOTALL,
+    )
+    _LIQUID_TAG_RE = re.compile(
+        r"\{%-?\s*([A-Za-z_][A-Za-z0-9_-]*)(.*?)-?%\}",
+        re.DOTALL,
+    )
+    _LIQUID_OUTPUT_RE = re.compile(r"\{\{-?\s*(.*?)\s*-?\}\}", re.DOTALL)
+    _LIQUID_ASSIGN_RE = re.compile(
+        r"(?<![\w-])assign\s+([A-Za-z_][A-Za-z0-9_]*)\s*="
+    )
+    _LIQUID_RENDER_RE = re.compile(
+        r"(?<![\w-])(render|include)\s+(['\"])([^'\"]+)\2"
+    )
     _LARAVEL_ROUTE_FACADE = "Illuminate\\Support\\Facades\\Route"
     _LARAVEL_ELOQUENT_MODEL = "Illuminate\\Database\\Eloquent\\Model"
     _LARAVEL_ROUTE_VERBS = frozenset({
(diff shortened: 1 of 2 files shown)


tirth8205:
Outside this sweep's mission of demonstrated coding-session token savings: this adds a new Liquid parser. Bare template references, raw-block handling and a complete cross-file impact workflow need evidence before that language expansion is included.

tirth8205:

- Test that {% raw %}...{% endraw %} content is not indexed (currently produces false-positive Output/Variable/REFERENCES nodes; the fix would mask raw blocks the same way comments are masked).
- Test that '#' comment lines inside {% liquid %} blocks are not indexed (currently 'assign'/'render' on '#' lines produce false positives).
- Test for {% include 'x' %} and for double-quoted render/include targets (include is never exercised).
- Test for empty file and for empty {{ }} output being skipped.
- Test for trim markers ({%- -%}, {{- -}}) and multi-line {{ ... }} output line ranges.

tirth8205:
`{% raw %}` blocks are indexed as live code, and so are `#` comment lines inside a liquid tag. Both reproduce on the head commit, and both are among the five tests I asked for on 15 September. Those tests are not in the branch and there was no reply here. The only commit since is the staging merge on 18 September, and `tests/test_liquid.py` still has exactly two tests.

A file containing only a raw block:

```liquid
{% raw %}
{% assign should_not_exist = 1 %}
{{ should_not_be_output }}
{% render 'not-a-real-snippet' %}
{% endraw %}
```

yields a Variable `should_not_exist`, an Output `output_3:should_not_be_output`, and a REFERENCES edge to `not-a-real-snippet`. None of it should reach the graph. `_mask_liquid_comments` masks `{% comment %}` and nothing else.

Same for `#` lines inside a liquid tag:

```liquid
{% liquid
  # assign ghost_var = 1
  # render 'ghost-snippet'
  assign real_var = 2 %}
```

Variables come out as `['ghost_var', 'real_var']` with a phantom REFERENCES edge to `ghost-snippet`.

Render targets still do not resolve. I built Shopify Dawn on this branch. All 127 new liquid REFERENCES edges are stored with `target_resolution='unresolved'` against a bare string such as `card-product` or `cart-drawer`. `snippets/card-product.liquid` is a File node in the same graph with 0 inbound edges, even though 7 templates render it. "Who renders this snippet" returns nothing, so the cross-file impact this is for is not there yet.

Output names embed the line number, so qualified_name is `<file>::output_{line}:<expr>`. I inserted one `{% comment %}` line at the top of `sections/image-banner.liquid` and ran `code-review-graph update`. 30 of that file's 44 nodes were destroyed and recreated under new names. No other node kind behaves this way, and `memory.py`, graph_diff snapshots and the embeddings index all key on qualified_name. The same build emits 4046 liquid nodes over Dawn but stores only 3836. 210 collapse across 156 names through `UNIQUE(qualified_name)`, for example `card-product.liquid::ratio` six times, so line_start lands on an arbitrary occurrence.

Gates are genuinely clean: 4237 passed, 0 failed, ruff clean, mypy clean. The table diff over Dawn is purely additive. 0 node names lost, 0 edges lost, Class 51 to 51, Function 429 to 429, CALLS 2522 to 2522. Nothing existing is touched or respelled.

One cost to weigh while fixing the rest. 3445 of 4352 nodes are Output nodes whose name is raw expression text. FTS hits for `product` go from 67 to 1151, and 85% of those are Output nodes. That works against the token-efficiency goal on exactly the repos this targets.

To merge:

- Mask `{% raw %}` blocks and `#` lines the way comments are masked, each with a test.
- Resolve render and include targets to the `snippets/<name>.liquid` File node so inbound edges exist.
- Drop the line number from Output names and make the key unique without it.
- Add the remaining three tests from my earlier list, plus `tests/fixtures/sample.liquid`.

PR #879, opened 2026-08-19
fix(watch): exit when the observer emitter thread dies silently

### LINKED ISSUE #811: watch: observer thread silently crashes on Maven target/ dir churn (KeyError in watchdog inotify) — daemon looks alive but stops updating
## Summary

`code-review-graph watch` (and by extension `crg-daemon`) schedules a single recursive `watchdog.Observer` over the entire repo root:

```python
# incremental.py
observer.schedule(handler, str(repo_root), recursive=True)
```

`ignore_patterns` (from `.gitignore` + `DEFAULT_IGNORE_PATTERNS`) only filters which changed files get re-parsed *after* an inotify event arrives — it does not reduce what the OS-level inotify watch actually covers. So even though `target/` is correctly excluded via `.gitignore` and matches `DEFAULT_IGNORE_PATTERNS` (`/target/**`), inotify still watches every subdirectory under it.

Directories with very high create/delete churn — in our case a Java/Maven repo's `intranet-backend/target/surefire/*` (Surefire forks create and delete many short-lived temp dirs per test run) — trigger a `KeyError` inside `watchdog`'s inotify buffer thread:

```
Exception in thread Thread-3:
  File ".../watchdog/observers/inotify_buffer.py", line 82, in run
    inotify_events = self._inotify.read_events()
  File ".../watchdog/observers/inotify_c.py", line 363, in read_events
    if self._wd_for_path[path] == wd:
       ~~~~~~~~~~~~~~~~~^^^^^^
KeyError: b'/path/to/repo/intranet-backend/target/surefire'
```

Since this is a **thread** crash (not the child process itself), the daemon's own health check never notices:

```python
# daemon.py
alive = proc is not None and proc.poll() is None
```

`proc.poll()` still returns `None` — the Python process is alive, only its internal inotify observer thread died. `daemon status` reports the watcher as `alive`, but it silently stops receiving any filesystem events from that point on. No further file changes (including everything from subsequent `git pull`/`rebase`) get picked up until the process is manually restarted.

## Reproduction

1. `code-review-graph daemon start` (or plain `code-review-graph watch --repo <path>`) on a Java/Maven repo.
2. Run `./mvnw test` (Surefire forks generate/delete many temp dirs under `target/surefire/`).
3. Watcher log shows the `KeyError` traceback above and stops updating.
4. `daemon status` / `proc.poll()` still reports the watcher as alive.
5. Subsequent commits, `git pull`, edits are never reflected in the graph (`built_at_sha` stays frozen at the pre-crash commit; `head_matches_build` stays `false` indefinitely).

## Impact

- Affects any language/build tool whose output directory churns subdirectories fast under load (Maven Surefire forks, Gradle test workers, etc.) — not exclusive to Java.
- Failure is silent: `daemon status` shows green/alive, no error surfaced to the user; only a stale `built_at_sha` / diffing graph stats reveal it.

## Suggested fixes (any of)

1. Actually honor `ignore_patterns` (or at least `DEFAULT_IGNORE_PATTERNS`) at `Observer.schedule()` time — schedule per-subdirectory instead of one recursive watch on repo root, skipping ignored directories from being watched at all.
2. Wrap the watchdog observer thread's event loop so an unhandled exception restarts the observer instead of silently dying (or exits the child process so the daemon's `proc.poll()`-based health check can catch and restart it).
3. Expose a `--watch-ignore` / config option to explicitly exclude paths from the OS-level watch, separate from the parse-time `ignore_patterns`.

## Environment

- OS: WSL2 / Linux
- `watchdog` inotify backend
- Repo: multi-module Maven + Vue/TS monorepo, `code-review-graph daemon` (`crg-watch`) managing 6 worktrees of the same repo in parallel

Happy to provide full logs if useful.

yzxcj797:
## Summary

A watchdog **emitter reader thread** can crash under heavy create/delete churn — the inotify `_wd_for_path` `KeyError` when Maven Surefire forks churn `target/surefire/`, or the Windows `ReadDirectoryChangesW` variant (both documented in #811) — while the observer thread itself stays alive. The child process keeps passing the daemon's `proc.poll()`-based health check while delivering **no** filesystem events: `daemon status` stays green, `built_at_sha` freezes, and every subsequent edit/`git pull` is missed until a manual restart.

The watch loop now checks `_observer_threads_alive(observer)` — the observer thread **and every emitter thread** — once per second. On the first dead thread it logs `critical` and raises `SystemExit(1)`, exiting the child so the daemon's existing restart path recovers watch instead of silently serving a stale graph forever.

This deliberately implements the issue's fix option 2 (exit so poll-based health checks catch it) rather than option 1 (per-subdirectory scheduling): it covers both the Linux and Windows thread variants, and the reproducer requires churn that a Windows-CI test cannot drive. A follow-up can layer schedule-time ignore filtering on top without conflicting.

## Testing

- New `TestWatchObserverLiveness` (3 cases): alive observer + alive emitter → healthy; dead observer → not healthy; **dead emitter with alive observer → not healthy** (the exact silent-death shape from #811).
- `pytest tests/test_graph.py tests/test_incremental.py` — **178 passed** (2 symlink-privilege tests deselected, the same pre-existing Windows-environment limitation noted in #860/#861/#870).
- `ruff check code_review_graph/incremental.py tests/test_incremental.py` — all checks passed.

Verification boundary: the end-to-end kill (driving a real inotify race under Maven churn) requires the Linux/WSL environment from the report; the unit tests pin the liveness predicate and the exit decision it feeds.

Fixes #811

### DIFF
--- a/code_review_graph/incremental.py
+++ b/code_review_graph/incremental.py
@@ -17,7 +17,7 @@ import sys
 import threading
 import time
 from pathlib import Path, PurePosixPath
-from typing import Callable, Optional
+from typing import Any, Callable, Optional
 
 from .graph import GraphStore
 from .parser import CodeParser, normalize_file_path
@@ -1537,6 +1537,21 @@ def _create_watch_handler(
     return GraphUpdateHandler()
 
 
+def _observer_threads_alive(observer: "Any") -> bool:
+    """Whether the watchdog observer AND every emitter thread are still alive.
+
+    The watchdog observer thread itself can stay alive after an internal
+    emitter reader thread has crashed (the inotify ``_wd_for_path`` KeyError
+    under heavy create/delete churn, or the Windows
+    ``ReadDirectoryChangesW`` equivalent). A dead emitter silently stops
+    delivering events while ``daemon status`` keeps reporting the child
+    process as alive, so watch() must treat emitter death as fatal (#811).
+    """
+    if not observer.is_alive():
+        return False
+    return all(emitter.is_alive() for emitter in observer.emitters)
+
+
 def watch(
     repo_root: Path,
     store: GraphStore,
@@ -1574,6 +1589,15 @@ def watch(
         while True:
             _time.sleep(1)
             handler.raise_if_failed()
+            if not _observer_threads_alive(observer):
+                # Emitter-thread death is silent for the process: poll-based
+                # daemon health checks cannot see it. Exit the child so the
+                # daemon restarts watch instead of serving a frozen graph (#811).
+                logger.critical(
+                    "watch observer thread died; exiting so the daemon can "
+                    "restart watch (see issue #811)"
+                )
+                raise SystemExit(1)
     except KeyboardInterrupt:
         observer.stop()
     finally:
--- a/tests/test_incremental.py
+++ b/tests/test_incremental.py
@@ -1813,3 +1813,37 @@ class TestRenamePurgeParity:
             assert store.get_nodes_by_file(str(tmp_path / "b.py"))
         finally:
             store.close()
+
+
+class TestWatchObserverLiveness:
+    """#811: an emitter-thread death must exit watch(), not freeze silently."""
+
+    @staticmethod
+    def _fake_observer(observer_alive: bool, emitter_alive: bool):
+        class _Emitter:
+            def __init__(self, alive: bool) -> None:
+                self._alive = alive
+
+            def is_alive(self) -> bool:
+                return self._alive
+
+        class _Shim:
+            is_alive = staticmethod(lambda: observer_alive)
+            emitters = [_Emitter(emitter_alive)]
+
+        return _Shim()
+
+    def test_alive_observer_with_alive_emitter_is_healthy(self):
+        from code_review_graph.incremental import _observer_threads_alive
+
+        assert _observer_threads_alive(self._fake_observer(True, True))
+
+    def test_dead_observer_is_not_healthy(self):
+        from code_review_graph.incremental import _observer_threads_alive
+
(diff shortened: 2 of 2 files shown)


tirth8205:
Superseded by the inode-aware supervisor and reader/emitter recovery already introduced in #890; no code from this PR was retained. The narrower #891 case, where a thread starts and dies before its first liveness check, was fixed separately on `integration/token-efficiency-hardening` as `d5eb1bc`. That separate correction retains the existing one-repair policy and the exclusion of threads that have never started.

tirth8205:
This no longer merges into `staging`. Conflicts in `code_review_graph/incremental.py`, `tests/test_incremental.py`.

```bash
# resolve, then
uv run pytest tests/ -q
git push
```

When resolving:
- Rebase onto origin/staging and resolve conflicts in code_review_graph/incremental.py and tests/test_incremental.py.
- Reconcile with the `_WatchSupervisor.check_liveness()` mechanism already on staging, which detects dead observer/emitter threads, writes dead_threads into the health file, and raises RuntimeError -- the PR's `_observer_threads_alive` helper plus `raise SystemExit(1)` duplicates that logic and would...
- If any part of the PR survives the rebase (e.g. stricter emitter-thread inspection than check_liveness performs), add tests against `_WatchSupervisor.check_liveness` rather than a parallel helper, and cover the in-process `start_watch_thread` path where SystemExit is not caught by `except RuntimeErr...

tirth8205:

- `tests/test_incremental.py`

```bash
git fetch origin
git merge origin/staging
# resolve, then
uv run pytest tests/ -q
git push
```

PR #880, opened 2026-08-19
fix(typescript): resolve npm aliases to workspace packages

### LINKED ISSUE #343: Resolve npm-aliased imports (the `"alias": "npm:@scope/pkg@ver"` pattern) back to local library sources
## Summary

The graph's import resolver treats modules imported via an npm alias (the `"alias-name": "npm:@scope/real-package@version"` syntax in package.json) as external. As a result, `importers_of` / `callers_of` / `impact_radius` queries on files inside a locally-vendored package miss consumers that import via the alias — even when both the consumer and the library source are inside the graph's scanned tree.

## Reproducer (generic)

1. A multi-app repository with multiple consumer apps (`appA/`, `appB/`) sharing a locally-developed library (`sharedLib/`) — a common pattern for monorepos that also publish the shared library to a private registry.
2. `sharedLib/package.json` declares `"name": "@scope/shared"`.
3. Each consumer `appX/package.json` pins the library via the npm-alias syntax:
   ```json
   "dependencies": {
     "sharedLib": "npm:@scope/shared@^1.0.0"
   }
   ```
4. Consumer source imports look like:
   ```js
   import foo from 'sharedLib/some/module';
   ```
5. Build/update the graph on the repository root:
   ```
   code-review-graph build .
   ```
6. Query:
   ```
   query_graph_tool pattern=importers_of target=sharedLib/some/module.js
   ```
   **Expected:** returns `appA/src/entry.js` and similar consumers in `appB`.
   **Actual:** returns 0 results. The alias-based imports are not reflected as edges in the graph.

The same gap shows up in `get_impact_radius_tool`, `callers_of` on exported functions, and any downstream feature that relies on `IMPORTS_FROM` edges between consumer modules and the aliased library.

## Why it matters

Alias-based local vendoring is a common pattern in monorepos / multi-app repos:

- A shared UI or utility library is developed inside the repository.
- It's published to a private registry for production consumption.
- Consumer apps pin it via the `"alias": "npm:@scope/pkg@version"` syntax so their original import paths (`import X from 'alias/...'`) keep working after the publish-and-install cycle, without a repo-wide refactor.

For repos following this pattern, every cross-app / cross-module question about the shared library currently falls through the graph — users have to fall back to `grep` for any `importers_of` / `callers_of` query that crosses the alias boundary. That undermines the graph's main value proposition for exactly the codebases that would benefit most from structural analysis.

## Proposed behaviour

When building/updating the graph:

1. For each `package.json` in the scanned tree, scan `dependencies` and `devDependencies` for entries whose value starts with `npm:`. Parse out the `alias → real-package-name` mapping.
2. For each `npm:` entry where the `real-package-name` is also present in the graph's scanned tree (as a local `package.json` with a matching `name` field), record a local alias mapping: `alias-name → local-library-path`.
3. During import analysis, when the analyzer sees `import X from 'alias-name/<subpath>'`, resolve it against `local-library-path/<subpath>` rather than treating it as external. Emit the `IMPORTS_FROM` edge accordingly.

## Smaller first step

If package.json scanning is too invasive for a first pass, an explicit config option would unblock users with this pattern:

```yaml
# .code-review-graph.yml
alias_mappings:
  - alias: "sharedLib"
    local_path: "./packages/shared-library"
```

The resolver then treats imports of `sharedLib/*` as resolving inside `./packages/shared-library/*`. Manual config gives users control today; auto-detection from `package.json` can come later.

## Current workaround

Users of repos with this pattern fall back to `grep -rnE "import.*['\"]alias-name/..."` for alias-crossing queries, and use the graph only for intra-module traversal.

## Environment

- `code-review-graph` 2.3.2
- Installed via `uv tool install`
- MCP server used via Claude Code

Happy to contribute a PR if the approach above sounds right. LMK which direction you'd prefer (package.json auto-detection vs explicit config first).

yzxcj797:
## Summary
- Add a dedicated npm dependency-alias resolver for JavaScript and TypeScript imports.
- Read `npm:` aliases from dependency and dev-dependency manifests, then index uniquely named local packages in the repository.
- Resolve `alias/subpath`, root entry points, NodeNext `.js`-to-`.ts` spellings, and package index files to real workspace files.
- Keep absent and ambiguous packages external, prefer explicit tsconfig paths, reject traversal outside the selected package, and skip generated `node_modules` trees.

This makes `IMPORTS_FROM`, `importers_of`, and impact-radius queries cross the alias boundary when the real package source is present locally.

Fixes #343.

## Tests
- Added six regressions covering:
  - the monorepo subpath case from the issue;
  - a real full-build plus `importers_of` query;
  - root entry-point resolution through package metadata;
  - absent and ambiguous local packages remaining external;
  - rejection of traversal outside the package;
  - explicit tsconfig aliases taking precedence.

Validation:
- `pytest tests/test_npm_alias_resolver.py tests/test_tsconfig_resolver.py tests/test_commonjs_imports.py -q` — 23 passed
- `ruff check code_review_graph/npm_alias_resolver.py code_review_graph/parser.py tests/test_npm_alias_resolver.py` — passed
- `git diff --check` — passed

### DIFF
--- a/code_review_graph/npm_alias_resolver.py
+++ b/code_review_graph/npm_alias_resolver.py
@@ -1,0 +1,258 @@
+"""Resolve npm dependency aliases to local workspace packages."""
+
+from __future__ import annotations
+
+import json
+import os
+from pathlib import Path
+from typing import Optional
+
+_PROBE_EXTENSIONS = (".ts", ".tsx", ".js", ".jsx", ".vue")
+_TYPESCRIPT_SOURCE_EXTENSIONS = {
+    ".js": (".ts", ".tsx"),
+    ".jsx": (".tsx",),
+    ".mjs": (".ts",),
+    ".cjs": (".ts",),
+}
+_DEPENDENCY_SECTIONS = (
+    "dependencies",
+    "devDependencies",
+    "optionalDependencies",
+    "peerDependencies",
+)
+# Generated dependency trees can contain thousands of manifests and duplicate
+# package names. npm aliases are useful to the graph only when they point at
+# source that is intentionally part of the repository.
+_SKIPPED_DIRECTORIES = {
+    ".git",
+    ".hg",
+    ".svn",
+    ".venv",
+    "node_modules",
+}
+
+
+class NpmAliasResolver:
+    """Resolve ``alias/subpath`` imports declared with npm aliases.
+
+    npm allows a dependency key to differ from its real package name:
+    ``"sharedLib": "npm:@scope/shared@^1.0.0"``. Source imports then use
+    ``sharedLib/...``. When the real package is also present as workspace
+    source, resolve the import to that source instead of storing an external,
+    unqueryable package string.
+    """
+
+    def __init__(self, repo_root: Optional[Path] = None) -> None:
+        self._repo_root = Path(repo_root).resolve() if repo_root is not None else None
+        self._manifest_cache: dict[Path, Optional[dict]] = {}
+        self._alias_cache: dict[Path, dict[str, str]] = {}
+        self._package_roots: Optional[dict[str, list[Path]]] = None
+
+    def resolve_alias(self, import_str: str, file_path: str) -> Optional[str]:
+        """Return a local source file for an npm-aliased import, or ``None``."""
+        if not import_str or import_str.startswith("."):
+            return None
+
+        aliases = self._aliases_for_file(file_path)
+        matching_alias = self._longest_matching_alias(import_str, aliases)
+        if matching_alias is None:
+            return None
+
+        real_name = _real_package_name(aliases[matching_alias])
+        if real_name is None:
+            return None
+
+        package_roots = self._local_package_roots()
+        candidates = package_roots.get(real_name, [])
+        if len(candidates) != 1:
+            # Absent and ambiguous packages must remain external rather than
+            # producing a confidently wrong edge to another local package.
+            return None
+
+        package_root = candidates[0]
+        subpath = import_str[len(matching_alias):].lstrip("/")
+        if not _safe_subpath(subpath):
+            return None
+
+        found = self._probe_package_path(package_root, subpath)
+        if found is None:
+            return None
+
+        try:
+            package_root_resolved = package_root.resolve()
+            found_resolved = found.resolve()
+            if package_root_resolved not in found_resolved.parents:
+                return None
+        except (OSError, ValueError, RuntimeError):
+            return None
+        return found_resolved.as_posix()
+
+    def _aliases_for_file(self, file_path: str) -> dict[str, str]:
+        """Collect npm aliases from the nearest package manifests."""
+        try:
+            start = Path(file_path).parent.resolve()
+        except (OSError, ValueError, RuntimeError):
+            return {}
+
+        cache_key = start
+        if cache_key in self._alias_cache:
+            return self._alias_cache[cache_key]
+
+        aliases: dict[str, str] = {}
+        current = start
+        while True:
+            manifest = self._read_manifest(current / "package.json")
+            if manifest is not None:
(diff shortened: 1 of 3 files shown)


tirth8205:
Changes required: an ordinary nearer dependency does not shadow a root npm alias, so the resolver invents a local import and call target. Put `{"dependencies": {"alias": "npm:shared@1.0.0"}}` in the root package.json, `{"dependencies": {"alias": "2.0.0"}}` in app/package.json and a local `shared` package at version 2.0.0 with `index.ts` exporting `help`; parsing app/caller.ts containing `import {help} from 'alias'; help();` with `CodeParser(root).parse_file` binds to that local package incorrectly. Respect nearest-package declarations and version/package provenance, leaving unsupported matches unresolved.

tirth8205:

- Test that a node_modules/<name>/package.json (and a dot-directory copy) is pruned from _local_package_roots so the real local package still resolves uniquely , this is a load-bearing guarantee the PR body advertises.
- Test the failure branches of _read_manifest / _probe_package_entry: invalid JSON, non-object JSON, >1MB manifest, both in the consumer's ancestor chain and inside the scanned package tree; assert no exception and the import stays external.
- Test symlink handling: a file inside the package symlinked to outside it must stay external (parents check), and a symlinked directory holding a package.json must not be indexed (followlinks=False).
- Test the index.* fallback for a package with no main/module (claimed in PR body as 'package index files').
- Test alias boundary matching: 'sharedLib-x/foo' must not match alias 'sharedLib'; a scoped alias key '@alias/lib' with a subpath must match.

tirth8205:
The bug from my 12 September review still reproduces. The branch has two commits, the original implementation and a staging merge, so neither review was acted on.

```
package.json         {"dependencies": {"alias": "npm:shared@1.0.0"}}
app/package.json     {"dependencies": {"alias": "2.0.0"}}
shared/package.json  {"name": "shared", "version": "2.0.0", "main": "index.ts"}
app/caller.ts        import {help} from 'alias'; help();
```

`CodeParser(root).parse_file` on `app/caller.ts` binds `alias` to `shared/index.ts` and writes two fabricated edges, an IMPORTS_FROM to that file and a CALLS to `shared/index.ts::help`. `app` declares `alias` as an ordinary dependency on the published package `alias` at 2.0.0. The nearest manifest governs and the import should stay external.

Both causes are in `code_review_graph/npm_alias_resolver.py`. `_aliases_for_file` at lines 106-115 records only specs starting with `npm:` and continues past every other spec, so a nearer ordinary declaration can never shadow an ancestor's npm alias. `_real_package_name` at line 198 drops the version range, so `npm:shared@1.0.0` binds to a local `shared` at 2.0.0.

The five tests I listed on 15 September are still missing. `tests/test_npm_alias_resolver.py` holds exactly the six tests from the original PR body, and grep finds nothing for node_modules pruning, symlinks, manifest failure branches, the `index.*` fallback, or alias boundary matching.

The resolver also reads only `main` and `module`, and only for the bare root import. It never reads the `exports` map, which is how modern packages declare subpath entry points. I built trpc/trpc on both branches. It has 93 real npm alias declarations and every one resolves to nothing, because `main` and `module` point at `./dist/*` build output absent from a clean checkout while subpaths live under `src/`. The graph came out identical, 7184 nodes and 57299 edges on both sides. Your fixtures pass because they put source at the package root with `main` pointing at a real `.ts` file.

The defensive design is right. Ambiguous and absent packages stay external at `resolve_alias` line 66, tsconfig paths win first, and the parents check at line 84 catches an absolute `main` such as `/etc/passwd`. Nothing is ever deleted. Builds got no slower, 25.7s and 29.2s on staging against 20.5s and 23.3s here.

To merge: nearest-manifest precedence so a closer ordinary declaration shadows an ancestor alias, a version or provenance check leaving a mismatch unresolved, and the five tests plus one for the reproducer above.

PR #881, opened 2026-08-19
fix(build): recover partial graphs before incremental updates

### LINKED ISSUE #803: Remaining commands still create an empty graph.db on first run; released-version poisoned graphs are not auto-healed (follow-up to #777 / #782)

yzxcj797:
## Summary
- Detect non-empty graphs that represent only a severe slice of the repository even when their stored VCS head makes an incremental base look valid.
- Use a cheap tracked-file ratio first, then confirm suspected graphs against the exact parseable inventory to avoid false positives in documentation- or asset-heavy repositories.
- Force one full rebuild for such graphs and expose `recovery_reason: partial_graph` in the build result.
- Mark `forget` results as intentionally incomplete and clear that marker on full builds, so explicit exclusions are not overridden by coverage recovery.

Fixes #803.

## Tests
- Added four regressions covering:
  - a one-of-eight synced partial graph being rebuilt;
  - a complete incremental graph remaining incremental;
  - many unknown tracked assets not being mistaken for poison;
  - the intentional `forget` marker disabling recovery.

Validation:
- `pytest tests/test_pr803_poisoned_graph_recovery.py -q` — 4 passed
- `pytest tests/test_tools.py::TestBuildPostprocess tests/test_cli_reconciliation.py tests/test_pr809_edges.py -q` — 51 passed
- `ruff check code_review_graph/tools/build.py code_review_graph/incremental.py code_review_graph/forget.py tests/test_pr803_poisoned_graph_recovery.py` — passed
- `git diff --check` — passed

### DIFF
--- a/code_review_graph/forget.py
+++ b/code_review_graph/forget.py
@@ -116,6 +116,7 @@ def forget_files(
     # 2. Drop the forgotten files' own nodes and edges.
     for file_path in targets:
         store.remove_file_data(file_path)
+    store.set_metadata("intentionally_incomplete", "1")
     # Persist deletions before store_file_nodes_edges() opens its own
     # explicit transaction (BEGIN IMMEDIATE) during the re-parse below.
     store.commit()
--- a/code_review_graph/incremental.py
+++ b/code_review_graph/incremental.py
@@ -1129,6 +1129,7 @@ def full_build(
 
     store.set_metadata("last_updated", time.strftime("%Y-%m-%dT%H:%M:%S"))
     store.set_metadata("last_build_type", "full")
+    store.set_metadata("intentionally_incomplete", "0")
     if not cpp_errors:
         store.set_metadata(_CPP_IDENTITY_METADATA_KEY, CPP_IDENTITY_VERSION)
     _store_vcs_metadata(repo_root, store)
--- a/code_review_graph/tools/build.py
+++ b/code_review_graph/tools/build.py
@@ -5,16 +5,67 @@ from __future__ import annotations
 import logging
 import sqlite3
 import time
+from pathlib import Path
 from typing import Any
 
 from ..incremental import (
+    collect_all_files,
     full_build,
+    get_all_tracked_files,
     incremental_update,
     resolve_incremental_base,
 )
+from ..parser import normalize_file_path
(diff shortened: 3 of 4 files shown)


tirth8205:
Changes required: automatic partial-graph recovery can overwrite a graph belonging to another repository before its root guard runs. Use `full_build` to populate a shared `GraphStore` named `graph.db` from repository A containing one Python file, set `last_build_type` to `incremental`, set `CRG_DATA_DIR` to that database's directory, then call `build_or_update_graph(repo_root=B, base='HEAD~1', postprocess='none')` with eight Python files in B; the low coverage ratio selects a full rebuild and replaces A's graph. Enforce repository identity before every automatic rebuild.

tirth8205:

```bash
# resolve, then
uv run pytest tests/ -q
git push
```

When resolving:
- Rebase PR onto current origin/staging and resolve the conflict in full_build() metadata block of code_review_graph/incremental.py (~line 1508): HEAD calls _store_cpp_identity_pending(store, cpp_errors) while the PR sets intentionally_incomplete=0 and conditionally sets _CPP_IDENTITY_METADATA_KEY; bo...
- After rebase, re-run pytest/ruff/mypy and re-evaluate edge-case coverage of tests/test_pr803_poisoned_graph_recovery.py against the rebased code.

tirth8205:

- `code_review_graph/tools/build.py`

```bash
git fetch origin
git merge origin/staging
# resolve, then
uv run pytest tests/ -q
git push
```

Issue #889, opened 2026-08-19
A relative --repo wipes the graph: reconciliation removes every file when the root spelling differs from the build


tirth8205:
Found while reviewing the #811 watch fix, but this is pre-existing on `main` and independent of that work.

## What happens

`build` stores file paths anchored to the root it was given. `_reconcile_stale_files` treats any stored path it cannot place under the root it is given as stale and deletes it. Those two roots are the same string only by luck, because the CLI never normalises `--repo`: `Path(args.repo)` appears without `.resolve()` at cli.py:287, :1580, :1607 and :1624.

So a graph built the normal way and then reconciled with `--repo .` loses every file.

## Reproduction, verified on clean main

```python
store = GraphStore(repo / ".code-review-graph" / "graph.db")

full_build(repo.resolve(), store)          # what `build` does
# -> files: ['app.py']

_reconcile_stale_files(Path("."), store)   # what `watch --repo .` does
# -> removed: ['app.py'], files: []
```

Output:

```
files after build (absolute root):   ['app.py']
reconcile removed:                   ['app.py']
files after reconcile (relative):    []
VERDICT: GRAPH WIPED
```

The full matrix of build spelling against watch spelling:

| build root | watch root | result |
|---|---|---|
| absolute | relative | **graph wiped** |
| relative | absolute | **graph wiped** |
| relative | relative | intact |
| absolute | absolute | intact |

## Why it is not a reconciliation bug

The stale-path rule is deliberate and #861 has tests for it: a path that cannot be placed under the repository root is genuinely orphaned and should go. The defect is that the root is allowed to arrive in two different spellings for the same directory. Weakening reconciliation would reintroduce the orphan rows #812 was about.

## Fix

Normalise `--repo` once, at the CLI boundary, for every command that accepts it, so a root is always absolute and symlink-resolved before it reaches the store. Anything that reconciles should also refuse to run, with a clear message, when the stored paths do not sit under the root it was handed, rather than silently deleting them.

Worth a guard test per command that accepts `--repo`, since the failure is silent and destroys the graph the user just built.

The branch for #811 resolves the root inside `watch()` and adds a refusal before reconciling, which covers the watch path. The CLI-wide normalisation is still owed, and `update --repo .` and `build --repo .` remain exposed.

yzxcj797:
[...]

PR #890, opened 2026-08-19
fix(watch): stop watching ignored trees and never stall silently

### LINKED ISSUE #811: watch: observer thread silently crashes on Maven target/ dir churn (KeyError in watchdog inotify) — daemon looks alive but stops updating

tirth8205:
Fixes #811. A `watch`-mode observer thread dies on directory churn inside ignored trees (a Maven `target/` under Surefire load), the process stays up, `status` keeps reporting healthy, and the graph silently stops updating. A commenter reported the same shape on Windows with a different backend, so this is not inotify-specific.

## Two problems

**Ignored trees were still watched.** `observer.schedule(handler, repo_root, recursive=True)` registers OS watches everywhere. Ignore patterns were only consulted inside the event handler, after the OS had already delivered the event, so churn in `target/` still reached the watcher at full rate.

**A dead observer was invisible.** The watch loop only re-raised failures from the project's own batch processor; `observer.is_alive()` was never consulted, and the daemon checked only `proc.poll()`.

## What changed

Watches are planned per directory: ignored trees are never registered, the repo root is watched non-recursively, and a subtree with nothing worth excluding still collapses to a single recursive watch, so a clean repo behaves as before. Directories that appear or disappear later are picked up from a per-tick listing. Liveness is checked each tick across the dispatcher, every emitter, and any reader thread an emitter owns, which covers both the inotify buffer thread and the Windows report. A death logs an error naming the threads and exits non-zero so the daemon restarts it. Watcher health is published to a file that `crg-daemon status` reads, adding a `Watcher` column and last-event age.

Measured on this repository, OS-level watches drop from 810 to 54, a 93% reduction, independently re-measured during review at 94%. The synthetic Maven monorepo from the issue drops from 365 to 13.

## This branch was rewritten after an adversarial review found it broke worse than the bug

The first implementation discovered new directories from watchdog `DirCreatedEvent`s. On macOS those never arrive. `FSEventsEmitter._is_recursive_event` compares `event.src_path` against the watch path for directory events, so every directory event for a child of a non-recursive watch is dropped before any handler sees it.

Verified with a bare watchdog probe, no project code involved:

```
control: file directly at the watch root -> delivered
mkdir child, then a file inside it       -> no events at all
rm -rf an existing subdir                -> no events at all
```

The effect would have been that any newly created top-level directory was never watched or indexed, permanently, and that `rm -rf src && mkdir src` left `src` dead. That is worse than the stall being fixed, because a stall is at least detectable. It survived the first round because every scheduling test used a fake observer.

Discovery is now polling-based and does not depend on directory events at all: one `scandir` per non-recursive watch per tick, measured at 0.076 ms on this repository and 0.465 ms in a worst case of 300 children under one parent, against a one-second tick.

Five further findings from the same review are fixed here:

- An ordinary `rm -rf` of a watched directory made its emitter stop itself, which read as a crash and exited 1 into an unbacked-off 30-second daemon restart loop, each iteration paying a full rebuild. A dead thread whose watch root no longer exists is now a release, not a death, and the daemon gained exponential backoff, a restart counter surfaced in `status`, and a reset once a watcher stays up.
- Watch-budget exhaustion silently stopped adopting new directories. It now promotes to a single recursive watch on the parent, so coverage is never reduced, and reports `degraded` in health and `partial` in status.
- Health was only published after the initial build, so a large repo reported a healthy watcher as stalled for minutes. An initial heartbeat is now published up front, and SIGTERM unwinds cleanly so `crg-daemon stop` no longer leaves a stale health file.
- The nested build-output heuristic affects indexing, not just watching, so a real Python package named `build` beside a gradle file was dropped from the graph with only a debug-level log. Exclusions are now reported at info level during a build, and `!path` in `.code-review-graphignore` opts a single path out.
- `observer.unschedule` joins an emitter with no timeout on the watch-loop thread, which would hang the loop given the very wedge this change assumes possible. It is now time-boxed.

## A pre-existing data-loss bug surfaced while fixing the relative-root handling

`watch --repo .` was deleting every file from the graph at startup. `build` stores paths under the root it was given, and `_reconcile_stale_files` deletes any stored path it cannot place under the root it is given, and the CLI never normalises `--repo`. Verified on clean main, independent of this branch:

```
files after build (absolute root):   ['app.py']
reconcile removed:                   ['app.py']
files after reconcile (relative):    []
```

The reconciliation rule itself is deliberate and #861 has tests for it, so it was not weakened. `watch()` now resolves its root before reconciling, and refuses with a clear error when stored paths do not sit under the root it was handed rather than silently deleting them. The CLI-wide normalisation, which `update --repo .` and `build --repo .` still need, is filed as #889.

## Tests

48 in `tests/test_watch_robustness.py`, including a real `watchdog.Observer` driven through the new-top-level-directory case, since fakes are what hid the critical bug. Both key regressions were shown to fail before the fix and pass after, using a temporary stub that was removed before committing:

```
$ CRG_TEMP_DISABLE_SYNC=1 pytest tests/test_watch_robustness.py::TestRealObserver -q
1 failed, 1 passed
AssertionError: a top-level directory created after startup was never watched or indexed
```

The dead-observer path was also verified against a real live FSEvents emitter rather than a mock: killing it produced an error naming the dead thread, exit code 1, and a health file showing `observer_alive: false`, with `crg-daemon status` printing `stalled`.

End-to-end run with a relative `--repo .`: a directory created after startup is watched and indexed, `rm -rf src && mkdir src` is re-adopted, the watcher survives the delete, 200 create/delete cycles inside nested `target/` produce no events, and SIGTERM removes the health file.

Full suite 2944 passed, 9 skipped, 2 xpassed. ruff and mypy clean.


tirth8205:
Holding this. A re-review found that delete-and-recreate of a watched directory, the `rm -rf src && mkdir src` case this PR lists as fixed, kills the watcher and then silently drops the recreated contents.

`dead_threads()` separates a release from a death by asking whether the root still exists. When a directory is deleted and recreated inside one tick, it does, so the emitter that stopped itself on the delete counts as a genuine death and the watcher exits 1. `sync_watches` cannot recover because it keys watches by path string, so a recreated directory looks identical to an untouched one.

Reproduced 3 of 3 at the shipped 1 second tick, through the CLI, and with two ordinary back-to-back `git checkout`s that drop and restore a directory. With a 3 second gap it is clean, so the window is one tick.

The silent half is worse than the crash: after the daemon restarts, the startup reconcile removes stale rows but never adds anything, so the recreated directory's files stay out of the graph until something touches them again or a full build runs, while status reports ok.

Being fixed by keying watch identity on `(st_dev, st_ino)` rather than path string, with a regression test that drives a real Observer through delete-then-recreate. The suite covers both halves separately today (one on a fake observer whose emitter list is empty so nothing can die, one on a real observer that never recreates); their interaction is what fails.

Everything else in the re-review held up: the polling rewrite genuinely fixes the macOS blindness for new, moved and deeply nested directories, no test-only bypass remains in the committed code, the 94 percent watch reduction was independently re-measured, and budget promotion, health timing, SIGTERM cleanup, the `!path` opt-out, the unschedule time-box and the graph-wipe guard are all verified fixed.

PR #893, opened 2026-08-20
fix(watch): accept resolved paths for symlinked roots

### LINKED ISSUE #892: watch silently drops every event when the repo is reached through a symlinked path
Found while building a test harness for the #811 watch work. Pre-existing (present since 802c7f7 / 62d3775) and independent of that branch.

## What happens

`_relative_path` compares incoming event paths against `os.path.abspath(repo_root)`, which normalises `..` and `.` but does not resolve symlinks. On macOS the FSEvents backend reports paths in their fully resolved form, so a repository reached through a symlinked path produces events whose paths never match the root the watcher is comparing against.

The classic case is anything under `/var`, which is a symlink to `/private/var`, so every temporary directory on macOS hits it. It also affects the common setup where a developer keeps checkouts under a symlinked home or project directory.

## Effect

Every real filesystem event is dropped. The watcher looks completely healthy: the observer thread is alive, the heartbeat is fresh, `crg-daemon status` reports `ok`, and the graph simply never updates.

Directory discovery still works, because that path is now driven by polling rather than by events, so new directories are adopted and their contents indexed while ordinary edits to existing files are silently ignored. That combination is particularly confusing to diagnose: the graph is not frozen, it is partially updating.

## Fix direction

Resolve symlinks on both sides of the comparison. `daemon.watch_health_path` already keys by `os.path.realpath`, so the same normalisation applied to the watcher's root and to incoming event paths would make the two consistent.

Care is needed not to break the case where a repository legitimately contains a symlink to a directory outside itself, which the ignore rules deliberately exclude today.

Worth a test that watches a repository through a symlinked path and asserts an edit to an existing file reaches the graph.

Related: #889 covers the other half of the root-spelling problem, where a relative `--repo` and an absolute build root disagree and reconciliation deletes the graph. Both come from the same root cause, that a repository root can arrive in several spellings and nothing normalises it at one boundary.

yzxcj797:
## Summary
- Match resolved backend event paths against the resolved repository root when watching through a symlinked root.
- Keep lexical-root handling and the existing in-repository symlink and escape checks.
- Add a regression that uses a directory symlink (or junction on Windows) and verifies a resolved event updates the graph.

Fixes #892.

## Testing
- Pristine main: the new regression failed before the fix (the resolved event was dropped and no node was stored).
- ```text
  uv run pytest tests/test_incremental.py -q --deselect tests/test_incremental.py::TestWatchReconciliation::test_watch_symlink_events_are_rejected --deselect tests/test_incremental.py::TestWatchReconciliation::test_watch_rejects_file_below_symlinked_ancestor_outside_repo
  106 passed, 2 deselected
  ```
- The two deselected pre-existing tests require unprivileged symlink creation on this Windows host; the new junction-based regression ran in the selected suite.
- ```text
  uv run ruff check code_review_graph/incremental.py tests/test_incremental.py
  git diff --check origin/main..HEAD
  ```

### DIFF
--- a/code_review_graph/incremental.py
+++ b/code_review_graph/incremental.py
@@ -1407,7 +1407,17 @@ def _create_watch_handler(
             try:
                 relative = candidate.relative_to(lexical_root)
             except ValueError:
-                return None
+                if lexical_root == resolved_root:
+                    return None
+                # Some backends—notably FSEvents on macOS—report the resolved
+                # path even when the watcher was scheduled through a symlinked
+                # repository root. Retry against the resolved root so both
+                # spellings refer to the same watched tree.
+                candidate = Path(os.path.realpath(path))
+                try:
+                    relative = candidate.relative_to(resolved_root)
+                except ValueError:
+                    return None
             existing = candidate
             while not existing.exists() and existing != lexical_root:
                 existing = existing.parent
--- a/tests/test_incremental.py
+++ b/tests/test_incremental.py
@@ -1564,6 +1564,39 @@ class TestWatchReconciliation:
         finally:
             store.close()
 
+    def test_watch_accepts_resolved_event_paths_for_symlinked_repo_root(self, tmp_path):
+        import os
+        import sys
+
+        repo = tmp_path / "repo"
+        repo.mkdir()
+        source = repo / "source.py"
+        source.write_text("def source():\n    pass\n")
+        linked_repo = tmp_path / "repo-link"
+        if sys.platform == "win32":
+            import _winapi
+
+            _winapi.CreateJunction(str(repo), str(linked_repo))
+        else:
+            linked_repo.symlink_to(repo, target_is_directory=True)
+
+        store = GraphStore(repo / "graph.db")
+        handler = _create_watch_handler(linked_repo, store, None)
+        try:
+            from watchdog.events import FileModifiedEvent
+
+            with patch(
+                "code_review_graph.incremental.collect_all_files",
+                side_effect=AssertionError("resolved root event inventoried repository"),
+            ):
+                handler.process([FileModifiedEvent(str(source))])
+
+            assert store.get_nodes_by_file(str(linked_repo / "source.py"))
+        finally:
+            store.close()
+            if sys.platform == "win32" and linked_repo.exists():
+                os.rmdir(linked_repo)
+
     def test_watch_rejects_file_below_symlinked_ancestor_outside_repo(self, tmp_path):
         outside = tmp_path.parent / f"{tmp_path.name}-outside"
         outside.mkdir()


tirth8205:
Thanks — the bug is real and I reproduced it end to end (real Observer, repo reached through `/var/folders/...`, edit dropped before the change, picked up after), and I could not break the security invariants: in-repo symlinks pointing outside, symlink chains, `../` escapes, a sibling directory sharing the root's prefix, and a root symlink re-pointed after the handler was built are all still rejected, on both spellings. That part is solid.

What holds it up is the layer.

**The flow in the issue title is still broken after this patch.** `watch()` normalizes its root with `os.path.abspath` (`incremental.py:2386`) while every other entry point resolves — `_validate_repo_root` does `Path(path).resolve()`, and the daemon does `.expanduser().resolve()`. That asymmetry is the actual bug; `_relative_path` is just where it shows up. With this PR merged onto current `main`, `crg watch --repo <symlinked path>` never reaches your new branch, because the graph-root check that landed with #890 fires first:

```
$ crg build --repo /var/folders/.../repo    # resolves, stores /private/var/... paths
$ crg watch --repo /var/folders/.../repo
Error: the graph holds 1 file(s) such as '/private/var/.../repo/mod.py', none of them under
'/var/folders/.../repo'; it was built with a different repository root.
```

I tested the one-line alternative — `os.path.realpath` instead of `os.path.abspath` in `watch()`, nothing else changed — on `main`:

```
$ crg watch --repo /var/folders/.../repo    # starts, picks up the edit
nodes: ['alpha', 'beta']
```

That fixes both halves: the `serve --auto-watch` path your patch targets (`main.py` passes `Path(repo_root)` unresolved) and the CLI path it doesn't, without relaxing `_relative_path` at all. It's also what #892 itself suggests — "the same normalisation applied to the watcher's root".

**The new test pins the fix shape rather than the reported symptom.** It calls `_create_watch_handler(linked_repo, ...)` directly, so it passes only for a handler-level fix and fails for a root-level one even when the user-visible bug is gone. A test that drives `watch()` (or the `serve` auto-watch path) through a symlinked root would be implementation-agnostic and would cover the CLI half too.

**Sequencing note:** #894 canonicalizes every path-valued `--repo` at the CLI boundary, which overlaps with this. Whichever lands first should own the normalization — I don't want two mechanisms resolving roots in different places.

Smaller points, worth addressing whichever direction you take:

- The new branch has no security test of its own. The two existing rejection tests never enter it, because their event paths are already lexically under the root. Please add at least one escape attempt through the resolved spelling.
- The symlink-rejection invariant becomes spelling-dependent: with the root reached through a symlink, an event on an in-repo symlink is accepted and attributed to its target, while the same repo watched by its real name drops it (`test_watch_symlink_events_are_rejected`). Not an escape — the target is inside the repo and `_parseable_file` re-validates — but it should be a deliberate, tested choice.
- `os.path.realpath()` at `incremental.py:1416` sits inside an `except ValueError:` block and can itself raise `ValueError` (a NUL byte in the path). It escapes to `process()` and kills the watch loop with `RuntimeError: watch update failed`, where `main` merely drops the event. Practically unreachable, but it belongs in its own guard.
- The junction half of the new test runs nowhere in CI: the `Windows daemon and file handles` job lists its test files explicitly and `tests/test_incremental.py` isn't among them. So "verified on Windows" rests entirely on your local run.

Gates are green on the PR head and after merging current `main` (2953 passed, ruff and mypy clean), and all 13 checks pass — the objection is design, not correctness.

tirth8205:
Follow-up on the sequencing note above, now that I've finished reviewing #894.

I merged #894 into current `main` locally (resolving its conflict with #890) and re-ran the #892 repro end to end: `crg watch --repo <symlinked path>` starts and indexes the edit, and the `serve --auto-watch` path is covered too, because `watch()` there canonicalizes with `.expanduser().resolve()` before anything compares paths. No change to `_relative_path` was involved.

So #894 fixes both halves of #892 at the root, and this PR's event-path branch becomes unnecessary once it lands. I'd rather not carry two mechanisms that resolve roots in different places, so I'm going to hold this one until #894 is merged, then close #892 against it — unless you find a case that survives root canonicalization and still needs the handler-level fix, in which case show me the repro and I'll reopen the question.

Your adversarial-escape coverage was useful either way; if you'd like to land something from this branch, a test that drives `watch()` through a symlinked root (rather than `_create_watch_handler` directly) would be welcome on its own, since nothing covers that flow today.

tirth8205:
Closing as superseded: #894 landed (via #899) and fixes both halves of #892 at the root, so the event-path branch here is no longer needed — I verified on main that a symlinked root is watched and indexed correctly with no change to `_relative_path`.

Your adversarial escape probes were genuinely useful and I kept the conclusions. Two pieces of this work are still wanted if you'd like to land them separately: a test that drives `watch()` (not `_create_watch_handler`) through a symlinked root, since nothing covers that flow today, and the `OSError` guard around the ancestor walk — I filed that one as #897. Thanks for the careful work.

PR #894, opened 2026-08-20
fix: preserve graphs across repository root spellings

### LINKED ISSUE #889: A relative --repo wipes the graph: reconciliation removes every file when the root spelling differs from the build

yzxcj797:
## Summary
- Canonicalize every path-valued `--repo` argument once at the CLI boundary, while leaving eval config names and daemon log aliases untouched.
- Resolve repository roots at the full-build, incremental-update, and watch entry points so equivalent relative and absolute spellings share one graph identity.
- Refuse an incremental reconciliation whose authoritative File nodes all belong to a different root, without weakening orphan-only cleanup from #861.

Fixes #889.

## Testing
- Pristine main regression: `incremental_update(Path("."), ...)` removed the sole graph file (`stale_files_removed == 1`); the new regression now retains it.
- `uv run pytest tests/test_repo_root_identity.py -q` — 33 passed.
- `uv run pytest tests/test_repo_root_identity.py tests/test_cli_reconciliation.py tests/test_cli.py tests/test_incremental.py tests/test_integration_git.py -q` with four host-specific pre-existing tests deselected — 208 passed, 1 skipped.
- `uv run pytest tests/test_pr861_edges.py -q` — 6 passed and the same 4 Windows path-separator failures reproduced on pristine main.
- `uv run ruff check code_review_graph/cli.py code_review_graph/graph.py code_review_graph/incremental.py tests/test_cli.py tests/test_repo_root_identity.py`
- `uv run python -m py_compile ...`
- `git diff --check origin/main..HEAD`
- End-to-end check: built with `--repo .`, then read status with an absolute `--repo`; the graph retained its one file.

### DIFF
--- a/code_review_graph/cli.py
+++ b/code_review_graph/cli.py
@@ -541,6 +541,40 @@ _GRAPH_TOOL_COMMANDS = {
 }
 
 
+_PATH_REPO_COMMANDS = frozenset({
+    "install",
+    "init",
+    "uninstall",
+    "build",
+    "update",
+    "postprocess",
+    "embed",
+    "watch",
+    "status",
+    "forget",
+    "visualize",
+    "wiki",
+    "detect-changes",
+    "dead-code",
+    "serve",
+    "mcp",
+    *_GRAPH_TOOL_COMMANDS,
+})
+
+
+def _canonicalize_repo_argument(args: argparse.Namespace) -> None:
+    """Canonicalize path-valued ``--repo`` arguments in place.
+
+    Commands whose ``--repo`` value is a repository *name* rather than a path
+    (eval configs and daemon log aliases) are deliberately excluded. Every
+    path consumer receives the same absolute, symlink-resolved spelling before
+    it opens a database or compares stored paths.
+    """
+    repo = getattr(args, "repo", None)
+    if args.command in _PATH_REPO_COMMANDS and repo:
+        args.repo = str(Path(repo).expanduser().resolve())
+
+
 def _find_explicit_repo_root(start: Path) -> "Path | None":
     """Resolve an explicit --repo for graph-tool commands.
 
@@ -1327,6 +1361,8 @@ def main() -> None:
         _print_banner()
         return
 
+    _canonicalize_repo_argument(args)
+
     if (
         args.command == "refactor"
         and args.mode == "rename"
@@ -1765,6 +1801,9 @@ def main() -> None:
                     postprocess=pp,
                     **embedding_refresh_kwargs,
                 )
+            except RuntimeError as exc:
+                print(f"Error: {exc}", file=sys.stderr)
+                sys.exit(1)
             finally:
                 logging.disable(previous_disable)
             nodes = result.get("total_nodes", 0)
--- a/code_review_graph/graph.py
+++ b/code_review_graph/graph.py
@@ -1199,6 +1199,17 @@ class GraphStore:
         ).fetchall()
         return [r["file_path"] for r in rows]
 
+    def get_file_marker_paths(self) -> list[str]:
+        """Return paths that have authoritative File nodes.
+
+        Reconciliation uses this to distinguish a graph anchored to another
+        repository from orphan Function/Edge rows left by a failed update.
+        """
+        rows = self._conn.execute(
+            "SELECT file_path FROM nodes WHERE kind = 'File' ORDER BY file_path"
+        ).fetchall()
+        return [row["file_path"] for row in rows]
+
     def search_nodes(self, query: str, limit: int = 20) -> list[GraphNode]:
         """Keyword search across node names.
 
--- a/code_review_graph/incremental.py
+++ b/code_review_graph/incremental.py
@@ -941,6 +941,30 @@ def _reconcile_stale_files(
     return stale_files
 
 
+def _assert_graph_matches_root(repo_root: Path, store: GraphStore) -> None:
+    """Refuse an incremental reconciliation anchored to a different root.
+
+    Authoritative File nodes identify the root a graph was built with. If none
+    of them are under the requested root, treating every row as stale is more
+    likely to destroy a usable graph than to clean up orphan rows. Orphan-only
+    databases have no File markers and retain the purge behavior from #861.
+    """
+    file_paths = store.get_file_marker_paths()
+    if not file_paths:
+        return
+    prefix = normalize_file_path(repo_root)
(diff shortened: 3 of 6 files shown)


tirth8205:
Reviewed in depth — the approach is right and I want this in. Canonicalizing at the three reconciliation choke points rather than only at the CLI boundary is the correct call: it covers library callers like `start_watch_thread` from `serve --auto-watch`, which never resolves its root. The new tests genuinely fail without the source change (I reverted `incremental.py` alone and got the two expected failures, and disabling the `_canonicalize_repo_argument` call alone reds 8 tests across three files), path-traversal invariants are untouched, and the #861 orphan cleanup still works on orphan-only graphs.

**It can't merge as-is, though — #890 landed in the meantime and the two collide.**

There's a textual conflict in `watch()` (both branches rewrote the root-normalization lines), and behind it a worse one: we both added a function named `_assert_graph_matches_root`, with the parameters in **opposite order** — yours is `(repo_root, store)`, mine is `(store, repo_root)`. Python takes the later definition, so the call sites silently disagree. After resolving the textual conflict:

```
code_review_graph/incremental.py:2339: error: Name "_assert_graph_matches_root" already defined on line 1112
code_review_graph/incremental.py:2425: error: Argument 1 has incompatible type "GraphStore"; expected "Path"
47 failed, 2938 passed
AttributeError("'PosixPath' object has no attribute 'get_all_files'")
```

I tested the resolution so you don't have to guess at it. Keep yours, drop mine:

1. In `watch()`, keep the `_WatchSupervisor` block from `main` but normalize with `repo_root = _canonical_repo_root(repo_root)` instead of `Path(os.path.abspath(repo_root))`.
2. Delete `main`'s `_assert_graph_matches_root(store, repo_root)` definition and its call in `watch()`. Yours sits at the `incremental_update` choke point, which `watch()` goes through anyway, so it subsumes mine.

That gives `2985 passed, 9 skipped, 2 xpassed`, ruff and mypy clean.

**One thing worth knowing:** with that merge in place, `crg watch --repo <symlinked path>` starts and indexes edits correctly — I verified it end to end on a `/var/folders/...` root. That is the CLI half of #892, which #893 is currently trying to fix at the event-path layer. Your change fixes both halves at the root, so please keep the `.resolve()` in `_canonical_repo_root` — it's load-bearing for more than #889.

Follow-ups, none blocking:

- The refusal message says "Rebuild it, or retry with the root it was built with". Since `_canonicalize_repo_argument` always makes the root absolute, the second option is unreachable through the CLI for a legacy graph holding relative paths. Only "rebuild" is actionable — I'd drop the clause.
- The guard uses `any(... startswith(prefix) ...)`, so a single in-root `File` node re-enables the purge for every other row. A half-migrated graph (mixed absolute and legacy relative rows) still loses its legacy rows silently. I couldn't reach that state through the CLI, since `watch`/`update` refuse before any per-event write, so it's a residual hole rather than a live bug — but the docstring should say the guard is all-or-nothing.
- The PR title says spellings are preserved; on a case-insensitive filesystem a case-differing root (`/x/OLDREPO` vs `/x/oldrepo`) is refused rather than preserved, because `Path.resolve()` doesn't case-fold. Refusing is the right safety outcome — just worth saying so in the docstring rather than implying all spellings converge.
- `daemon.py` respawns a dead watcher every 30s with no backoff, so a repo with a legacy graph will now respawn-and-die in a loop instead of wiping once and running. Worth a separate issue.
- `except RuntimeError` around `build_or_update_graph` (`cli.py:1804`) swallows unrelated `RuntimeError`s into a bare `Error: <msg>`. A dedicated exception type for the root mismatch would be tighter.
- `_PATH_REPO_COMMANDS` is a hand-maintained allowlist, and the test parametrizes over the constant itself, so an omission can't be caught. I audited it against the parser — it's complete and correct today — but an opt-*out* set (`{"eval", "daemon"}`) would stay correct on its own.
- Please add `tests/test_repo_root_identity.py` to the Test Structure list in `CLAUDE.md`.

Merge `main` in, apply 1–2 above, and I'll take it.

tirth8205:
Correction to one bullet above: the daemon does already back off. `_restart_delay` doubles the delay per consecutive failure from `_RESTART_BACKOFF_BASE` (30s) up to `_RESTART_BACKOFF_MAX` (900s) and resets once a watcher has stayed up for `_RESTART_HEALTHY_SECONDS`, so a repo with a legacy graph will not respawn every 30s forever. Ignore that item — nothing for you to do there. The rest of the review stands.

PR #904, opened 2026-08-22
fix(python): resolve imports whose package name is not the on-disk layout

### LINKED ISSUE #903: [Bug]: tests_for returns 0 for Python — bare-name fallback compares file paths against dotted module names

## Symptom

```
  status: ok | result_count: 0

  status: ok | result_count: 2
```

## Root cause

```
```

```python
supported = [
    qualified
]
```

```
odoo
odoo.exceptions
```

## Reproduction

## Related

  identical defect.
  language.

## Suggested direction

rsat:
Fixes #903.

## The gap

`resolve_python_imports` indexes every path suffix of every indexed Python file,
then looks each import up by its **full** module name:

```python
candidates = sorted(modules.get(raw_module, ()))
```

That works whenever the package name matches the repository layout — `src`
layouts resolve because `mypkg.runner` is a literal suffix of
`src/mypkg/runner.py`, which is what `test_src_layout_imports_resolve_before_test_coverage`
covers.

It does not work when a framework assembles the package at runtime. Odoo serves
`odoo.addons.foo.bar` from `addons/<any>/foo/bar.py`, and no directory named
`odoo` exists in the repository at all, so the key is never present and the
import is recorded `unresolved`.

Everything downstream inherits that. `_resolve_bare_endpoints` has no import
evidence to match bare `TESTED_BY` sources against, so
`_select_evidence_backed_candidate` returns `None` and the edge is dropped —
which surfaces to the user as `tests_for` reporting covered code as uncovered,
silently. Django app registries and plugin namespaces land in the same place.

## Why not just strip segments until something matches

That was my first cut, and it was wrong. An import of a module that is simply
not indexed keeps shortening until it finds *some* file sharing its last two
segments:

```
odoo.addons.web.controllers.utils  ->  addons/enterprise/sale_renting/controllers/utils.py
odoo.addons.base.models.ir_qweb    ->  addons/enterprise/web_studio/models/ir_qweb.py
```

`web` and `base` ship with the framework and are not in the repository. On the
same corpus that produced the numbers below, 20 of 326 resolutions were
fabricated this way. A confidently wrong edge is worse than a missing one, so
that approach is not in this PR — but it is in the tests, as a regression.

## The change

The prefix has to be attested. Count how many distinct imports each candidate
prefix would resolve uniquely, and accept only those supported well enough to be
a namespace rather than a coincidence. A real one is shared by everything the
framework loads, which puts it far above the noise — on the corpus below,
`odoo.addons` scores 305 against 6 for the runner-up, and is the only prefix
accepted.

Once a prefix is established the remainder is looked up as is. No match means
unresolved; it never falls through to a shorter, less specific suffix. Ambiguity
is still the caller's call: more than one candidate is reported ambiguous, as
before.

## Measurements

Odoo 18 repository, 9,891 files, 921 distinct dotted imports:

| | before | after |
|---|---:|---:|
| resolved | 0 | **305** |
| ambiguous | 0 | 87 |
| unresolved | 921 | 529 |

The 529 that remain are genuinely external — `odoo` itself, stdlib,
site-packages — and correctly stay unresolved.

Correctness was checked, not just volume: for every resolution of an
`odoo.addons.X...` import, the destination file must live under a directory
named `X`. All 305 do. The naive version scored 306 coherent and 20 fabricated;
this one trades that single extra resolution for eliminating all 20.

## Tests

`tests/test_python_namespace_imports.py`, eight cases: two end-to-end (imports
resolve; coverage becomes visible through `get_transitive_tests`) and six on the
boundaries — a prefix is learned only when attested, a single coincidence
establishes nothing, an unindexed module stays unresolved, ambiguity is
preserved, an unattested prefix resolves nothing, and a one-segment remainder is
never evidence.

Full suite: **2996 passed, 5 skipped, 1 failed**. The failure is
`test_windows_server_still_prewarms_before_mcp_run`, which fails identically on
`upstream/main` without this change (`NameError: windows_events` — a Windows
path exercised on macOS). `ruff check code_review_graph/` is clean.

## Prior art

#799 fixed the same namespace-versus-path mismatch for C#, by mapping declared
namespaces back to the files declaring them. Python has no declared namespace to
map, since the module name is derived from the path — hence inferring the
prefix from repeated evidence instead of a lookup table.

### DIFF
--- a/code_review_graph/python_resolver.py
+++ b/code_review_graph/python_resolver.py
@@ -4,16 +4,109 @@ from __future__ import annotations
 
 import json
 import logging
+from collections import Counter
 from typing import TYPE_CHECKING
 
 if TYPE_CHECKING:
+    from collections.abc import Iterable
+
     from .graph import GraphStore
 
 logger = logging.getLogger(__name__)
 
+# A one-segment suffix ("exceptions") matches far too many unrelated files to
+# be evidence of anything. Two is the shortest suffix that still carries the
+# defining directory, and uniqueness is enforced on top of it.
+_MIN_SUFFIX_SEGMENTS = 2
+
+# Floor for the attestation count, so a two-file repository can still establish
+# a prefix while a single coincidence cannot.
+_MIN_PREFIX_EVIDENCE = 2
+
+
+def learn_synthetic_prefixes(
+    modules: dict[str, set[str]], raw_modules: Iterable[str],
+) -> set[str]:
+    """Infer which leading segments name a package the repository never spells.
+
+    ``modules`` is keyed by every path suffix of every indexed file, so an
+    import is found when the package name matches the repository layout. It is
+    not when a framework assembles the package at runtime: Odoo serves
+    ``odoo.addons.foo.bar`` from ``addons/<any>/foo/bar.py``, and no directory
+    named ``odoo`` exists in the repository at all.
+
+    Stripping whatever prefix happens to make a suffix match is not safe -- an
+    import of a module that simply is not indexed will find some unrelated file
+    that shares its last two segments. So the prefix has to earn it: count how
+    many distinct imports each candidate prefix would resolve uniquely, and keep
+    only those attested well enough to be a real namespace rather than a
+    coincidence. A genuine one is shared by everything the framework loads,
+    which puts it orders of magnitude above the noise.
+    """
+    votes: Counter[str] = Counter()
+    for raw_module in raw_modules:
+        segments = raw_module.split(".")
+        for cut in range(1, len(segments) - _MIN_SUFFIX_SEGMENTS + 1):
+            if len(modules.get(".".join(segments[cut:]), ())) == 1:
+                votes[".".join(segments[:cut])] += 1
+    if not votes:
+        return set()
+    strongest = max(votes.values())
+    threshold = max(_MIN_PREFIX_EVIDENCE, strongest // 4)
+    return {prefix for prefix, count in votes.items() if count >= threshold}
+
+
+def _candidates_behind_prefix(
+    modules: dict[str, set[str]], raw_module: str, prefixes: set[str],
+) -> set[str]:
+    """Look the import up with an attested synthetic prefix removed.
+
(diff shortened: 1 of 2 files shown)


rsat:
Force-pushed a rewrite of the approach — flagging it in case anyone read the first version.

The original commit stripped leading segments until a suffix matched. Auditing the resolutions rather than counting them showed that fabricates edges: an import of a module that is not indexed (`odoo.addons.web.controllers.utils`, where `web` ships with the framework) keeps shortening until it finds an unrelated `controllers/utils.py` that happens to be unique. 20 of 326 resolutions were wrong that way.

The prefix is now inferred from repeated evidence instead, and the remainder is looked up as is. Same corpus: 305 resolutions, all verified to land under the directory the import names, and none fabricated. The old behaviour is pinned as a regression test.

rsat:
Control run, since the change only makes sense if it stays inert where there is nothing to infer.

Built the graph for **this repository** — a flat package, no synthetic namespace — and ran the inference over it:

```
imports pending    : 190
prefixes learned   : none
edges added by fix : 0
```

Nothing is learned and nothing is resolved, so repositories whose layout already matches their package names are unaffected. `src` layouts keep working through the existing exact-match path (`test_src_layout_imports_resolve_before_test_coverage` still passes).

One limitation worth stating: the attestation threshold is `max(2, strongest // 4)`, calibrated against a corpus where the signal is unambiguous (305 attestations vs 6 for the runner-up). A repository assembling **two** synthetic namespaces of very different sizes could have the smaller one fall below the cut and stay unresolved — conservative in the safe direction, but a real limit. Happy to change the rule if you would rather it be absolute, or configurable.

rsat:
Retiro este PR: dejamos de usar la herramienta, así que el fork se va y con él la rama de origen.

El diagnóstico de #903 queda en pie por si le sirve a alguien — el resolver de imports busca el nombre de módulo completo, y cuando el paquete no coincide con el layout del disco (un framework que lo arma en runtime) nunca acierta. Medido: 0 de 921 imports resueltos en ese repositorio. Gracias.

Issue #311, opened 2026-04-17
How to configure code-review-graph for a workspace of sibling repos + worktrees (Claude Code in VS Code)?


noamsuissa01:
I'm using Claude Code in VS Code with a workspace that is not a single git repo — it's a directory (/Users/me/github/org-workspace/) containing ~15 sibling git repos and their active git worktrees. The VS Code extension always launches Claude with cwd = org-workspace/, which I want to keep so I can jump between repos freely. I've registered 4 repos in ~/.code-review-graph/registry.json and each has a correctly built .code-review-graph/graph.db. 

**The problem:** .mcp.json runs uvx code-review-graph serve with no --repo, so _resolve_repo_root() falls back to cwd (the non-git workspace root), the PostToolUse code-review-graph update --skip-flows hook does the same, and that builds a junk 12-file graph at the workspace root that all tool calls without an explicit repo_root hit by default. cross_repo_search_tool works across the registry, but the other 20+ tools still need repo_root on every call. 

What's the recommended setup for this topology — outer git repo whose real purpose is to host many nested registered sibling repos + worktrees, IDE-fixed cwd on the outer — so that 

> (a) tool calls without explicit repo_root route sensibly (fan out over the registry, or infer from the edited/queried file path) instead of defaulting to the outer repo, 
> (b) the update hook writes to the correct nested graph, 
> (c) worktrees auto-resolve from their main repo instead of needing individual registration, and 
> (d) the outer repo can be excluded from serving entirely (is there a flag, or is the intended pattern to run one serve per registered repo)? 

Running code-review-graph 2.3.2 on macOS with the Claude Code VS Code extension.

admin-wal:
+1!!

souravjamwal77:
Yes, this is required. I have a directoty called `work-dir` in which I have below file to open my different projects at the same time in a workspace.

current.code-workspace
```
{ "folders": [
  { "path": "./../project-1" }
,
  { "path": "./../project-2" }
,
  { "path": "./../project-3" }
,
  { "path": "./../project-4" }
], "settings": {} }
```
Then, I simply to `code current.code-workspace` and that opens my required projects in the same IDE workspace. But, the code-review-graph MCP and other tools don't work there. 

MCP error. 
<img width="906" height="183" alt="Image" src="https://github.com/user-attachments/assets/767ccd6e-ff03-494f-ac17-7336a4056023" />

However, MCP is shown available in the IDE settings.

tirth8205:
[...]

PR #915, opened 2026-08-24
feat(registry): scope cross_repo_search to a subset of the registry

### LINKED ISSUE #914: [Feature]: scope cross_repo_search to a subset of the registry
### Problem

`cross_repo_search_tool` always fans out over the whole registry, and there is no way to
narrow it. That is the right default when the registry holds one product, but registries
grow: mine currently holds 10 repositories belonging to four unrelated products, and every
cross-repo query pays for all of them.

A concrete measurement from my own registry. One query, `kind=Function`, `limit=3`,
`max_results=8`, against 10 registered repos: 23 merged results across 10 repos, and the top
of the list is taken by helpers from products that have nothing to do with the question,
while the two functions I was actually looking for rank below them.

Nothing here is a ranking bug. The rank-interleave merge added for #619 does exactly what it
says, and `_MAX_CROSS_REPO_RESULTS` correctly bounds the response. But both of those bound
the *size* of the result set, not its *relevance*: a repository that has nothing to do with
the question still contributes its best local rank to the top of the merged list. As the
registry grows, the share of the answer that can possibly be useful shrinks, and every
irrelevant hit is context an agent has to read and discard.

The workspace topology in #311 makes this the normal case rather than an edge case: one
IDE-fixed cwd over many sibling repos, all registered, so cross-repo search is the primary
entry point rather than an occasional one.

### Proposed solution

An optional `repos` parameter on `cross_repo_search`, on both `cross_repo_search_func` and
the MCP tool:

```python
cross_repo_search(
    query: str,
    repos: list[str] | None = None,   # new
    kind: str | None = None,
    limit: int = 20,
    max_results: int = 50,
)
```

Semantics, chosen to keep the change as small as it can be:

- Omitted or empty: current behaviour exactly, every registered repo is searched. Fully
  backward compatible, no existing caller changes.
- Provided: entries are matched against the registry by alias first, then by the path's
  folder name, and only the matching repos are searched. Registry order is preserved, so
  the existing rank-interleave and its tie-breaker are untouched.
- Names that match no registry entry are returned in an `unknown` field rather than
  silently dropped, so a typo or an unregistered repo is visible in the response instead of
  looking like "that repo has no matches". This mirrors the existing behaviour for a repo
  whose graph DB is missing, which is currently skipped without any signal.

Same measurement as above with `repos` set to the three repos of the product in question:
5 results across 3 repos, all of them from that product, and both functions I wanted in the
top three. Same query, same graphs, same ranking code.

I am happy to send the PR, with tests, if the shape looks right to you.

### Affected area

MCP tools / server

### Alternatives considered

- **Raise `limit` / `max_results`.** Returns more of everything, including more noise. It
  bounds response size, which is already handled, not relevance.
- **A second registry via `CRG_HOME` and a second server on another port.** Works, but it
  costs a second resident process with its own embedding model loaded, and the scope is
  then fixed at process start rather than chosen per call. For a single laptop that
  resident cost is the whole objection.
- **A wrapper that registers an extra MCP tool on the same FastMCP instance before serving.**
  This is what I run locally today and it is why I can quote the numbers above. It works,
  but it is a private tool sitting beside the real one, and it has to be re-checked on every
  upgrade of this package. That fragility is exactly why I would rather the filter lived
  upstream, on the existing tool, than in my launcher.
- **Per-call `repo_root` on `semantic_search_nodes`.** This is the documented answer today
  and it is correct, but it is one call per repository, and the caller then merges the
  results by hand without the rank-interleave the cross-repo tool already implements.

### Additional context

Related: #311 (sibling-repo workspace topology, where this is the normal case), #607
(multi-repo HTTP server), #619 (the cross-repo merge this proposal deliberately leaves
unchanged).

mohammad-rj:
Closes #914

## Problem

`cross_repo_search` fans out over the whole registry with no way to narrow it. On a registry
of 10 repositories spanning four unrelated products, one query returned 23 results across all
10, with the two functions it was actually about ranked below hits from products that could
not answer it. The rank-interleave from #619 and `_MAX_CROSS_REPO_RESULTS` bound the size of
that response, not its relevance. #311 makes this the normal case rather than an edge case:
one IDE-fixed cwd over many sibling repositories, all registered.

## Change

- Optional `repos` on `cross_repo_search_func` and `cross_repo_search_tool`. Omitted or empty
  keeps current behaviour exactly.
- Names match an entry's alias first, then its folder name; only those repositories are opened.
- Selection preserves registry order, so the merge tie-breaker cannot be changed by reordering
  the argument.
- Names matching no entry come back in `unknown` rather than dropped, and a selection matching
  nothing searches nothing rather than the full registry.

Same query scoped to one product's three repositories: 5 results across 3, both wanted
functions in the top three.

## Tests

Nine new tests: four on the selection helper, four on the func end to end, one on the MCP tool
forwarding the argument. Full suite passes; ruff and mypy clean. Docs updated in `README.md`,
`docs/COMMANDS.md` and the tool docstrings.


tirth8205:
Good feature and the right shape — a pure filter over the already-trusted registry, backward compatible to the byte (no argument, `None` and `[]` all return the identical unscoped response), and I could not get it to reach anything outside the registry: absolute paths, `../` and a path to a genuinely registered repo all land in `unknown` with `repos_searched: []` and no sqlite file opened. `_select_repos` can only ever return a subset of `list_repos()`, which is the property that matters.

One thing blocks it, and it goes straight at what this project is for.

**`unknown` and the summary are the only unbounded fields in the tool.** `results` goes through `_bounded()`; `unknown = [name for name in names if name not in by_name]` echoes every unmatched name verbatim, and the summary interpolates the raw list a second time (`f"No registered repository matches {repos}."`). Measured with this repo's own counter against a six-repo registry:

```
unscoped baseline                    1,068 tokens   truncated=False
documented worst (limit=10**6)       1,068 tokens   truncated=False
repos=["z"*200]*500                 50,917 tokens   truncated=False   0 results
repos=["z"*500]*1000               251,792 tokens   truncated=False   0 results
```

`BUDGETS["cross_repo_search_tool"]` sets `worst_max: 30_000` and `ABSOLUTE_MAX_TOKENS = 50_000` is documented as the ceiling no tool may exceed with every knob maxed. The last row is 8× the tool's own budget and 5× the absolute backstop, returns nothing, and reports `truncated: False` — so the response actively claims nothing was cut. That also makes two sentences in `docs/COMMANDS.md:319` false for this tool.

The budget test can't catch it, which is worth knowing: `tests/conftest.py` points `CRG_HOME` at an empty temp directory, so the tool short-circuits at `if not all_repos:` before any of the new code runs. Adding `"repos": ["z"*200]*500` to that tool's `worst` kwargs still passes. Please `_bounded()` the `unknown` list with an `unknown_total`, put a count rather than `{repos}` in the summary, and register the `repos` knob against a fixture that has at least one repo registered.

**Second, colliding names silently search a subset.** `by_name` is one flat namespace mixing aliases and folder names, built with `setdefault`, and `wanted` is a set of indices — so a name resolves to exactly one entry and the earliest registry position wins:

- `/repos/android` (alias `droid`) and `/repos/nested/android` (alias `tv`): `repos=["android"]` searches only `droid`, with `unknown: []`.
- `[{"path":"/a/web"}, {"path":"/b/frontend","alias":"web"}]`: `repos=["web"]` selects `/a/web` — an explicit alias loses to another entry's incidental folder name.

That is the sibling-repo topology #914 and #311 use as the motivating example (`~/work/acme/api` and `~/work/beta/api`), and it is the exact failure mode the `unknown` field exists to prevent: fewer repositories searched than asked for, with no signal. Map name → list of indices and select all matches, and resolve every alias before any folder name so an explicit alias always wins. An `ambiguous` field would be even better.

Smaller ones, worth folding into the same push:

- A dropped name never reaches `summary`. `repos=["billing","nope"]` gives `status: "ok"` and `Found 1 result(s) across 1 repo(s)`, with `nope` only in the JSON — while the all-unknown branch *does* put it in the summary. CLAUDE.md trains agents to read `summary` first, so the two branches should be consistent.
- The echoed names skip `_sanitize_name`. A name containing NUL, BEL and newlines comes back intact in both `summary` and `unknown`, while every other name in the same response is sanitized. Low practical risk since it reflects the caller's own input, but its 256-character cap is also the natural cap for the first finding.
- `repos` is exact-case and rejects paths, while `registry.resolve_repo()` and `unregister()` both take alias-or-path. The strictness is the safe direction — just say so in the MCP docstring and `docs/COMMANDS.md`.
- Six other docs still say "across all registered repositories" (`docs/USAGE.md:130`, `docs/FEATURES.md:54`, and the four translated READMEs).

Gates are green — 2994 passed, ruff and mypy clean, 13/13 checks — so this is about bounds and resolution semantics, not correctness of the plumbing.

mohammad-rj:
All of it is addressed in the amended commit.

**Bounds.** `unknown` and `ambiguous` both go through `_bounded()` now (cap 20), with `unknown_total` / `ambiguous_total` and `unknown_truncated` / `ambiguous_truncated`, and every echoed name passes through `_sanitize_name` first, so each entry is capped at 256 characters like every other name in a response. Neither summary branch interpolates the caller's list any more; both carry counts instead. The last row of your table measures 1,087 tokens where it measured 251,792.

**The budget test.** The entry now carries `needs_registry`, which registers the fixture repo into the temporary `CRG_HOME` before the call, plus a `FLOOD_NAMES` knob resolving to one name that matches and 500 that do not. The tool no longer short-circuits on the empty registry, so the ceiling binds against exactly the field you found.

**Resolution.** `_select_repos` maps a name to a list of indices and selects every match; a name matching more than one entry is reported in `ambiguous`. Aliases are resolved across the whole registry before any folder name, and only an explicit alias enters the alias namespace, so an entry registered without one cannot outrank another entry's deliberate alias. Both of your examples are tests now: two sibling checkouts sharing `api` search both and report `ambiguous: ["api"]`; `[{"path":"/a/web"}, {"path":"/b/frontend","alias":"web"}]` with `repos=["web"]` selects `/b/frontend`.

**Docs.** The MCP docstring and `docs/COMMANDS.md` state that matching is exact, case-sensitive and rejects paths, and describe the new fields. `docs/USAGE.md`, `docs/FEATURES.md` and the four translated READMEs no longer say "across all registered repositories".

Verification: full suite 3,003 passed, coverage 85.06%, ruff and mypy clean.

tirth8205:
Integrated on `integration/token-efficiency-hardening` as `dae2711`. Cross-repository search can now select registered aliases or folder names without unrelated repositories consuming the compact result budget. The selection retains alias precedence, registry order and ambiguous matches, with bounded diagnostics; the registry and token-budget regressions passed.

tirth8205:
Merged into `staging` through the integration PR #988 (merge commit 3030ca8). Your change is carried by commit dae2711, which keeps you as the author (Mohammad Rahimi Jafari). It will be promoted staging → testing → main and released from there. Thank you for the contribution! Closing this PR since its content is now on staging.

PR #916, opened 2026-08-24
feat(java): resolve Spring AOP @annotation/execution pointcuts to CALLS edges

### LINKED ISSUE #592: feat(java): support Spring AOP (@Aspect/@Around/@Before) via pointcut-based CALLS edges
## Summary

Spring AOP uses `@Aspect` classes with pointcut expressions to intercept method calls at runtime. These relationships are currently invisible in CRG.

```java
@Aspect
@Component
public class LoggingAspect {
    @Around("execution(* com.example.service.*.*(..))")
    public Object logAround(ProceedingJoinPoint joinPoint) throws Throwable {
        // wraps every method in com.example.service
    }
}
```

## Proposed approach

Emit `CALLS` edges from aspect methods to matched target methods by parsing AspectJ pointcut expressions.

**Parse phase**:
- Detect `@Aspect` class
- Extract pointcut strings from `@Around` / `@Before` / `@After` / `@AfterReturning` / `@AfterThrowing`
- Parse the AspectJ `execution()` pattern and match against method nodes in the graph

## Complexity considerations

AspectJ pointcut expressions are a rich DSL (wildcards, boolean operators, annotations, named references). Options:

1. **Regex approximation** — handle `execution()` pattern covering ~80% of real-world cases; explicitly unsupported for `call()`, `within()`, named pointcuts
2. **External library** — depend on an AspectJ parsing library if license compatible
3. **Metadata-only** — emit a node `ASPECT::LoggingAspect` with the pointcut string in `extra`, no CALLS edges; lets users query via `get_knowledge_gaps`

Option 1 is the most pragmatic starting point and suitable as a PR. Happy to implement if the maintainers are interested in this direction.

## Related PRs (same Spring annotation series)

- #589 — Lombok `@RequiredArgsConstructor`/`@AllArgsConstructor` DI semantics fix
- #590 — `@Scheduled` TRIGGERS edges
- #591 — `@EventListener` HANDLES/PUBLISHES/CALLS edges

YutenC:
Closes #592 (Option 1 — regex-based approximation).

## Summary

`@Aspect` classes are currently invisible to CRG: an advice method (`@Before`/`@Around`/...) and the business methods its pointcut actually intercepts show up as two unrelated nodes, because the interception happens through a runtime Spring AOP proxy, not a static method call in the advice's source text.

This adds a post-build resolver, `aop_resolver.py`, that reads the pointcut expression each advice annotation already carries (`extra["decorators"]`, captured verbatim by the existing parser) and matches it against candidate target methods, emitting `CALLS` edges tagged `extra.aop_resolved = true`.

## Scope (Option 1, as discussed in the issue)

Resolved:
- `@annotation(some.pkg.Foo)` pointcuts — matched by annotation bare name against methods and class-level stereotypes.
- `execution(...)` pointcuts — matched via a hand-rolled AspectJ-pattern-to-regex converter (`..` → `.*`, `*` → `[^.]*` within a segment). Only the `ClassName.methodName` portion is matched; the declaring-type's package segment is dropped, because parsed nodes aren't persisted with a resolved Java package. This is a known, documented approximation — see the module docstring for the false-positive/negative tradeoff.
- Same-class named `@Pointcut` references (`@Before("pointcutName()")`, including named-argument form `@AfterThrowing(pointcut = "...", throwing = "...")`).

Explicitly out of scope for this PR (left for a follow-up, and called out in `uncertainty.py`'s narrowed AOP gap note):
- `within()`, `args()`, `target()`, `this()` pointcut designators.
- Cross-aspect named pointcut references (e.g. `Other.pointcut()`).
- Compound boolean pointcuts (`&&`, `||`) — detected up front and skipped wholesale rather than partially parsed, to avoid emitting a wrong edge.

Reuses the existing `CALLS` edge kind (tagged via `extra`) rather than introducing a new one, so `callers_of`/`callees_of`/`get_impact_radius` pick up advice→target edges automatically without touching their edge-kind allowlists. Two dedicated patterns, `advises` / `advised_by`, are added on top so a query can ask for just the AOP-derived relationships instead of every caller regardless of provenance (mirroring the existing `triggers_of`/`triggered_by` and `publishers_of`/`listeners_of` pairs).

Every resolved edge carries `confidence: 0.7, confidence_tier: "INFERRED"` plus `pointcut_kind`/`pointcut_expr`/`advice_kind` for debugging, distinguishing it from statically-extracted edges.

`uncertainty.py`'s java `LanguageGap` note is narrowed from "aop advice and reflective invocation are not statically traced" to name only what's still actually unresolved after this change (the out-of-scope pointcut styles above, plus reflective `Method.invoke` calls, folded into one note since `gap_note()` returns only the first match per `(language, patterns)`).

## Testing

- New `tests/test_aop_resolver.py` (6 cases): `@annotation` + `@Before`, `execution(...)` + `@Around`, same-class named-pointcut reference resolving two advices, compound `within() && !execution()` safely skipped with zero edges/no exception, end-to-end `callers_of` via `GraphStore` + `query_graph`, and `advises`/`advised_by` correctly filtering out a plain (non-AOP) `CALLS` edge that `callers_of` still returns.
- Full suite: `uv run pytest tests/ --tb=short -q` → 2991 passed, 0 failed (9 skipped / 2 xpassed are pre-existing, unrelated).
- `uv run ruff check code_review_graph/` → clean.
- `uv run mypy code_review_graph/ --ignore-missing-imports --no-strict-optional` → no issues.
- Manually validated against real `@Aspect` classes from a production Spring Boot codebase (custom `AccessLimitAspect` using `@Pointcut("@annotation(...))")` + `@Before`/`@AfterThrowing`, same named-pointcut pattern this PR's tests mirror): `callers_of` on the `@AccessLimit`-annotated target method went from 0 results before this change to correctly finding both advice methods after, with `advised_by` isolating just those two from an unrelated plain caller.

## Not included

Filed as scoped-out follow-ups per the issue's own option list, not attempted here: `within(Type+)` (needs a new recursive `INHERITS` traversal — no existing helper to build on) and `args()`/`target()`/`this()` (depend on runtime object identity, not resolvable by static analysis at all).

### DIFF
--- a/code_review_graph/aop_resolver.py
+++ b/code_review_graph/aop_resolver.py
@@ -21,11 +21,15 @@ Resolution chain:
 Scope (see issue #592 — Option 1, regex-based approximation):
 
 * ``@annotation(some.pkg.Foo)`` pointcuts are matched by simple annotation
-  name (``Foo``) against ``extra["decorators"]`` (methods) and
-  ``extra["spring_annotations"]`` (classes, matching every method in that
-  class). This is exact enough for the common case, but two same-named
+  name (``Foo``) against ``extra["decorators"]`` on the candidate method
+  itself. This is exact enough for the common case, but two same-named
   annotations in different packages are indistinguishable — a known
-  approximation, consistent with the rest of Option 1.
+  approximation, consistent with the rest of Option 1. A *class-level*
+  ``extra["spring_annotations"]`` match (e.g. a class-level
+  ``@Transactional``) is intentionally **not** expanded to every method in
+  that class — that is ``@within()`` semantics, a different pointcut
+  designator, and conflating the two produced widespread false positives
+  on common class-level stereotype annotations (see PR #916 review).
 * ``execution(...)`` pointcuts are converted into a regex applied only to
   the ``ClassName.methodName`` portion of a candidate — the return-type
   pattern, parameter pattern (``(..)``), and any package portion of the
@@ -35,9 +39,19 @@ Scope (see issue #592 — Option 1, regex-based approximation):
   that one file); matching therefore is a strict subset of full AspectJ
   semantics and **can produce both false positives and false negatives**.
   This is deliberate and documented, not an oversight — see issue #592.
-* Pointcut expressions combining multiple sub-expressions with ``&&``,
-  ``||`` (or a bare ``!``) are compound boolean pointcuts and are skipped
-  entirely rather than partially parsed, to avoid emitting a wrong edge.
+  When dropping the package portion reduces the pattern to something that
+  would match virtually every method (e.g. ``com.foo.service.*.*`` losing
+  its package prefix down to a bare ``*.*``), the expression is treated as
+  unresolvable rather than emitting a nearly-unfiltered edge set — see
+  ``_parse_execution_expression``.
+* Pointcut expressions combining multiple sub-expressions with ``&&`` or
+  ``||`` are compound boolean pointcuts and are skipped entirely rather
+  than partially parsed, to avoid emitting a wrong edge. This check also
+  applies to the expression a same-class named ``@Pointcut`` reference
+  resolves to, not just the reference text itself. A bare ``!`` negation
+  is not recognized by either pointcut-kind matcher below, so it is
+  skipped too — just by falling through unparsed rather than via an
+  explicit compound-operator check.
 * ``within(Type+)``, ``args()``/``target()``/``this()``, and named pointcut
   references that cross an ``@Aspect`` class boundary are out of scope for
   this pass (see the narrower AOP ``LanguageGap`` note in
@@ -82,6 +96,13 @@ _ANNOTATION_POINTCUT = re.compile(r"^\s*@annotation\s*\((.+)\)\s*$", re.DOTALL)
 _EXECUTION_POINTCUT = re.compile(r"^\s*execution\s*\((.*)\)\s*$", re.DOTALL)
 
 _DERIVED_FLAG = "aop_resolved"
+
+# What target_pattern reduces to when every segment collapsed to a bare
+# wildcard — i.e. the package portion was the only thing that made the
+# original execution() pattern selective. Resolving these would match
+# nearly every method in the codebase instead of the intended scope, so
+# _parse_execution_expression treats them as unresolvable.
+_UNIVERSAL_REGEXES = frozenset({"^[^.]*$", r"^[^.]*\.[^.]*$"})
 
 
 def _extract_annotation_string_arg(deco_text: str) -> Optional[str]:
@@ -147,7 +168,9 @@ def _parse_execution_expression(expr: str) -> Optional[tuple[str, bool]]:
     method name.  Only the last one or two dot-separated segments of the
     declaring-type+method pattern are used — see module docstring for why
     the package portion is dropped.  Returns ``None`` when the expression
-    cannot be safely reduced (e.g. no parameter-pattern parens found).
+    cannot be safely reduced (e.g. no parameter-pattern parens found), or
+    when discarding the package portion left a pattern that matches
+    virtually every method (see ``_UNIVERSAL_REGEXES`` below).
     """
     m = _EXECUTION_POINTCUT.match(expr)
     if not m:
@@ -175,6 +198,8 @@ def _parse_execution_expression(expr: str) -> Optional[tuple[str, bool]]:
     if not target_pattern:
         return None
     regex_str = "^" + _aspectj_pattern_to_regex(target_pattern) + "$"
+    if regex_str in _UNIVERSAL_REGEXES:
+        return None
     return regex_str, needs_class
 
 
@@ -186,7 +211,11 @@ def _resolve_pointcut_expr(
     Returns the expression to actually parse (``expr`` itself when it is
     already an inline expression), or ``None`` when ``expr`` is a
     cross-aspect reference, an unresolvable reference, or a compound boolean
-    expression — all out of scope for this pass.
+    expression — all out of scope for this pass. The compound check is
+    applied both to ``expr`` and, when ``expr`` is a named reference, to the
+    ``@Pointcut`` body it resolves to — a bare reference like
+    ``"myPointcut()"`` carries no ``&&``/``||`` itself even when the
+    pointcut it names does.
     """
     if any(op in expr for op in _COMPOUND_OPERATORS):
         return None
@@ -194,7 +223,10 @@ def _resolve_pointcut_expr(
     ref = _NAMED_POINTCUT_REF.match(expr)
     if ref:
         name = ref.group(1)
-        return pointcuts.get(name)
+        resolved = pointcuts.get(name)
+        if resolved is None or any(op in resolved for op in _COMPOUND_OPERATORS):
+            return None
+        return resolved
 
     if "." in expr and "(" in expr and not expr.lstrip().startswith(
         ("execution(", "@annotation(", "within(", "args(", "target(", "this("),
@@ -230,7 +262,12 @@ def resolve_aop_advice(store: "GraphStore") -> dict:
         ).fetchall()
     }
     if not java_files:
-        return {"aspects_indexed": 0, "advice_resolved": 0, "calls_created": 0}
+        return {
+            "aspects_indexed": 0,
+            "advice_resolved": 0,
+            "calls_created": 0,
+            "stale_calls_removed": 0,
+        }
 
     # -------------------------------------------------------------------
     # Clear previously derived edges so stale advice→target links from a
@@ -249,25 +286,26 @@ def resolve_aop_advice(store: "GraphStore") -> dict:
     # Index @Aspect classes and every Java method, grouped by (class, file).
     # -------------------------------------------------------------------
     aspect_classes: list[dict] = []
-    all_classes: list[dict] = []
     for row in conn.execute(
         "SELECT name, qualified_name, file_path, extra FROM nodes "
         "WHERE kind = 'Class' AND language = 'java'"
     ).fetchall():
         extra = _load_extra(row["extra"])
-        entry = {
-            "name": row["name"],
-            "qualified_name": row["qualified_name"],
-            "file_path": row["file_path"],
-            "spring_annotations": extra.get("spring_annotations") or [],
-        }
-        all_classes.append(entry)
-        if "Aspect" in entry["spring_annotations"]:
-            aspect_classes.append(entry)
+        if "Aspect" in (extra.get("spring_annotations") or []):
+            aspect_classes.append({
+                "name": row["name"],
+                "qualified_name": row["qualified_name"],
+                "file_path": row["file_path"],
+            })
 
     if not aspect_classes:
         logger.info("AOP resolver: no @Aspect classes found, skipping")
-        return {"aspects_indexed": 0, "advice_resolved": 0, "calls_created": 0}
+        return {
+            "aspects_indexed": 0,
+            "advice_resolved": 0,
+            "calls_created": 0,
+            "stale_calls_removed": len(stale_ids),
+        }
 
     all_methods: list[dict] = []
     methods_by_class_file: dict[tuple[str, str], list[dict]] = {}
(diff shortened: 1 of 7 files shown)


YutenC:
Pushed a follow-up commit adding direct unit tests for \`_aspectj_pattern_to_regex\` and \`_parse_execution_expression\` — the two functions with the most edge-case-sensitive logic in \`aop_resolver.py\`, flagged as untested by this repo's own dogfooded review bot on the first push. They were previously only exercised indirectly through \`resolve_aop_advice()\`'s integration-style tests.

One of the new tests (\`test_double_dot_becomes_a_permissive_any_match\`) intentionally pins the current behavior of \`..\` compiling to a bare \`.*\` (matches any characters, not only well-formed dot-separated segments) rather than asserting a stricter reading of the docstring's "any number of intervening path segments" wording — flagging this here in case it's worth tightening in a follow-up, but didn't want to change matching behavior in this PR without discussion.

tirth8205:
Real work here, and the parts that are usually got wrong are right: the resolver follows `event_resolver`'s shape exactly (same signature, same `_DERIVED_FLAG` clear-and-rebuild idempotency, same wiring point), the edges carry `confidence: 0.7` / `INFERRED` matching the convention for derived relationships, and genuine EXTRACTED calls are not clobbered because AOP edges use `line=0` and dedup on `(kind, source, target, file_path, line)`. Re-runs are stable.

The pointcut matcher is what blocks it.

**1. Package scope is discarded, so the canonical #592 pointcut matches the entire repository.** `_parse_execution_expression` keeps only the last two dot-segments:

```python
target_pattern = f"{segments[-2]}.{segments[-1]}"
```

so `com.example.service.*.*` reduces to `*.*` — the same regex as `execution(* *(..))`. On a fixture of 20 `com.example.service` plus 20 `com.example.repo` classes (610 methods), one `@Around("execution(* com.example.service.*.*(..))")` matched **609 of 610 methods**: 300 correct, 300 in the `repo` package the pointcut explicitly excludes, and 9 other aspects' advice methods. Precision 49%. `callers_of(Repo0.repoMethod5)` then returns a caller where the truth is none.

The matcher is accurate when a pointcut names a class — it is maximally wrong in exactly the package-scoped form the issue was filed about. Returning `None` when the reduced regex is universal (`^[^.]*$` or `^[^.]*\.[^.]*$`) is about four lines, drops the fixture from 649 edges to 40 with 100% precision, and leaves all 14 of your tests passing.

**2. `@annotation(X)` is implemented as `@within(X)`.** A class-level annotation expands to every method of the class:

```
TxAspect.beforeTx -> BillingService.charge      (class-level @Transactional — wrong)
TxAspect.beforeTx -> BillingService.refund      (wrong)
TxAspect.beforeTx -> ReportService.writeReport  (correct, method-level)
```

`@Transactional` on a class is everywhere in Spring, so this fires constantly. Either drop the class-level expansion or label it as `@within`, which is a different designator.

**3. The compound guard is bypassed by named references.** The `&&`/`||` check runs on the reference text, not the resolved body, so `@Pointcut("execution(* *(..)) || execution(* run(..))")` referenced by name silently narrows to `^run$` instead of being skipped. The PR says compounds are "detected up front and skipped wholesale rather than partially parsed" — true inline, false through a named reference.

**4. Fan-out makes incremental updates unusable, and it follows from (1).** Edges ≈ broad advices × all Java methods. At ~3,015 methods:

```
 3 broad advices ->  3,209 edges, resolver 0.63s
 8 broad advices -> 18,284 edges, resolver 6.13s
13 broad advices -> 33,409 edges, resolver 8.63s   (full build without it: 0.71s)
```

Synthetic edges outnumber real ones 10:1, and because the gate is `any(path.endswith(".java"))`, a one-line edit to a single unrelated Java file costs:

```
incremental_update: without resolver 0.49s -> with resolver 12.01s  (25x)
```

The graph auto-updates on save via hooks, so that is 12 seconds on every Java keystroke-save in a 3k-method fixture; real Spring monoliths are an order of magnitude bigger. Fixing (1) removes most of this, but the stale-clear also does `SELECT id, extra FROM edges WHERE kind='CALLS'` and JSON-parses every row in Python, which is why run #2 is slower than run #1 — worth narrowing.

**5. Edge direction answers the less useful question.** CALLS propagates target→source, so with advice→target edges, changing the advice that wraps 300 methods reports a blast radius of `(0, 0, 2)` — nothing — while changing an unrelated `repo` class falsely pulls in the aspect. For an impact tool, "I changed this aspect, what does it affect" is the query that matters.

Minor: `aop_resolver.py:233` and `:270` return early without `stale_calls_removed`, so the `:270` path reports 0 having just deleted rows; the module docstring says a bare `!` is treated as compound but `_COMPOUND_OPERATORS` is only `("&&", "||")`; and `query.py:261` doesn't list the new `advises`/`advised_by` patterns.

On tests: 14 pass, but two mutations also leave 14 passing — a hard `targets[:3]` cap, and rejecting universal regexes. Every `execution` test names an explicit class, so the wildcard form from #592 is unspecified. Please add the package-wildcard case, a class-level `@Transactional`, and a named compound reference.

Gates are green (2999 passed, ruff and mypy clean, 13/13 checks) — this is about precision, not plumbing. Fix 1, 2 and 3 and I think 4 largely disappears with them.

tirth8205:
Changes required: explicit package qualification is discarded during AOP matching. Create `a/Service.java` and `b/Service.java`, each declaring `public void run()`, plus an `@Aspect` advice annotated `@Before("execution(* a.Service.run(..))")`, then call `full_build(root, store)`; the advice receives CALLS edges to both services. Preserve package selectivity and verify the direction of advice-impact relationships before adding those edges to compact context.

YutenC:
Pushed `54b7a7e` addressing both the package-qualification bug and the edge-direction concern from the last two rounds.

## Scope

**1. Package selectivity (the `a.Service`/`b.Service` repro).** The root cause was that `_parse_execution_expression` kept only the last two dot-segments of the declaring-type pattern, so any package qualifier was discarded before matching — the `ea8ddd1` fix only caught the case where that discarding collapsed the pattern to a near-universal regex, not the general case.

Fixed by actually persisting the package instead of continuing to drop it: `_extract_classes` in `parser.py` now stores each Java class's package (`extra["java_package"]`, from the file's `package` declaration, which was previously only available transiently during that file's own parse) on the Class node. `_parse_execution_expression` no longer truncates the pattern — it keeps the full `package.Class.method` shape — and `resolve_aop_advice` matches it against each candidate's real `package.Class.method`, not a truncated `Class.method`.

Concretely, your repro (`a/Service.run` vs `b/Service.run`, pointcut naming only `a.Service.run`) now resolves to exactly one edge, into `a.Service.run`. This also removes the `_UNIVERSAL_REGEXES` safety net added in `ea8ddd1` — no longer needed, since a package-scoped pattern doesn't collapse to a near-universal one anymore — and, as a side effect, should substantially shrink the fan-out from point 4 of the first review: a `com.example.service.*.*`-style pointcut now only matches classes actually in `com.example.service`, not the whole codebase.

**2. Edge direction.** Agreed that `CALLS`'s default target→source impact policy answers the wrong question for these edges — "did the target change affect the aspect" instead of "did the aspect change affect its targets." Rather than reversing `source`/`target` on the edge itself (which would have inverted `callers_of`/`advises`/`advised_by` semantics and broken the tests built around them), I special-cased impact-radius specifically: both the SQL (`get_impact_radius_sql`) and NetworkX (`_build_networkx_graph`) engines now check `extra.aop_resolved` and force source→target propagation for just those edges, overriding `IMPACT_EDGE_DIRECTIONS["CALLS"]`. Changing an aspect's file now correctly shows its targets as impacted; changing a target's file no longer pulls the aspect in.

## Testing

- `tests/test_aop_resolver.py`: 21 cases (6 new/rewritten) — including your exact same-named-class/different-package reproduction, a package-wildcard test that now asserts it correctly stays scoped to its own package (previously asserted zero resolution, which is no longer the right behavior), and a new impact-radius direction test (advice-file change reaches the target; target-file change does not reach the advice).
- Full suite: `uv run pytest tests/ --tb=short -q` → 3006 passed, 9 skipped, 2 xpassed (same pre-existing skip/xpass set as before this commit, 0 failed).
- `uv run ruff check code_review_graph/` → clean.
- `uv run mypy code_review_graph/ --ignore-missing-imports --no-strict-optional` → no issues.

Everything else from the original review (annotation-vs-within, compound-guard-through-named-reference, `stale_calls_removed` early returns) is unchanged from `ea8ddd1`.

tirth8205:

- Test for the NetworkX impact engine: call store._get_impact_radius_networkx([aspect_path]) and assert the target is impacted and the reverse is not (mirrors the SQL-engine test and the parity pattern in tests/test_graph.py).
- Test that resolve_aop_advice is idempotent: run twice, assert stale_calls_removed equals the first run's calls_created and the edge set is unchanged.
- End-to-end full_build -> incremental_update test (like tests/test_spring_events.py::test_incremental_listener_change_removes_stale_event_call): change the pointcut (or remove the target's annotation) and assert the stale aop_resolved edge is gone and result['aop_resolution'] is present.
- Resolver-level tests for the skip branches: cross-aspect dotted reference, unresolvable named reference, advice with no string argument, and the self-edge guard for execution(* *(..)).
- Test for the value = "..." named-argument form.

YutenC:
Pushed `f6d828e` adding the five test groups requested above — no production-code changes, `tests/test_aop_resolver.py` only.

- **NetworkX impact engine**: `test_networkx_impact_radius_propagates_from_advice_to_target_not_the_reverse`, calling `store._get_impact_radius_networkx([aspect_path])` directly, mirroring the SQL-engine direction test right above it.
- **Idempotency**: `test_resolve_aop_advice_is_idempotent` runs `resolve_aop_advice` twice with no source change and asserts the second run's `stale_calls_removed` equals the first run's `calls_created`, with an unchanged edge set.
- **End-to-end `full_build` -> `incremental_update`**: `test_incremental_annotation_removal_removes_stale_aop_call`, following the shape of `test_spring_events.py::test_incremental_listener_change_removes_stale_event_call` — removes the target's `@AccessLimit` annotation and asserts the stale `aop_resolved` edge is gone and `result["aop_resolution"]` is populated.
- **Skip-branch coverage**: `test_cross_aspect_dotted_reference_is_safely_skipped`, `test_unresolvable_named_reference_is_safely_skipped`, `test_advice_with_no_string_argument_is_safely_skipped`, and `test_self_edge_guard_skips_advice_matching_itself` (an `execution(* *(..))` pointcut that would otherwise match the advice method's own signature).
- **`value = "..."` form**: `test_value_named_argument_form_is_extracted`.

Full suite: `uv run pytest tests/ --tb=short -q` -> 3014 passed, 9 skipped, 2 xpassed, 0 failed (same skip/xpass set as `54b7a7e`, +8 for the new tests). `ruff check` and `mypy --ignore-missing-imports --no-strict-optional` both clean.

Didn't disagree with any of the five — each maps to a real, currently-untested branch — so went ahead rather than posting first.

tirth8205:
`f6d828e` covers all five groups, and they are real guards rather than decoration.

Gates on your head merged into `staging` 9c13f4c: `3364 passed, 9 skipped, 2 xpassed`, ruff clean, mypy clean. `tests/test_aop_resolver.py` is 29 passing. All 16 CI checks green.

I re-ran the original blocking repro independently rather than relying on your test for it. `a/Service.java` and `b/Service.java` each declaring `run()`, an `@Aspect` with `@Before("execution(* a.Service.run(..))")`, then `full_build`: exactly one `aop_resolved` edge, `Audit.log -> a/Service.java::Service.run`, and `aop_resolution: {'aspects_indexed': 1, 'advice_resolved': 1, 'calls_created': 1, 'stale_calls_removed': 0}`. Package selectivity holds.

I also mutation-checked the direction test, since that was the claim I trusted least. Replacing the `'"aop_resolved"' in (r["extra"] or "")` condition in `_build_networkx_graph` with `False` fails `test_networkx_impact_radius_propagates_from_advice_to_target_not_the_reverse` and nothing else. It pins the branch it says it pins.

This is good. Queued for merge.

Removing `needs-tests`.

tirth8205:
This no longer merges into `staging`. The one conflict is `code_review_graph/graph.py`, and the branch is 226 commits behind.

- `code_review_graph/graph.py`: staging added the corrupt-database classifiers and a `crg_parent_dir` SQLite function registered outside the open guard. Keep that guard wrapped around exactly the statements it covers.

```bash
git fetch origin
git merge origin/staging
# resolve, then
uv run pytest tests/ -q
git push
```

Issue #913, opened 2026-08-23
A git timeout in changes.py is returned as 'no changes': same empty result, exit code 0


FranzS2025:
## Summary

When a `git` invocation in `changes.py` exceeds `CRG_GIT_TIMEOUT`, the timeout is caught by a broad handler, logged at warning level, and turned into an **empty result**. To every caller this is indistinguishable from "the diff is empty, nothing changed" - and the process still exits 0. A review that silently analysed nothing looks exactly like a review that found nothing to flag.

## Root cause

`changes.py`, the `git diff` helper:

    try:
        result = subprocess.run(
            ["git", "diff", "--unified=0", base, "--"],
            capture_output=True,
            stdin=subprocess.DEVNULL,
            text=True,
            encoding="utf-8",
            errors="replace",
            cwd=repo_root,
            timeout=_GIT_TIMEOUT,
        )
        if result.returncode != 0:
            logger.warning("git diff failed (rc=%d): %s", result.returncode, result.stderr[:200])
            return {}
    except (OSError, subprocess.SubprocessError) as exc:
        logger.warning("git diff error: %s", exc)
        return {}

`subprocess.TimeoutExpired` is a subclass of `subprocess.SubprocessError`, so a timeout lands in the second handler and returns `{}`. The same `timeout=_GIT_TIMEOUT` pattern appears at three call sites in the file (lines 71, 113, 258).

`{}` is also the legitimate return value for "no changes", so nothing downstream can tell the two apart. On a large repository, or a cold filesystem cache, or a network-mounted worktree, the default 30 seconds is reachable.

## Why the warning is not enough

In MCP use the logger output does not reach the caller - the tool result does. A client receives a well-formed response describing zero changed functions, zero affected flows and `risk_score` 0.00, with no field indicating that the underlying command never completed.

## Suggested direction

Distinguish the two cases in the returned value, not only in the log: catch `subprocess.TimeoutExpired` separately and either raise, or return a result carrying an explicit marker (a `status` of `"timeout"`, or the `confidence` field 2.3.8 already introduced for empty results elsewhere - "the git command timed out after Ns" is exactly the kind of one-sentence explanation that field was added for). Exiting non-zero on the CLI path would also be defensible: a review that could not read the diff has not reviewed anything.

PR #921, opened 2026-08-25
fix: treat empty integer env vars as unset, not ValueError

### LINKED ISSUE #912: An empty integer env var (CRG_GIT_TIMEOUT=) kills the tool at import with an uncaught ValueError
## Summary

Setting any of the integer-valued configuration variables to an **empty string** kills the tool at import time with an uncaught `ValueError`, instead of falling back to the default or reporting a configuration problem. An empty environment variable is a very common accident - an unset shell variable expanded into a wrapper script, a `.env` line with nothing after the `=`, a CI secret that resolved to nothing.

## Reproduction

    CRG_GIT_TIMEOUT="" code-review-graph status --repo .

Observed on 2.3.8 (Windows 11, Python 3.13):

      File ".../code_review_graph/incremental.py", line 672, in <module>
        _GIT_TIMEOUT = int(os.environ.get("CRG_GIT_TIMEOUT", "30"))  # seconds, configurable
    ValueError: invalid literal for int() with base 10: ''

Exit code 1. Every subcommand that imports `incremental` is affected, so the tool is unusable until the variable is unset - and the traceback names the variable only incidentally, in the source line that happens to be echoed.

## Why the default does not help

`os.environ.get(NAME, "30")` returns the default only when the variable is **absent**. A variable that is present but empty returns `""`, which then reaches `int()`.

## Scope

The same pattern appears at several module-level sites, so the same input kills the tool from more than one direction:

- `incremental.py:672` - `CRG_GIT_TIMEOUT`
- `changes.py:34` - `CRG_GIT_TIMEOUT`
- `changes.py:465` - `CRG_MAX_CHANGED_FUNCS`
- `constants.py:43` - `CRG_MAX_IMPACT_NODES`
- `incremental.py` - `CRG_DEPENDENT_HOPS`

Because these run at import time, there is no opportunity for a caller to catch or report the failure.

## Suggested direction

A single helper used by all of them, e.g.

    def _int_env(name: str, default: int) -> int:
        raw = os.environ.get(name, "").strip()
        if not raw:
            return default
        try:
            return int(raw)
        except ValueError:
            logger.warning("%s=%r is not an integer; using %d", name, raw, default)
            return default

That turns a fatal import error into a named, actionable warning, and treats "set but empty" the same as "unset", which is what a user setting `VAR=` almost always means.

yzxcj797:
## Summary

An empty string environment variable (`VAR=` from a shell wrapper, CI template, or `.env` file) hit `int()` at module import and killed every subcommand with an uncaught `ValueError` (#912). `os.environ.get(NAME, default)` only returns the default when the variable is **absent**, not when it is present but empty.

A single `_int_env()` helper in `constants.py` now:
- treats empty/whitespace as unset (returns default)
- parses valid integers normally
- logs a named warning with the variable name and falls back to the default for invalid values

All module-level `int(os.environ.get(...))` sites now use it:
- `constants.py`: `CRG_MAX_IMPACT_NODES`, `CRG_MAX_IMPACT_DEPTH`, `CRG_MAX_BFS_DEPTH`, `CRG_MAX_SEARCH_RESULTS`
- `changes.py`: `CRG_GIT_TIMEOUT`, `CRG_MAX_CHANGED_FUNCS`
- `incremental.py`: `CRG_PARSE_WORKERS`, `CRG_MODULE_SCAN_DEPTH`, `CRG_MODULE_SCAN_MAX_DIRS`, `CRG_GIT_TIMEOUT`

## Testing

New `tests/test_int_env.py` (7 tests):
- `test_unset_returns_default` / `test_empty_string_returns_default` / `test_whitespace_returns_default` — the helper's three "fall back" branches
- `test_valid_integer` — normal parsing
- `test_invalid_integer_warns_and_returns_default` — warning includes the variable name
- `test_empty_git_timeout_does_not_crash_import` / `test_empty_max_impact_nodes_does_not_crash_import` — subprocess verification that the original reproduction commands now succeed

```
pytest tests/test_int_env.py -q
→ 7 passed

pytest tests/test_int_env.py tests/test_incremental.py -q (2 symlink-privilege deselected)
→ 112 passed, 2 deselected

ruff check code_review_graph/constants.py code_review_graph/changes.py code_review_graph/incremental.py tests/test_int_env.py
→ All checks passed!
```

Fixes #912


tirth8205:
The helper is the right idea and it does fix the reported command — `CRG_GIT_TIMEOUT="" python -c "import code_review_graph.incremental"` crashes on the parent and imports cleanly here. Two things to finish before it goes in.

**The sweep stops less than half way, and #912's own scope list is not covered.** The PR body says "All module-level `int(os.environ.get(...))` sites now use it." Ten module-level sites still crash at import with an empty value on this branch:

```
incremental.py:1136  CRG_DEPENDENT_HOPS          ← named in #912's Scope section
incremental.py:1601  CRG_WATCH_PLAN_DEPTH
incremental.py:1602  CRG_MAX_WATCH_SCHEDULES
incremental.py:1607  CRG_WATCH_SPLIT_MIN_DIRS
incremental.py:243   CRG_NESTED_IGNORE_TTL       (float)
incremental.py:1608  CRG_WATCH_HEALTH_INTERVAL   (float)
daemon.py:106-108,544  CRG_RESTART_BACKOFF, _MAX, CRG_RESTART_HEALTHY_AFTER, CRG_WATCH_HEALTH_STALE (float)
```

`CRG_DEPENDENT_HOPS=""` reproduces the reporter's exact traceback, and it lives in the same file you edited in three other places. Three call-time sites also still raise, and one of them is user-facing: `CRG_TOOL_TIMEOUT=""` makes `detect_changes_tool` raise `ValueError` instead of returning a structured error (and leaks an un-awaited `asyncio.to_thread` coroutine on the way out). `CRG_MAX_TRANSITIVE_FRONTIER` and `CRG_LEIDEN_SEED` are the other two. Both `CRG_TOOL_TIMEOUT` and `CRG_MAX_TRANSITIVE_FRONTIER` are documented in the README env table.

**No clamping, so the crash moves rather than disappears.** Empty → default is right, and invalid → warning + default is a reasonable call. But out-of-range values pass through untouched, and two of them are actively harmful:

- `CRG_PARSE_WORKERS=0` (or `-4`) now imports fine and then kills `code-review-graph build` on any repo with 8+ files with an unhandled `ValueError: max_workers must be greater than 0`.
- `CRG_GIT_TIMEOUT=0` or `-1` makes `subprocess.run(timeout=…)` raise `TimeoutExpired` on every git call, which `changes.py` swallows into `{}` — a typo'd config silently produces exactly the "analysed nothing" failure that #913 is about.
- `CRG_MAX_IMPACT_NODES=999999999999999999999999` is accepted, which defeats the response bounds.

`constants.py:10` already has `_bounded_float_env(name, default, *, lower, upper)` with a range check, tested in `tests/test_graph.py:778`. Your `_int_env` lands 34 lines below it and reimplements the parse half without the bounds half, and `changes.py:244` has yet a third pattern for the same job. A bounded `_int_env(name, default, *, lower, upper)` would cover the finding above and let all three converge — and the float vars need the same treatment.

Minor: the tests fail on the parent as a *collection* error (`cannot import name '_int_env'`) rather than a behavioral failure, so the two subprocess tests that actually reproduce #912 never run; decoupling them from the helper import makes them genuinely protective. `import logging` is deferred inside `_int_env` with no circular-import reason. And the README env table should state the new "invalid → warning + default" contract.

Gates are green: 2992 passed, ruff and mypy clean, 13/13 checks.

tirth8205:
Changes required: an empty documented numeric setting still prevents the module from importing. Reproduce with `CRG_DEPENDENT_HOPS='' python -c 'import code_review_graph.incremental'`, which raises `ValueError` on this PR. Apply consistent bounded parsing across the remaining integer and float settings, including zero and out-of-range values.

tirth8205:

- Convert the four remaining module-level int() sites in code_review_graph/incremental.py to _int_env: line 1313 _MAX_DEPENDENT_HOPS (CRG_DEPENDENT_HOPS, explicitly named in issue #912), line 1865 _WATCH_PLAN_DEPTH (CRG_WATCH_PLAN_DEPTH), line 1866 _MAX_WATCH_SCHEDULES (CRG_MAX_WATCH_SCHEDULES), line...
- Handle the module-level float() sites with the same empty/invalid semantics (a _float_env helper or by extending _bounded_float_env to treat '' as unset): incremental.py:279 CRG_NESTED_IGNORE_TTL, incremental.py:1871 CRG_WATCH_HEALTH_INTERVAL, daemon.py:106-108 CRG_RESTART_BACKOFF / CRG_RESTART_BACK...
- Optionally convert the call-time int() sites (graph.py:665 CRG_MAX_TRANSITIVE_FRONTIER, main.py:725 CRG_TOOL_TIMEOUT, communities.py:413 and :650 CRG_LEIDEN_SEED) so the helper is the single parsing path the PR description claims it is.
- Correct the PR description: 'All module-level int(os.environ.get(...)) sites now use it' is false on the current head, and 'Fixes #912' is not accurate while CRG_DEPENDENT_HOPS (listed in the issue) still crashes.

tirth8205:
This no longer merges into `staging`. 3 files conflict and the branch is 226 commits behind:

- `code_review_graph/changes.py`
- `code_review_graph/constants.py`

```bash
git fetch origin
git merge origin/staging
# resolve, then
uv run pytest tests/ -q
git push
```

PR #922, opened 2026-08-25
fix: raise GitTimeoutError instead of returning empty on git/svn timeout

### LINKED ISSUE #913: A git timeout in changes.py is returned as 'no changes': same empty result, exit code 0

yzxcj797:
## Summary

A git or svn diff that exceeded `CRG_GIT_TIMEOUT` was caught by a broad `OSError`/`SubprocessError` handler and returned an empty dict — indistinguishable from a genuine no-change diff. The process still exited 0, so an MCP caller received a well-formed zero-change review with no indication the underlying command never completed (#913).

`subprocess.TimeoutExpired` is now caught **separately** at both the git diff and svn diff call sites and raises `GitTimeoutError`, a `RuntimeError` subclass carrying:

- `command` — which VCS command timed out (`"git diff"` or `"svn diff"`)
- `timeout` — the configured `CRG_GIT_TIMEOUT` value
- a message that says the diff was not read and the review analysed nothing

Non-timeout `OSError` and `SubprocessError` cases still return `{}` as before — the distinction is only for the timeout path where silence is most dangerous.

## Testing

New `tests/test_git_timeout.py` (6 tests):
- `GitTimeoutError` is a `RuntimeError`; its message names the command and timeout; attributes are accessible.
- `parse_git_diff_ranges` raises `GitTimeoutError` when `subprocess.run` raises `TimeoutExpired` (mocked).
- Non-timeout `SubprocessError` and `OSError` still return `{}` (no behavior change for those paths).

```
pytest tests/test_git_timeout.py -q
→ 6 passed

pytest tests/test_changes.py -q
→ 38 passed

ruff check code_review_graph/changes.py tests/test_git_timeout.py
→ All checks passed!
```

Fixes #913


tirth8205:
The premise is right — a timeout that reads as "no changes" is the worst possible failure for a review tool, which is why I filed #913. But this doesn't fix the path the issue describes, and it breaks one that works today. I'd like it reworked rather than merged.

**It doesn't fix the reported symptom.** #913's complaint is a well-formed response saying zero changed functions, zero flows, risk 0.00. On this branch, with a forced timeout:

```
detect_changes_func(base="HEAD~1", ...)
{"status": "ok", "summary": "No changed files detected.", "risk_score": 0.0, ...}

code-review-graph detect-changes --repo .
No changes detected.        exit code: 0
```

The reason is upstream of your change: `review.py:573` calls `get_changed_files()` *before* `parse_diff_ranges()`, and `incremental.py:875` is

```python
except (FileNotFoundError, subprocess.TimeoutExpired):
    return []
```

— a silent `[]`, no log line, same again at `:959` for `get_staged_and_unstaged`. With an empty file list `detect_changes_func` short-circuits at `review.py:577` to the "No changed files detected" response and your `GitTimeoutError` is never reached. The CLI short-circuits identically at `cli.py:2090`. The fix has to land in `incremental.py`.

**It converts a designed degradation into a crash.** `changes.py:466` deliberately falls back to whole-file node analysis when the diff ranges are empty but the changed-file list is known, so that case produces a correct, lower-granularity review with a real risk score — not a false "nothing changed". Turning it into an exception loses that. Concretely, on `get_minimal_context`, the tool CLAUDE.md tells every agent to call first:

```
parent:  {"status":"ok","summary":"5942 nodes, 49847 edges across 269 files. Risk: high (0.75). 39 test gaps.", ...}
this PR: GitTimeoutError: git diff timed out after 0s; the diff was not read so the review analysed nothing
```

`context.py:126` catches `(ImportError, OSError, ValueError, sqlite3.Error, subprocess.SubprocessError)`; `GitTimeoutError` is a `RuntimeError`, so it escapes, and neither `get_minimal_context` nor `main.py:214` has a handler. The CLI is the same story — `cli.py:2097` is unguarded, so the realistic case where `git diff --name-status` succeeds and the expensive `--unified=0` times out prints a raw traceback at exit 1, where the parent printed a full report at exit 0. That also hard-fails the shipped Action step at `action.yml:93`. Non-zero exit is what #913 asked for; a traceback is not — `cli.py:1725` shows the convention (`SystemExit(1)` with a clean stderr message).

What I'd take instead:

1. Raise from `get_changed_files` / `get_staged_and_unstaged` on timeout, since that is where "empty" becomes indistinguishable from "no changes".
2. Catch it at every boundary that has a defined failure shape: `detect_changes_func` already has a blanket handler and returns a structured error (verified, no regression there), but `get_minimal_context` needs `GitTimeoutError` in its tuple, and the CLI needs `SystemExit(1)` with a one-line message.
3. Keep the empty-ranges-with-known-files path as graceful degradation — do not raise there.
4. Say what to do in the message. Name `CRG_GIT_TIMEOUT` as the knob, the way `main.py:729` already does for `CRG_TOOL_TIMEOUT`.

Two smaller things for the rework: `parse_git_diff_ranges` and `parse_svn_diff_ranges` still document "Returns an empty dict on error" while now raising, and both are in `tools/__init__.__all__`, so that is a silent breaking change for library callers — `GitTimeoutError` itself is exported nowhere, so a caller cannot even catch it without reaching into `code_review_graph.changes`. And `compute_file_churn` (`changes.py:287`) still swallows the same timeout into `{}`, which is the third site #913 names.

One warning about the interaction with #921: that PR accepts `CRG_GIT_TIMEOUT=0` unclamped, which forces `TimeoutExpired` on every git call. With both merged as they stand, one typo'd env var makes `get_minimal_context_tool` raise on every single invocation.

Gates are green on the branch (2991 passed, ruff and mypy clean, 13/13) — all six new tests patch `subprocess.run` and assert on `parse_git_diff_ranges` in isolation, which is exactly why CI cannot see either regression above. A test through `detect_changes_func` and one through the CLI would have caught both.

tirth8205:
Changes required: the false-clean response remains at changed-file discovery, before the modified range parser is reached. On a built Git graph, make `code_review_graph.incremental.subprocess.run` raise `subprocess.TimeoutExpired` and call `code_review_graph.tools.review.detect_changes_func(base='HEAD~1', repo_root=...)`; this PR still returns `status='ok'`, no changes and zero risk. Report discovery failures at the public boundary while preserving conservative whole-file context when filenames are known but line ranges are unavailable.

tirth8205:

- Test for parse_svn_diff_ranges raising GitTimeoutError('svn diff', _GIT_TIMEOUT) on subprocess.TimeoutExpired (the second code path changed by the PR is untested).
- End-to-end test for the MCP tool: detect_changes_func with get_changed_files patched to a non-empty list and parse_diff_ranges/subprocess.run raising TimeoutExpired must return status='error' whose error mentions 'timed out', not status='ok' with risk_score 0 (extend the pattern in tests/test_change...
- Decide and test CLI behavior for `detect-changes` on timeout: currently an uncaught traceback with exit 1. Either catch GitTimeoutError in cli.py (print 'Error: ...' to stderr, sys.exit(1), like the `watch` branch at cli.py:2015) and test the exit code and message, or explicitly document the traceba...
- Decide and test `update --brief` on timeout: the update itself succeeds and then the process dies with a traceback (regression from exit 0). Likely should catch GitTimeoutError around the brief summary and warn rather than fail the update.
- Decide and test get_minimal_context on timeout: it now raises GitTimeoutError out of the tool instead of degrading to risk='unknown'; either add RuntimeError/GitTimeoutError to the except tuple at tools/context.py:127 (with a test asserting risk='unknown' and the response still returns), or return a...

tirth8205:
This no longer merges into `staging`. The one conflict is `code_review_graph/changes.py`, and the branch is 226 commits behind.

```bash
git fetch origin
git merge origin/staging
# resolve, then
uv run pytest tests/ -q
git push
```

PR #928, opened 2026-08-27
fix: read source files as UTF-8 regardless of system locale

### LINKED ISSUE #926: review snippets and eval readers decode with the system locale, so non-ASCII source is mojibake on Windows
Surfaced while reviewing #918. Verified by reading the source; not reproduced on a Windows host.

## What happens

Several production reads decode files with the platform's default encoding instead of UTF-8. On Windows that is cp1252, so any source file containing non-ASCII bytes is decoded incorrectly:

- `code_review_graph/tools/review.py:264` and `:625` — the source snippets handed to the model
- `code_review_graph/tools/flows_tools.py:187`
- `code_review_graph/context_savings.py:178` — also skews the token-savings figures
- `code_review_graph/token_benchmark.py:41`

These pass `errors="replace"`, so they do not crash; they silently produce mojibake. The review snippets are the ones that matter: a reviewer model reading corrupted identifiers or strings is being handed bad evidence.

Worse, these call `open()` with neither `encoding=` nor `errors=`, so they can raise `UnicodeDecodeError` outright:

- `code_review_graph/eval/runner.py:67, :76, :174`
- `code_review_graph/eval/reporter.py:78`

## Note

The codebase already gets this right in most places — `incremental.py:353`, `daemon.py:575`, `parser.py:83` and `:599` all pass `encoding="utf-8"`. This is drift, not a design choice.

## Suggested fix

Pass `encoding="utf-8"` (keeping `errors="replace"` where a partial read is better than a failure) at every listed site, and add a test that reads a fixture containing non-ASCII source through the review path and asserts the round-trip.

yzxcj797:
## Summary

Review snippets, `flows_tools` source reads, and eval runner/reporter file opens used the **system locale** for decoding. On Windows with `cp936` (Chinese) or other non-UTF-8 locales, non-ASCII source code became mojibake in review output and eval data (#926).

All `read_text` and `open` calls in these paths now specify `encoding="utf-8"` explicitly:

- `tools/review.py` — 2 `read_text` calls (snippet source)
- `tools/flows_tools.py` — 1 `read_text` call (source lines)
- `eval/runner.py` — read and write opens
- `eval/reporter.py` — write open

## Testing

New `tests/test_utf8_encoding.py` (4 tests) — source-level guards verifying each file contains `encoding="utf-8"` in its read/open calls:

```
pytest tests/test_utf8_encoding.py -q
→ 4 passed

ruff check <all changed files>
→ All checks passed!
```

Fixes #926


tirth8205:
Integrated on `integration/token-efficiency-hardening` as `cd2e174`. Explicit UTF-8 decoding preserves non-ASCII review and flow snippets and evaluation data. A behavioural regression simulating a cp1252 default reproduces corruption of `你好 café` on main and verifies exact text after the change; this does not claim a native Windows run.

tirth8205:
Merged into `staging` through the integration PR #988 (merge commit 3030ca8). Your change is carried by commit cd2e174, which keeps you as the author (yzxcj797). It will be promoted staging → testing → main and released from there. Thank you for the contribution! Closing this PR since its content is now on staging.

PR #933, opened 2026-08-29
fix(ruby): emit Class nodes for compact namespace declarations (`class Foo::Bar`)

### LINKED ISSUE #932: Ruby: `class Foo::Bar` emits no Class node — compact namespaces drop the class and orphan its methods
## Summary

For Ruby, a class declared with a compact namespace — `class Foo::Bar` — produces **no `Class` node at all**. The file and its methods are still indexed, but the methods end up with `parent_name = NULL` and no `CONTAINS` edge, so the class is invisible to `query_graph`, `search`, `dead_code` and impact analysis.

The nested form (`module Foo` + `class Bar`) works correctly.

Version: **2.3.8** (current latest on PyPI), installed via `uvx code-review-graph serve`.

## Minimal reproduction

Two files that differ only in declaration style:

```ruby
# app/compact.rb
class Widgets::CompactStyle
  def render
    :ok
  end
end
```

```ruby
# app/nested.rb
module Widgets
  class NestedStyle
    def render
      :ok
    end
  end
end
```

```console
$ uvx code-review-graph build --repo . -q
$ sqlite3 .code-review-graph/graph.db \
    "SELECT kind, name, parent_name FROM nodes ORDER BY file_path, line_start;"
```

**Actual:**

```
File      .../app/compact.rb    parent=None
Function  render                parent=None      <-- orphaned, no Class node emitted
File      .../app/nested.rb     parent=None
Class     Widgets               parent=None
Class     NestedStyle           parent=Widgets
Function  render                parent=NestedStyle
```

**Expected:** `app/compact.rb` should emit a `Class` node (`CompactStyle`, or `Widgets::CompactStyle`) with `render` as its child.

## Root cause

`CLASS_NODE_TYPES["ruby"] = ["class", "module"]` (parser.py:715) is correct — the `class` node *is* visited. The gap is in name extraction.

tree-sitter-ruby types the `name` field differently for the two styles:

```
class Widgets::CompactStyle     ->  (class name: (scope_resolution (constant) :: (constant)))
module Widgets / class Nested   ->  (class name: (constant))
```

`_get_name` handles `constant` but not `scope_resolution`, so it returns `None` and the node is dropped.

There is existing precedent for exactly this shape elsewhere in `_get_name` — the C/C++/ObjC branch peels `qualified_identifier` down to the rightmost leaf (the `_leaf_name` helper). Ruby's `scope_resolution` needs the same treatment: take the rightmost `constant` as the class name, and optionally the left-hand side as the namespace/parent.

Related previously-fixed issues of the same family: #463 (C++ scoped `Class::method()`), #567 (PHP/Rust `Class::method` scoped calls), #741 (Ruby CALLS edges).

## Impact

This is the dominant declaration style in Rails codebases. Measured on a real Rails 8.1 app (`app/**/*.rb`), counting files with at least one `Class` node in the graph:

| Declaration style | files | files with a `Class` node |
|---|---:|---:|
| `class Foo::Bar` (compact) | 64 | **2** |
| `module Foo` + `class Bar` (nested) | 19 | 19 |

62 of 64 files are missing class-level structure entirely. In that codebase every Phlex view component uses the compact form, so the whole component layer has no class nodes.

Downstream effects:

- `query_graph callers_of/inheritors_of <ClassName>` → `not_found`
- `semantic_search_nodes` → 0 results for a class that plainly exists
- Methods are orphaned (`parent_name = NULL`), so `children_of` on the class is impossible and community/flow detection loses the grouping
- `refactor_tool mode=dead_code` cannot reason about these classes at all

## A note on the failure mode

`query_graph` is admirably careful here — it returns

```
"confidence": "target not indexed: no node matching 'WatchlistManagerComponent',
               so this 0 is not evidence that none exist"
```

which is exactly right. But `semantic_search_nodes` returns a plain `result_count: 0` with no such caveat, and a caller that only checks `result_count` can easily read "not indexed" as "no callers" — and delete a live class. Surfacing the same not-indexed caveat from `semantic_search_nodes` would make this much safer even before the parser fix lands.

idarfan:
# Pull Request

## Linked issue

Closes #932

## What & why

`class Foo::Bar` produced no `Class` node at all.

tree-sitter-ruby types the `name` field of a compact declaration as `scope_resolution` rather than `constant`, so `_get_name` fell through to the generic child-type loop, returned `None`, and `_extract_classes` bailed at `if not name: return False`. The class vanished and its methods were left with `parent_name = None`.

The nested form (`module Foo` + `class Bar`) was unaffected, which made the gap easy to miss — but the compact form is the dominant style in Rails codebases. On one Rails 8.1 app, **62 of 64** compact-form files had no `Class` node (total `Class` nodes in the graph: 231 → 415 after this change), so the entire Phlex component layer was invisible to `callers_of`, `inheritors_of` and `semantic_search_nodes`.

The fix takes the rightmost `constant` as the name, mirroring the existing C/C++/ObjC `qualified_identifier` handling in the same function. This matches the nested form — `module Foo` + `class Bar` already yields Class `"Bar"` — so the same class gets the same node name regardless of declaration style. The namespace belongs in `parent_name`, not in `name`.

### Known gap this does not close

Compact-form classes still get `parent_name = None`, whereas the nested form records the enclosing module. `_get_name` cannot fix that: `enclosing_class` comes from the tree walk and there is no enclosing `module` node in the compact form. Deriving it from the left-hand side of the `scope_resolution` is a natural follow-up; I kept this PR to the change that unblocks node emission.

## How it was tested

```bash
# 2987 passed, 9 skipped, 2 xpassed, 1 warning in 440.22s (0:07:20)

# All checks passed!

```

The two new tests fail without the parser change, confirming they guard the regression:

```
tests/test_multilang.py::TestRubyParsing::test_compact_namespace_class_is_indexed
  AssertionError: assert 'CompactStyle' in {}
tests/test_multilang.py::TestRubyParsing::test_compact_namespace_module_is_indexed
  AssertionError: assert 'Helpers' in set()
```

Before/after on the minimal repro from #932:

```
                     before                    after
compact.rb  File                       File
            Function render (parent=None)  Class    CompactStyle
                                           Function render (parent=CompactStyle)
```

## Checklist

- [x] Docs updated where behavior changed (docstrings on both new tests explain the grammar detail; inline comment at the fix site)


tirth8205:
Changes required: compact Ruby namespace syntax collapses distinct classes into one identity. Save `class A::Same; def a; end; end` and `class B::Same; def b; end; end` in one Ruby file and call `CodeParser(repo_root).parse_file(path)`; both classes become `file::Same`, with both methods attached to that merged class. Preserve the complete namespace and method ownership before integrating.

tirth8205:

- Test for 3+ segment compact names: class A::B::C -> Class 'C' (and method parent_name 'C')
- Test for compact form with superclass: class Foo::Bar < Base -> Class 'Bar' emitted (Rails-dominant form; the class node has an extra superclass child)
- Test for top-level scope form: class ::Foo -> Class 'Foo'
- Test that a compact declaration inside a module keeps the enclosing module as parent_name (module A; class B::C -> parent 'A')
- Test for the fallthrough branch: scope_resolution with no constant child yields no crash (may emit no node)

idarfan:
Agreed on the five tests — written, and all pass. Before I finalise them I need one thing settled, because two of your requirements point in opposite directions.

**The Sept 12 identity collapse is still present.** No code has changed since `0b61a97` (Aug 29), so flagging in case it was assumed resolved. Repro — two compact classes with the same leaf, one file, `CodeParser().parse_file(path)`:

```ruby
class Alpha::Same
  def a; end
end

class Beta::Same
  def b; end
end
```

```
Class    name='Same'  parent=None    qualified='file.rb::Same'
Function name='a'     parent='Same'  qualified='file.rb::Same.a'
Class    name='Same'  parent=None    qualified='file.rb::Same'   <- same identity
Function name='b'     parent='Same'  qualified='file.rb::Same.b'
```

The nested form does not have this problem:

| Declaration | `parent_name` | identity |
|---|---|---|
| `module Alpha; class Same` | `'Alpha'` | `file.rb::Alpha.Same` |
| `module Beta; class Same` | `'Beta'` | `file.rb::Beta.Same` |
| `class Alpha::Same` | `None` | `file.rb::Same` |
| `class Beta::Same` | `None` | `file.rb::Same` |

My patch's own comment claims the result "matches the nested form". The *name* matches; the *identity* does not. That's my error and I'd rather fix it than document around it.

**Where the two requirements collide.** Your test 4 asks for `module A; class B::C` -> parent `'A'`. That already holds today, so the test passes as written — but it codifies discarding the `B` segment. The nested equivalent, `module Alpha; module Beta; class Gamma`, yields parent `'Beta'` (verified: parser.py:10299-10301 sets `recursive_class = name`, immediate scope only). So under test 4 as worded, `module A; class B::Same` and `module A; class C::Same` both become `file.rb::A.Same` — the Sept 12 collapse, reinstated.

**Two ways forward — I'd like your call:**

**(a) Match the nested form.** Derive `class_parent` for compact declarations from the scope segments, so `class Alpha::Beta::Gamma` gets parent `'Beta'`. Test 4 then expects `'B'`, not `'A'`, and test 1 gains an assertion that two same-leaf classes under different scopes keep distinct identities. Deliberately mirrors the nested form's existing shallowness (only the immediate scope is retained) rather than quietly changing it.

**(b) Keep test 4 as written.** Then the Sept 12 collapse is explicitly out of scope here; happy to open a follow-up issue for it if you confirm.

I'd recommend (a) — it's a few lines in the same function, reuses the existing `parent_name` mechanism with no need for `identity_name`, and it delivers what the Sept 12 comment asked for. But it's your call and I'll write whichever you pick.

One thing (a) will *not* fix, and I don't propose to touch it here: method identities are shallow in both forms. `module Alpha; class Same; def run` and `module Beta; class Same; def run` both resolve to `file.rb::Same.run` today. Pre-existing and orthogonal — worth a separate issue if you agree.

**Current state of the branch:** 6 new tests covering your five items (the fallthrough branch gets two — a direct `_get_name` unit test that deterministically hits the no-`constant` path, plus a malformed-source smoke test). Full suite `2993 passed, 9 skipped, 2 xpassed`; `ruff check` clean. Reverting `parser.py` to `origin/staging` fails 6 of them, confirming they are real regression guards.

tirth8205:
You are right, on every claim. I reproduced all of it against current `staging` (9c13f4c) with `CodeParser(repo_root).parse_file(path)`.

The collapse, exactly as you reported it:

```
class Alpha::Same -> Class 'Same' parent=None -> file.rb::Same
class Beta::Same  -> Class 'Same' parent=None -> file.rb::Same
```

The nested form does keep them apart (`file.rb::Alpha.Same`, `file.rb::Beta.Same`), and the nested form is immediate-scope only exactly as you said: `module Alpha; module Beta; class Gamma` gives Gamma `parent_name` `'Beta'`, not `'Alpha.Beta'`.

And the collision is real. Test 4 as I worded it:

```ruby
module A
  class B::Same
    def b; end
  end
  class C::Same
    def c; end
  end
end
```

gives both classes `parent_name` `'A'` and identity `file.rb::A.Same`, twice. So test 4 as written does reinstate the Sept 12 collapse. That was my error, not yours.

## (a) wins. Write (a).

Test 4 was written from what the parser does today, not from a decision about what it should do, and what it does today is the defect. The rule this PR exists to establish is that the compact form gets the same identity as the nested form, and (a) is what that rule produces.

Test 4 becomes: `module A; class B::C` expects `parent_name` `'B'`.

That is not discarding `A`. Check the nested equivalent: `module A; module B; class C` also gives C `parent_name` `'B'` and drops `A`. Option (a) makes the compact form agree with the nested form exactly, including agreeing about the shallowness. Keeping test 4 as written is the option that makes the two forms disagree.

## The change

I prototyped (a) to be sure it holds up. In `_extract_classes`, immediately after `class_parent = enclosing_class`:

```python
if language == "ruby" and child.type in ("class", "module"):
    _name_child = child.child_by_field_name("name")
    if _name_child is not None and _name_child.type == "scope_resolution":
        _scope = _name_child.child_by_field_name("scope")
        if _scope is not None:
            _immediate = (
                _scope if _scope.type == "constant"
                else _scope.child_by_field_name("name")
            )
            if _immediate is not None:
                class_parent = _immediate.text.decode("utf-8", errors="replace")
```

tree-sitter-ruby hands you the immediate scope for free. `Alpha::Beta::Gamma` has `scope` = `scope_resolution(Alpha::Beta)` whose `name` field is `Beta`. `Bar::Baz` has `scope` = `constant(Bar)`. `::Foo` has no `scope` field at all, so it falls straight through and your test 3 keeps passing unchanged. Leave `recursive_class = name` alone so method `parent_name` stays the leaf.

With that in place:

| Declaration | `parent_name` | identity |
|---|---|---|
| `class Alpha::Same` | `'Alpha'` | `file.rb::Alpha.Same` |
| `class Beta::Same` | `'Beta'` | `file.rb::Beta.Same` |
| `module Alpha; class Same` | `'Alpha'` | `file.rb::Alpha.Same` |
| `module A; class B::C` | `'B'` | `file.rb::B.C` |
| `module A; module B; class C` | `'B'` | `file.rb::B.C` |
| `class ::Foo` | `None` | `file.rb::Foo` |
| `class Bar::Baz < Base` | `'Bar'` | `file.rb::Bar.Baz` |

Full suite with the prototype applied: `3337 passed, 9 skipped, 2 xpassed`. Identical to without it, so nothing in the tree leans on the old behaviour.

## What to do

1. Apply the above.
2. Test 4 expects `'B'`, not `'A'`. Add to it that `module A; class B::Same` and `module A; class C::Same` keep distinct identities.
3. Test 1 gains the same-leaf assertion: `class Alpha::Same` and `class Beta::Same` give two Class nodes with different qualified names.
4. Tests 2, 3 and 5 are unaffected.
5. Push the tests. Your branch head is still `0b61a97` and carries two tests, not six. Nothing you described in that comment is on the PR, so I could not run any of it. That is the only reason this is not already merged.

## On the method identities

You are right there too, it is real, and it is not Ruby-specific. Python does the same thing with nested classes: two `class Same` under different outer classes in one file both resolve to `m.py::Same.run`. Correctly out of scope for this PR. Filed as #999 with both reproductions and your credit on it.

Gates for the record, your head merged into `staging` 9c13f4c: `3337 passed, 9 skipped, 2 xpassed`, ruff clean, mypy clean. The plumbing is fine. This is only about the `parent_name` decision and the missing push.

Keeping `needs-tests` until the tests land, adding `changes-requested` for the `parent_name` change.

idarfan:
Pushed. Head is now `250a77e`; the branch carries three commits (`0b61a97` fix, `a60a076` tests, `250a77e` the `parent_name` change).

Applied (a) as you wrote it — the `scope` field lookup in `_extract_classes` right after `class_parent = enclosing_class`, with `recursive_class = name` left alone so method `parent_name` stays the leaf. Your seven-row table reproduces exactly on my side.

Tests, now eight in `TestRubyParsing`:

- Test 4 expects `'B'`. I also renamed it `test_compact_namespace_inside_module_uses_immediate_scope` — the old name (`..._keeps_enclosing_parent`) and its docstring asserted the opposite of what it now checks.
- New `test_compact_namespaces_inside_one_module_stay_distinct`: `module Alpha` containing `Beta::Same` and `Gamma::Same` keeps two Class nodes with distinct qualified names.
- New `test_compact_namespace_same_leaf_keeps_distinct_identities`: the same at top level, `class Alpha::Same` vs `class Beta::Same`.
- Tests 1, 2, 3 and 5 unchanged. `class ::Foo` falls through as you said.

Local gates: full suite `2995 passed, 9 skipped, 2 xpassed`; `ruff check code_review_graph/` clean; `mypy code_review_graph/ --ignore-missing-imports --no-strict-optional` clean. My base is older than `9c13f4c`, so the count is lower than yours — your merge run is the authoritative one.

Regression check: reverting `parser.py` to `origin/staging` fails 8 of the 8 compact-namespace tests; reverting to `0b61a97` (the fix without the `parent_name` change) fails exactly 3 — the two new identity tests and test 4 — so the `parent_name` change is guarded on its own and not just riding on the original fix.

Thanks for filing #999 and for the credit.

tirth8205:
Ordinary classes nested inside a compact module lose their method edge.

```ruby
# lib/adapters.rb
module Billing::Adapters
  class Stripe
    def charge(amount); amount * 100; end
  end
end
```

On staging that class is keyed `lib/adapters.rb::Stripe`, the CONTAINS source matches, and `query children_of "lib/adapters.rb::Stripe"` returns `charge`. On your head merged into staging the class node becomes `lib/adapters.rb::Adapters.Stripe` while the edge stays:

```
CONTAINS lib/adapters.rb::Stripe -> lib/adapters.rb::Stripe.charge
```

That source node no longer exists, so `children_of` returns 0 under either name and the method is unreachable from its class. `uncertainty.py` only checks that the queried node is indexed, and it now is, so the empty result comes back asserted as real: `"'...::Adapters.Stripe' is indexed and the graph is current, so this 0 is a real absence"`.

The second case drops a node outright:

```ruby
# app/models/user.rb
class Admin::User
  def display_name; "admin"; end
end

class User
  def display_name; "user"; end
end
```

Both methods now carry `parent_name` `User`, so both qualify to `app/models/user.rb::User.display_name`. `qualified_name` is `TEXT NOT NULL UNIQUE` at graph.py:95, so `Admin::User#display_name` at line 2 vanishes and the surviving row absorbs both CONTAINS edges, attributing the compact class's body to the unrelated plain `User`. Staging keeps both, as `::display_name` and `::User.display_name`.

The cause is one line. parser.py:11343 recurses with `recursive_class = name`, the bare leaf, while the class node is keyed `_qualify(name, file_path, class_parent)`. Any class with a non-null `parent_name` ends up with a container key its own methods never use. That mismatch is older than this PR and the nested form has it too. What changed is the population: option (a) hands `parent_name` to classes that did not have one before, including ones whose edges were correct on staging. On the two inputs above staging is right and this branch is not, so it has to be dealt with here rather than in #999.

The suite cannot see either case. All nine new parse calls are written `nodes, _ = self.parser.parse_file(...)`, and `git diff origin/staging...250a77e -- tests/test_multilang.py | grep -c CONTAINS` gives 0. No new test puts an ordinary class inside a compact module. 4109 passing is real, but none of it looks at the structure.

Smaller one: `class klass::Inner` and `class self::Foo` now emit file-scope Class nodes. The scope node is an `identifier` or `self`, so `class_parent` is left alone, but `_get_name` has already returned the rightmost constant. Staging emitted nothing there.

The core fix does work, and I checked by building rather than reading. `class Admin::UsersController < ApplicationController` produces no Class node at all on staging and orphans `index` at file scope. On your branch it becomes `::Admin.UsersController` with its methods attached. The seven-row table reproduces row for row. Reverting parser.py to staging fails 8 of the new tests, and reverting only the `parent_name` commit fails exactly the 3 you named. Wall time is about +13% for +20% nodes on a 400-file synthetic corpus, and `tests/fixtures/sample.rb` parses byte-identically.

To make it mergeable:

1. Make the method container agree with the class node identity, so a class keyed `file::Parent.Class` gets `file::Parent.Class.method`. Passing the qualified identity down instead of `name` at parser.py:11343 is the smallest version, but that line is shared by every language, so scope it to Ruby or check the others.
2. Assert on edges in the new tests, including at least one ordinary class inside a compact module, checking the CONTAINS source node actually exists.
3. Skip the node when the scope is not a constant.
4. Note what happens to existing graphs. `module Foo::Bar; class Baz` re-keys `Baz` from `file.rb::Baz` to `file.rb::Bar.Baz`, and migrations.py is untouched, so `code-review-graph update` will leave both shapes in one database.

idarfan:
All three reproduce on my side, and all three are fixed locally. But requirement 1 turns out to be incompatible with the shallowness rule we agreed in (a), so I need your call before I finalise.

**The three reports, confirmed**

`module Billing::Adapters; class Stripe` gave class node `adapters.rb::Adapters.Stripe` against a CONTAINS source of `adapters.rb::Stripe` — dangling, exactly as you said. `Admin::User` and `User` in one file both produced `user.rb::User.display_name`. `class klass::Inner` and `class self::Foo` both emitted file-scope Class nodes.

Your diagnosis of the cause is right. Confirming the mechanism at my line numbers: the class node is keyed `_qualify(name, file_path, class_parent)` (parser.py:10293) while the method's CONTAINS source is `_qualify(container_scope, file_path, None)` with `container_scope = enclosing_class` (parser.py:10581, 10424). Any class with a non-null `parent_name` points its methods at a container that was never emitted.

You are also right that the tests could not see it — every new parse call discards `edges`. That is on me.

**Item 3 is done.** tree-sitter types the scope field distinctly, so the guard is exact:

| declaration | scope type | emit |
|---|---|---|
| `class Alpha::Beta::Gamma` | `scope_resolution` | yes |
| `class Bar::Baz` | `constant` | yes |
| `class ::Foo` | no scope field | yes |
| `class klass::Inner` | `identifier` | no |
| `class self::Outer` | `self` | no |

Restricting `_get_name` to a constant scope makes both weird forms fall through and emit nothing, matching staging.

**Where item 1 breaks (a)**

Container agreement requires descending with the class's own key scope, `class_parent + "." + name`. The next level down then receives that as its `class_parent`, so it is no longer a single segment. Depth accumulates by construction — shallowness only survives at depth 1. The two requirements cannot both hold.

Applying item 1 literally, keeping (a)'s immediate-scope rule for compact names:

| declaration | identity |
|---|---|
| `module A; class B::C` | `file.rb::B.C` |
| `module A; module B; class C` | `file.rb::A.B.C` |

The compact and nested forms disagree again at depth 2 — the thing this PR exists to prevent. Row 5 of the seven-row table no longer holds either.

The alternative is to let compact names accumulate too, so `class A::B::C` takes `parent_name` `A.B`:

| declaration | identity |
|---|---|
| `module A; class B::C` | `file.rb::A.B.C` |
| `module A; module B; class C` | `file.rb::A.B.C` |

Both forms agree at every depth, and this is what your Sept 12 comment asked for — the complete namespace preserved. Rows 1, 2, 3, 6 and 7 are unchanged under either option; only rows 4 and 5 move.

I'd recommend the second. The first is only literal compliance and hands the inconsistency straight back to you.

**Test fallout, either way**

`TestRubyParsing::test_finds_calls` asserts `t.endswith("sample.rb::UserRepository.save")`. `sample.rb` wraps both classes in `module Auth`, so that identity becomes `sample.rb::Auth.UserRepository.save`. That is an existing test of yours, it is a direct consequence of item 1, and it happens under both options — flagging rather than quietly editing it. The other failures are my own tests asserting a bare-leaf method `parent_name`, plus two more under the second option. No non-Ruby test moves under either.

**Item 4, existing graphs**

Re-keying is wider than `module Foo::Bar; class Baz`. Item 1 moves every Ruby method whose class has a parent, compact or nested, so `Auth::User#to_s` goes from `sample.rb::User.to_s` to `sample.rb::Auth.User.to_s` on a plain nested file that this PR otherwise does not touch. `migrations.py` is untouched, so `code-review-graph update` leaves both shapes in one database and `children_of` answers differently depending on which shape a file was last indexed under. That wants either a schema-version bump forcing a rebuild of Ruby files, or an explicit note in the release. Tell me which you prefer and I'll write it.

Nothing is pushed yet — this is a design question, and I'd rather not push a branch whose expectations I'm about to change again. Say the word and I'll push either prototype to a scratch branch so you can build against it.

tirth8205:
Take the second option. `module A; class B::C` and `module A; module B; class C` should both be `file.rb::A.B.C`.

You are right that the two requirements cannot both hold, and right that the first option hands the inconsistency back to me. The reason to prefer the second is not only consistency though. Ruby resolves the `B` in `class B::C` by lexical lookup, and the enclosing module is the first place it looks, so `A::B::C` is what that declaration usually means. The graph is a static approximation and cannot know for certain, but preferring the enclosing scope is the better guess and it is the one that makes the two spellings agree. Where it guesses wrong, `B` was a top-level constant shadowed by nothing, which is the rarer shape.

So drop the shallowness rule from (a). It was my rule and it does not survive contact with depth 2. Compact names accumulate like nested ones.

On the existing test, change it. `sample.rb::UserRepository.save` becoming `sample.rb::Auth.UserRepository.save` is the change working, not fallout to route around. You were right to flag it rather than edit it quietly.

On item 4, do the schema bump rather than the release note. A note does not help someone who already has a database, and you are re-keying every Ruby method whose class has a parent, not just the compact ones. `code-review-graph update` leaving two shapes in one database is the failure that costs people real answers, and it is silent.

The version is 13 today and it is coupled in four places that have to move together, or CI fails on the schema-sync job:

```
code_review_graph/migrations.py      a new _migrate_v14 entry in MIGRATIONS
code-review-graph-vscode/src/backend/sqlite.ts   SUPPORTED_SCHEMA_VERSION = 14
action.yml                           the schema13 cache key becomes schema14
docs/GITHUB_ACTION.md                the same key where it is documented
```

The migration itself only needs to force Ruby files to be re-parsed, not rewrite identities in place. Deleting the Ruby rows and letting the next build repopulate them is simpler and safer than trying to transform a name you no longer have the parse for.

Item 3 is right as you have it, and the scope-type table is the correct way to decide it.

No need to push a scratch branch. Push the real thing when it is ready.

idarfan:
Done and pushed. Head is `7fcaa97`.

You had merged `staging` into the branch yourself while I was working, so my push was rejected. I merged your `4642d51` into mine rather than force-pushing; it was clean, and I re-ran everything after it.

**Option two is in.** Compact names stack with the enclosing scope, so `module A; class B::C` and `module A; module B; class C` both key to `file.rb::A.B.C`. The shallowness rule is gone. Your lexical-lookup argument is in the commit message and in the test docstring, since it is the reason the guess is the right one rather than just the consistent one.

All three reports are fixed. `module Billing::Adapters; class Stripe` now keys the class `adapters.rb::Billing.Adapters.Stripe` and its method's CONTAINS source names that same node. `Admin::User` and `User` in one file qualify to `user.rb::Admin.User.display_name` and `user.rb::User.display_name`. `class klass::Inner` and `class self::Foo` emit nothing, on the scope-type table from my last comment.

`test_finds_calls` now expects `sample.rb::Auth.UserRepository.save`, as you said.

**Tests assert on edges now.** Two new ones: an ordinary class inside a compact module, checking the CONTAINS source is the class node's own identity; and a sweep over a file mixing all four shapes, asserting no CONTAINS source names a node the parse did not emit, and that no two Function nodes share a qualified name. Reverting `parser.py` to `250a77e` fails both.

**I had to merge `staging` first.** My base was 224 commits behind and still on schema 9 — `sqlite.ts` said 9, the cache key said `schema9`. There was no way to write `_migrate_v14` on that tree.

**On item 4, I did not do what you asked, and I want to be explicit about it.** You said to delete the Ruby rows. That fails your own blocking gate:

```
tests/test_upgrade_path.py::test_opening_loses_no_nodes_or_edges
"Opening only migrates; it re-parses nothing, so the counts must be equal."
```

Measured, deleting rows versus clearing `file_hash`, `CRG_UPGRADE_TEST=1 pytest -m upgrade`:

| `_migrate_v14` | result |
|---|---|
| delete Ruby rows | 3 failed — `test_opening_loses_no_nodes_or_edges` on 2.3.6, 2.3.7 and 2.3.8 |
| clear `file_hash` | 40 passed, 2 skipped, 6 xfailed |

So v14 is a single UPDATE setting `file_hash = ''` on Ruby File nodes. `get_file_hashes` reads that column, an empty hash cannot match the file on disk, and `store_file_nodes_edges` replaces a file's data atomically — so the next `update` re-parses every Ruby file and re-keys it through the normal path. Your goal, one step later, at the point where re-parsing actually happens and where the upgrade path already tolerates churn. Node and edge counts are identical across open, with nothing lost, and non-Ruby files are untouched.

The six xfails are pre-existing and unrelated: `run_migrations` returning early on a newer database, and the v10 `symbol` backfill.

The four coupled places all moved: the `MIGRATIONS` registry, `SUPPORTED_SCHEMA_VERSION`, the `action.yml` cache key, and the same key in `docs/GITHUB_ACTION.md`. I ran the schema-sync check's own logic locally: Python 14, VS Code 14. There is a CHANGELOG entry under Fixed, since nearly every commit here carries one.

Gates on my side: `ruff check code_review_graph/` clean; `mypy code_review_graph/ --ignore-missing-imports --no-strict-optional` clean over 77 files; `upgrade-path` as above; full suite `4246 passed, 848 skipped, 2 xfailed, 2 xpassed` with one failure I should flag rather than round off: `test_watch_robustness.py::TestRealObserver::test_new_top_level_directory_is_indexed` failed once under full-suite load and passes 3 for 3 in isolation. It arrived with your merge, this branch does not touch it, and our diff against `staging` is seven files, none of them the watcher. Reading it as load-dependent timing rather than anything here, but it is your call.

PR #937, opened 2026-08-30
fix(parser): preserve full C# nested type identity

### LINKED ISSUE #934: [Bug]: C# nested type identity is truncated to one level, merging distinct methods into one node
### code-review-graph version

2.3.8

### Operating system

macOS

### Python version

3.12.4

### AI platform

claude-code

```shell
Method nodes inside a nested class are qualified with only the immediate enclosing class, not the full containing-type path. When two nested classes share a name — the standard MediatR-on-ASP.NET feature layout used by gothinkster/aspnetcore-realworld-example-app — their methods collapse into a single node. One method is silently lost. No uncertainty caveat covers this.
```

### Steps to reproduce

mkdir -p /tmp/crg-nested && cd /tmp/crg-nested && git init -q .
cat > Features.cs <<'EOF'
using MediatR;

namespace App.Articles;

public class Details
{
    public record Query(string Slug) : IRequest<string>;

    public class QueryHandler : IRequestHandler<Query, string>
    {
        public Task<string> Handle(Query request, CancellationToken ct) => Task.FromResult("details");
    }
}

public class Edit
{
    public record Query(string Slug) : IRequest<string>;

    public class QueryHandler : IRequestHandler<Query, string>
    {
        public Task<string> Handle(Query request, CancellationToken ct) => Task.FromResult("edit");
    }
}
EOF
code-review-graph build
python3 - <<'EOF'
import sqlite3
c = sqlite3.connect('.code-review-graph/graph.db')
for r in c.execute("select kind, qualified_name from nodes where kind!='File' order by qualified_name"):
    print(f"{r[0]:9} {r[1]}")
EOF

Expected
Features.cs::Details.QueryHandler.Handle and Features.cs::Edit.QueryHandler.Handle as separate nodes.

Actual:
Class     Features.cs::Details
Class     Features.cs::Details.Query
Class     Features.cs::Details.QueryHandler
Class     Features.cs::Edit
Class     Features.cs::Edit.Query
Class     Features.cs::Edit.QueryHandler
Function  Features.cs::QueryHandler.Handle      ← one node for two distinct methods

### Additional context

This issue seems to affect more than C#

Cause
parser.py:10298, when recursing into a class body

Julia is the only language that accumulates nested scope; every other language passes the bare simple name to _qualify (parser.py:14378). _julia_scope_join (parser.py:7428) already implements the needed join, including its repeated-prefix guard.

Impact

Silent node loss with no caveat. Any tool that addresses a method — callers_of, tests_for, impact radius, review context, refactor/rename — targets a node standing for two unrelated methods. This is the failure mode CRG's uncertainty machinery exists to prevent, and it is not covered.

Second symptom: only one level is ever retained

mkdir -p /tmp/crg-deep && cd /tmp/crg-deep && git init -q .
cat > Deep.cs <<'EOF'
namespace App;
public class A { public class B { public class C { public void M() { } } } }
EOF
code-review-graph build

Nodes: Deep.cs::A, Deep.cs::A.B, Deep.cs::B.C (should be A.B.C), Deep.cs::C.M (should be A.B.C.M).
The containment tree is broken as a result: A.B.C's CONTAINS edge comes from the File rather than from A.B, and the edge Deep.cs::C -> Deep.cs::C.M has a source node that does not exist.
Separately, querying by the correctly-qualified nested name does not resolve — code-review-graph query children_of Details.QueryHandler returns target not indexed: no node matching 'Details.QueryHandler' even though that node exists.

merlincat11:
## Summary

Method nodes inside a nested class were qualified with only the immediate enclosing
class, so two nested classes sharing a name collapsed their methods into a single
node — `Details.QueryHandler.Handle` and `Edit.QueryHandler.Handle` both became
`QueryHandler.Handle`, silently losing one method with no uncertainty caveat. Only
one level of nesting was ever retained.

- accumulate the containing-type path when recursing into a class body, as Julia already did
- emit `CONTAINS` edges against the full path
- resolve C# receivers against the new dotted parent names
- rebuild persistent graphs once through the existing version gate shared with C++

Fixes #934

## Why the resolver moves with the parser

These are one change, not two. `parent_name` is now a dotted path, and the old
receiver lookup reduced a scope to its last segment (`scope.rsplit(".", 1)[-1]`).
Shipping the parser fix alone would miss every nested type and stop resolving C#
receiver calls entirely.

## Resolution model

A receiver is matched against a containing-type path **exactly**, in two phases:

1. **lexical** — the receiver under each enclosing type, innermost first
   (`Outer.Consumer.D.E`, then `Outer.D.E`)
2. **exact** — the receiver itself, then forms with a leading *namespace* removed

There is deliberately **no suffix fallback**. Selecting a nested type from a bare
receiver would bind names C# cannot bind: a bare `QueryHandler` does not name
`Details.QueryHandler`, which has to be qualified. A lone indexed candidate would
otherwise be rewritten as a single match with no visibility check at all, and the
real type is often in a dependency that was never indexed.

Phase 1 matters because inside `Outer.Consumer` a receiver `D.E` denotes `Outer.D.E`
even when a top-level `D.E` exists.

Phase 2 exists only because namespaces are absent from `parent_name` — they live in
`csharp_namespaces_by_file` — so `App.Report.ExportHandler` is stored as
`Report.ExportHandler`. **The only thing a receiver may drop is a verified namespace
prefix.** A shortened form must prove the part it dropped is the namespace declared
by the candidate's defining file:

- file declares exactly `App` → resolves
- file declares `Other` → unresolved
- file declares both `App` and `Other` → unresolved, because file-level evidence
  cannot show which namespace encloses the candidate
- dropping a *type* name (`Details.QueryHandler` → `QueryHandler`) → unresolved,
  since `Details` is not a namespace

The multi-namespace case is conservative on purpose: it leaves genuine matches
unresolved rather than fabricating an edge. Resolving them properly needs per-node
namespace information rather than per-file, which is a follow-up.

Gated to C#. `_SCOPED_LANGUAGES` is `php`, `rust`, `csharp`; PHP and Rust are untouched.

## Scope

This PR is the identity fix and nothing else. Short-name addressing of a dotted
nested path (`children_of Details.QueryHandler`) was split out to #942 — it carried a
schema migration, an index, and Java-FQN precedence questions unrelated to this bug.
Nested types are addressable here by qualified name, which is what demonstrates the fix.

## Safety

- receivers only ever match a containing-type path exactly
- the sole permitted reduction is dropping an unambiguously verified namespace
- ambiguous or unverifiable receivers stay unresolved instead of being guessed
- PHP and Rust scoped resolution is unchanged
- no schema change

## Testing

- `uv run pytest tests/ --tb=short -q` (2996 passed, 9 skipped, 2 xpassed)
- distinct identities and `CONTAINS` parents for `Details.QueryHandler` vs `Edit.QueryHandler`
- three-level nesting (`A.B.C.M`)
- lexical shadowing: nested `Shadower.D.E` beats top-level `D.E`
- namespaced receiver resolves when the file declares exactly that namespace
- receiver stays unresolved when the dropped name is the wrong namespace,
  one of several namespaces in the file, or a type rather than a namespace
- bare receiver never selects a nested type, unique candidate or not
- legacy identity rebuild through the version gate

### DIFF
--- a/code_review_graph/incremental.py
+++ b/code_review_graph/incremental.py
@@ -77,6 +77,12 @@ logger = logging.getLogger(__name__)
 
 CPP_IDENTITY_VERSION = "1"
 _CPP_IDENTITY_METADATA_KEY = "cpp_identity_version"
+CSHARP_IDENTITY_VERSION = "1"
+_CSHARP_IDENTITY_METADATA_KEY = "csharp_identity_version"
+_IDENTITY_FORMATS = (
+    ("cpp", "C++", _CPP_IDENTITY_METADATA_KEY, CPP_IDENTITY_VERSION),
+    ("csharp", "C#", _CSHARP_IDENTITY_METADATA_KEY, CSHARP_IDENTITY_VERSION),
+)
 
 
 def _run_python_resolver(store: GraphStore) -> Optional[dict]:
@@ -1273,8 +1279,15 @@ def full_build(
     total_nodes = 0
     total_edges = 0
     errors = []
-    cpp_errors: set[str] = set()
+    identity_errors: dict[str, set[str]] = {
+        language: set() for language, _, _, _ in _IDENTITY_FORMATS
+    }
     file_count = len(files)
+
+    def record_identity_error(rel_path: str) -> None:
+        language = parser.detect_language(repo_root / rel_path)
+        if language in identity_errors:
+            identity_errors[language].add(str(rel_path))
 
     use_serial = os.environ.get("CRG_SERIAL_PARSE", "") == "1"
 
@@ -1291,13 +1304,11 @@ def full_build(
                 total_edges += len(edges)
             except (OSError, PermissionError) as e:
                 errors.append({"file": rel_path, "error": str(e)})
-                if parser.detect_language(full_path) == "cpp":
-                    cpp_errors.add(str(rel_path))
+                record_identity_error(rel_path)
             except Exception as e:
                 logger.warning("Error parsing %s: %s", rel_path, e)
                 errors.append({"file": rel_path, "error": str(e)})
-                if parser.detect_language(full_path) == "cpp":
-                    cpp_errors.add(str(rel_path))
+                record_identity_error(rel_path)
             if i % 50 == 0 or i == file_count:
                 logger.info("Progress: %d/%d files parsed", i, file_count)
     else:
@@ -1315,8 +1326,7 @@ def full_build(
                 if error:
                     logger.warning("Error parsing %s: %s", rel_path, error)
                     errors.append({"file": rel_path, "error": error})
-                    if parser.detect_language(repo_root / rel_path) == "cpp":
-                        cpp_errors.add(str(rel_path))
+                    record_identity_error(rel_path)
                     continue
                 full_path = repo_root / rel_path
                 store.store_file_nodes_edges(
@@ -1332,8 +1342,9 @@ def full_build(
 
     store.set_metadata("last_updated", time.strftime("%Y-%m-%dT%H:%M:%S"))
     store.set_metadata("last_build_type", "full")
-    if not cpp_errors:
-        store.set_metadata(_CPP_IDENTITY_METADATA_KEY, CPP_IDENTITY_VERSION)
+    for language, _, metadata_key, version in _IDENTITY_FORMATS:
+        if not identity_errors[language]:
+            store.set_metadata(metadata_key, version)
     _store_vcs_metadata(repo_root, store)
     store.commit()
 
@@ -1375,12 +1386,16 @@ def incremental_update(
     parser = CodeParser(repo_root)
     ignore_patterns = _load_ignore_patterns(repo_root)
 
-    if (
-        store.get_metadata(_CPP_IDENTITY_METADATA_KEY) != CPP_IDENTITY_VERSION
-        and store.has_nodes_for_language("cpp")
-    ):
+    pending_identity_formats = [
+        (language, label)
+        for language, label, metadata_key, version in _IDENTITY_FORMATS
+        if store.get_metadata(metadata_key) != version
+        and store.has_nodes_for_language(language)
+    ]
+    if pending_identity_formats:
         logger.info(
-            "C++ identity format changed; rebuilding the graph before incremental update",
+            "%s identity format changed; rebuilding the graph before incremental update",
+            ", ".join(label for _, label in pending_identity_formats),
         )
         rebuilt = full_build(repo_root, store)
         return {
@@ -1397,6 +1412,7 @@ def incremental_update(
             "event_resolution": rebuilt["event_resolution"],
             "temporal_resolution": rebuilt["temporal_resolution"],
             "hcl_resolution": rebuilt["hcl_resolution"],
+            "scoped_resolution": rebuilt["scoped_resolution"],
         }
 
     # Determine changed files
@@ -1507,7 +1523,8 @@ def incremental_update(
     if files_updated:
         store.set_metadata("last_updated", time.strftime("%Y-%m-%dT%H:%M:%S"))
         store.set_metadata("last_build_type", "incremental")
-        store.set_metadata(_CPP_IDENTITY_METADATA_KEY, CPP_IDENTITY_VERSION)
+        for _, _, metadata_key, version in _IDENTITY_FORMATS:
+            store.set_metadata(metadata_key, version)
         _store_vcs_metadata(repo_root, store)
         store.commit()
 
--- a/code_review_graph/parser.py
+++ b/code_review_graph/parser.py
@@ -5376,6 +5376,7 @@ class CodeParser:
                     result,
                     child.child_by_field_name("name"),
                     declared_type,
+                    keep_qualifier=True,
                 )
 
         elif language == "csharp" and node.type == "parameter":
@@ -5383,6 +5384,7 @@ class CodeParser:
                 result,
                 node.child_by_field_name("name"),
                 node.child_by_field_name("type"),
+                keep_qualifier=True,
             )
 
         elif language == "php" and node.type == "assignment_expression":
@@ -5426,18 +5428,28 @@ class CodeParser:
         return {name} if name else set()
 
     @classmethod
-    def _store_typed_binding(cls, result: dict[str, str], name_node, type_node) -> None:
+    def _store_typed_binding(
+        cls,
+        result: dict[str, str],
+        name_node,
+        type_node,
+        *,
+        keep_qualifier: bool = False,
+    ) -> None:
         if name_node is None or type_node is None:
             return
         name = name_node.text.decode("utf-8", errors="replace")
         type_name = cls._base_type_name(
             type_node.text.decode("utf-8", errors="replace"),
+            keep_qualifier=keep_qualifier,
         )
         if name and type_name:
             result[name] = type_name
 
     @classmethod
-    def _base_type_name(cls, annotation: str) -> Optional[str]:
+    def _base_type_name(
+        cls, annotation: str, *, keep_qualifier: bool = False,
+    ) -> Optional[str]:
         """Return the receiver class from a generic/nullable annotation."""
         current = annotation.strip()
         for _ in range(4):
@@ -5464,7 +5476,7 @@ class CodeParser:
                 "string", "unknown", "void",
             ):
                 return None
-            return outer
+            return match.group(1) if keep_qualifier else outer
         return None
 
     def _resolve_typed_method_target(
@@ -5490,7 +5502,7 @@ class CodeParser:
             # class receiver cannot be resolved to its defining file during
             # a single-file parse. Emit the receiver class as a scope target
             # for the graph-wide scoped resolver (#612).
-            base_type = self._base_type_name(type_name)
+            base_type = self._base_type_name(type_name, keep_qualifier=True)
             if not base_type:
                 return None
             return f"{base_type}::{method}"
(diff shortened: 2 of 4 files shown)


merlincat11:
The concern is real — I reproduced it before changing anything.

Your exact case (`Details.QueryHandler` nested, bare `QueryHandler` receiver in another file, `using External;`) resolved to `Details.QueryHandler.Handle` with `scoped_via: "single_match"`. With one indexed candidate the suffix map hands back a unique answer, `csharp_disambiguate` is never reached, and no visibility check runs at all.

So I took the fix you suggested and dropped the containing-type suffix fallback entirely.

It turned out to cost nothing. Your read was right that the valid cases are already covered:

- receivers bound through an enclosing type are handled by the lexical phase
- namespace-qualified receivers are handled by the exact phase, which shortens progressively — `App.Report.ExportHandler` misses exactly, then matches at `Report.ExportHandler`, because the namespace is absent from `parent_name`

So the separate namespace-metadata validation you offered as a fallback plan isn't needed; the exact phase already gets there. The whole suite passes with the map removed (2994 passed, 9 skipped, 2 xpassed), and the resolver is now two phases that both match a containing-type path exactly, with no arbitrary suffix matching left in it.

Added `test_bare_receiver_never_selects_a_nested_type`, which asserts the call stays unresolved. I checked it fails against the previous commit with exactly the mislink you described, so it's a real regression test rather than one that passes vacuously:

```
AssertionError: {'LoneNested.cs::Wrapper.LoneHandler.Handle'}
assert {...} == {'Handle'}
```

Also renamed `test_longer_suffix_match_beats_shorter_exact_type` — it described the mechanism that's now gone rather than what it asserts.

Pushed as 0e9b891.

merlincat11:
Confirmed before changing anything — your case resolved to `Wrong.cs::Report.ExportHandler.Run` with `scoped_via: "single_match"`, exactly as you described. One candidate means `csharp_disambiguate` never runs, so nothing downstream catches it.

Fixed in d62623b, and it's the check you proposed in your previous round. I said then that the exact phase made namespace validation unnecessary — that was wrong, and this is the case that shows it.

Shortening is only legitimate in the first place because namespaces are absent from `parent_name`. So a shortened form now has to prove the part it dropped really is a namespace declared by the candidate's defining file, using the `csharp_namespaces` metadata the resolver already builds:

- `App.Report.ExportHandler` → `Report.ExportHandler` where the file declares `namespace App` — resolves
- the same shortening where the file declares `namespace Other` — stays unresolved

Unresolved is the right answer there, since the requested type may simply not be indexed.

One useful side effect: this closes the last route by which a leading *type* name could be dropped. Shortening `Details.QueryHandler` to `QueryHandler` now fails the check too, because `Details` is a type rather than a namespace — the same class of error as the suffix map removed in the previous round, now ruled out structurally rather than by having deleted one code path.

Added `test_shortened_receiver_requires_the_dropped_part_to_be_a_namespace`. Verified it fails against 0e9b891 with the exact mislink, so it isn't passing vacuously:

```
AssertionError: {'WrongNamespace.cs::Ledger.AuditHandler.Run'}
```

2995 passed, 9 skipped, 2 xpassed; ruff and mypy clean.

merlincat11:
Confirmed — the file-level set really does leak. Your case resolved to `Definitions.cs::Ledger.AuditHandler.Run` via `single_match`, with the file metadata reading `{"csharp_namespaces": ["App", "Other"]}`. You put it exactly right: the check proved `App` belongs to the candidate's *file*, not to the candidate.

Took the conservative option, in b9c93f3:

```python
if csharp_namespaces_by_file.get(file, set()) != {dropped_namespace}:
    reject
```

So a shortened form now requires the candidate's file to declare exactly the dropped namespace. Genuine matches in multi-namespace files are left unresolved, which as you say is the right trade for a resolver that must not fabricate edges.

I deliberately did **not** take the per-node namespace option here. It's the better architecture and I agree it's where this should end up, but it changes what the parser records per node, which is well outside a node-identity bug fix — this PR has already been split once for scope. Worth its own issue; happy to file it if you want, and I'd rather it be discussed on its own merits than smuggled in here.

Added `test_dropped_namespace_must_be_the_files_only_namespace`, verified failing against d62623b with the exact mislink:

```
AssertionError: {'MultiNamespace.cs::Journal.PostHandler.Post'}
```

The invariant across the last three rounds is now a single sentence: **the only thing a receiver may drop is an unambiguously verified namespace prefix.** Wrong namespace, one-of-several namespaces, and a leading type name are all rejected by the same check.

Also corrected the stale PR description — thanks for catching it. It no longer describes a third suffix phase, and now documents the multi-namespace limitation and its rationale.

2996 passed, 9 skipped, 2 xpassed; ruff and mypy clean.

merlincat11:
Fixed in 2fb7b6d, and your test case is in. One correction to the framing, though: this is **not a regression** from this branch.

I ran your exact case against `main`:

| | result |
| --- | --- |
| `main` | `Indexed.cs::ExportHandler.Run` — resolved |
| this branch, before 2fb7b6d | `Indexed.cs::App.Report.ExportHandler.Run` — resolved |
| this branch, now | unresolved |

`main` mislinks the same call to the same wrong type, and it does so matching the *bare* name `ExportHandler`, so on `main` any `ExportHandler.Run` anywhere collides. This branch already narrowed that to requiring the full containing path to coincide. So the bug is a long-standing consequence of namespaces sitting outside the C# identity, not something introduced here — worth being accurate about, since it affects whether it blocks this PR.

That said, you're right that it's real and that the PR asserted an invariant it didn't hold, so I've closed it. When a receiver names no namespace, the candidate's namespace must now be reachable from the call site: same file, global namespace, a `using`, or a namespace the caller declares. `imports_by_file` and `csharp_namespaces_by_file` already carry all of that, so no new metadata was needed — and no existing test changed behaviour.

Added `test_exact_path_match_still_checks_the_candidate_namespace`, verified failing against b9c93f3 with exactly your prediction:

```
AssertionError: {'WrongNamespaceExactPath.cs::Vault.Archive.StoreHandler.Store'}
```

On the architecture: I still think per-node namespaces belong in their own change rather than here. This PR fixes a node-identity bug and has already been split once for scope; folding a parser-level identity change into it would repeat exactly the mistake that split corrected. I'm not disputing that it's the right end state — file-level evidence remains genuinely unable to decide the multi-namespace case, and that limitation is documented in the PR description. I'd rather that land as its own reviewable change than as a fifth follow-up commit here.

2997 passed, 9 skipped, 2 xpassed; ruff and mypy clean.

merlincat11:
You're right on all three, and the third one is decisive — so I've **reverted 2fb7b6d** rather than patch it.

I measured each case against `main`:

| case | `main` | 2fb7b6d | after revert |
| --- | --- | --- | --- |
| same file, different namespaces | RESOLVED (`Both.cs::StoreHandler.Store`) | RESOLVED (wrong) | RESOLVED (wrong) |
| `global using` in another file | **RESOLVED (correct)** | **unresolved** | RESOLVED (correct) |
| exact path, wrong namespace | RESOLVED (wrong) | unresolved | RESOLVED (wrong) |

The global-using row is the one that matters: my predicate broke a call `main` gets right. Meanwhile it didn't fix your first two cases at all — same-file and file-scoped-`using` both still over-permit. So it under-permits and over-permits simultaneously, which is the signature of the wrong evidence rather than a bug in the rule. Trading a false negative for a partial fix to a false positive isn't a good trade, and adding global-using support would have fixed one row and left the other two for a sixth round.

After the revert the branch is at parity with `main` on every one of these: the wrong-namespace mislink remains, but it is pre-existing, `main` hits it on the bare name `StoreHandler`, and this branch already narrows it to requiring the full containing-type path to coincide.

Two tests to keep it honest:

- `test_exact_path_match_should_check_the_candidate_namespace` — kept as `xfail` with the reason recorded, so the limitation is documented rather than forgotten, and flips to passing the day it's fixed properly
- `test_global_using_target_still_resolves` — new, so any future visibility work has to preserve the case my patch broke

On the underlying point, I now agree with you without reservation: file-level namespace evidence cannot establish this invariant in either direction, and per-node namespaces are the actual fix. What I'd push back on is doing it in this PR. It is a parser-level change to the C# identity, it needs its own rebuild-gate consideration, and this PR is a node-identity bug fix that has already been split once for scope. Four rounds here have each closed one instance of the same root cause; the fifth should close the cause. I'd rather open that as its own issue referencing this thread than grow this one again.

Happy to file it now if you want, and to hold #937 until it lands if you'd prefer they ship together.

2997 passed, 9 skipped, 1 xfailed, 2 xpassed; ruff and mypy clean.

merlincat11:
Holding this PR pending #946.

Filed the root cause as its own issue: C# namespaces are absent from the graph identity, so namespace-sensitive resolution is undecidable from the file-level evidence available. It collects all five failing cases with their `main`-vs-branch measurements, the reverted attempt and why it failed in both directions, and the design questions worth settling first — sidecar vs identity change, namespace-body-scoped usings, and project-wide global usings.

To be clear about what this PR does and doesn't do in the meantime: it fixes #934, and it is at parity with `main` on every namespace case in #946 — none of them are regressions, and case 1 is strictly narrowed here. So this is not blocked on correctness so much as on building it against a foundation that can actually express the invariant, rather than adding a fifth file-level heuristic on top of four.

Marking it draft. Once #946 lands, the `xfail` here should flip to passing, and `test_global_using_target_still_resolves` should stay green.

Thanks for the persistence across these rounds — the same-file and `global using` cases in particular were things I'd have shipped wrong.

tirth8205:
Superseded by #964 at the namespace-identity boundary, so no code from this draft has been retained. Its discussion explicitly parks it pending #946, and the known wrong-namespace case remains an expected failure; #964 itself is also held because it loses supported generic-receiver context.

tirth8205:
This no longer merges into `staging`. The one conflict is `code_review_graph/incremental.py`, and the branch is 224 commits behind.

```bash
git fetch origin
git merge origin/staging
# resolve, then
uv run pytest tests/ -q
git push
```

PR #939, opened 2026-08-30
fix(parser): resolve generic inheritance targets

### LINKED ISSUE #935: [Bug]: C# and Java. Inheritance through a generic base class is dropped, and inheritors_of reports it as a confirmed absence
### code-review-graph version

2.3.8

### Operating system

macOS

### Python version

3.12.4

### AI platform

claude-code

```shell
INHERITS edges store the base type as raw source text. When the base is generic, the target is Base<Arg>, which never matches the class node Base, so the relationship is invisible. Worse than the omission: inheritors_of returns 0 with the confidence string "is indexed and no such edge is recorded" — an affirmative claim of absence, when the edge is present in the edges table and merely unresolved.
```

### Steps to reproduce

mkdir -p /tmp/crg-generic && cd /tmp/crg-generic && git init -q .
cat > Bases.cs <<'EOF'
namespace App;

public abstract class PlainBase { }
public abstract class GenericBase<T> { }

public class FromPlain : PlainBase { }
public class FromGeneric : GenericBase<string> { }
EOF
code-review-graph build
code-review-graph query inheritors_of PlainBase
code-review-graph query inheritors_of GenericBase

Actual
inheritors_of PlainBase    → count=1  ['FromPlain']   confidence=None
inheritors_of GenericBase  → count=0  []
   confidence: "'…::GenericBase' is indexed and no such edge is recorded; graph currency unverified"
INHERITS  Bases.cs::FromGeneric -> GenericBase<string>

The target is the raw text GenericBase<string>, which never matches node Bases.cs::GenericBase.

Expected

inheritors_of GenericBase returns FromGeneric. Failing that, a caveat that does not assert absence — no LanguageGap in uncertainty.py covers inheritors_of for C#, so this false negative is delivered unqualified.

For Java

package app;
abstract class PlainBase { }
abstract class GenericBase<T> { }
class FromPlain extends PlainBase { }
class FromGeneric extends GenericBase<String> { }

inheritors_of PlainBase = 1, inheritors_of GenericBase = 0. 

### Additional context

eShop project routes ordering commands through IdentifiedCommandHandler<T, R>. Every concrete handler inherits from a closed generic
public class CreateOrderIdentifiedCommandHandler : IdentifiedCommandHandler<CreateOrderCommand, bool> { }

So every one of them is invisible to inheritors_of, and CRG states positively that the base class has no inheritors. Since these handlers also declare no Handle of their own — it's inherited from the abstract base — there is no other edge connecting a concrete handler to its behavior.

Suggested fix

Resolve INHERITS targets by stripping type arguments before node matching, retaining the full constructed type in extra so the argument list is not lost (it is needed for any later type-substitution work). At minimum, stop emitting "no such edge is recorded" when an unresolved edge of that kind exists from the target.

### LINKED ISSUE #941: Existing graphs keep pre-#935 generic INHERITS targets until rebuilt
Follow-up to #939, which fixes #935 for newly parsed files only.

## Situation

#939 makes the parser store the declaration name (`GenericBase`) as the `INHERITS` target
instead of the constructed spelling (`GenericBase<string>`). Graphs built before that
change still hold the old targets, so `inheritors_of` keeps returning nothing for a
generic base until those files are reparsed.

## Why no migration shipped

#939 originally carried a v10 migration that rewrote the rows in place. It was dropped
deliberately:

- v2-v9 are pure DDL. v10 would have been the first migration in this codebase to mutate
  and delete graph rows.
- A migration sees only stored text, not the syntax tree, so it had to reimplement the
  parser's type-argument erasure — including C#/Java comment handling, because `<` inside
  a comment is not a delimiter. Two implementations of one transformation, required to
  stay in lockstep. Both review findings against that PR landed on exactly that seam.

## Current recovery path

- `code-review-graph build` regenerates everything.
- Otherwise self-healing per file: `store_file_nodes_edges` replaces a file's edges on
  every reparse, so any touched file picks up correct targets via the normal watch/update
  path.

## Possible work

- Note in release notes that a rebuild is needed for full effect on existing graphs.
- Consider a targeted reindex command (invalidate stored file hashes for files whose
  `INHERITS` targets contain `<`, so the next `update` reparses them with the real
  parser). This keeps a single source of truth for the transformation, unlike a
  text-rewriting migration.

merlincat11:
## Summary

Resolve Java and C# generic inheritance against the base type's declaration name instead
of the constructed spelling, so `class Child : GenericBase<string>` links to
`GenericBase`. Every type argument is retained on the edge in `extra.constructed_types`.

Fixes #935. Closes #941.

## How the target is derived

The declaration name comes from the syntax tree, skipping the type-argument and comment
subtrees. It is deliberately not derived by scanning for `<` and `>`: those characters
also occur in trivia the grammar ignores, so `class C : I</* < */ string> {}` — legal C#
— reads as unbalanced to a character counter and would keep the constructed spelling as
its target, leaving the original bug in place.

`_get_base_type_nodes` is split out of `_get_bases` so the source spelling and the target
come from a single traversal.

## Repeated generic bases

C# permits one class to implement several closed constructions of the same generic
interface (`class C : I<int>, I<string> {}`). These erase to one target and share the
class declaration line, so `upsert_edge`'s identity — `(kind, source, target, file_path,
line)`, which excludes `extra` — would treat one-edge-per-base as the same edge and keep
only the last. Bases are grouped by target and emitted as one edge recording every
construction.

## Existing graphs

Changing a persisted spelling leaves old graphs stale, since nothing reparses an
unchanged file. This uses the mechanism already in the project for that situation,
`CPP_IDENTITY_VERSION` in `incremental.py`: an `INHERITS_IDENTITY_VERSION` records the
spelling a graph was built with, and `incremental_update` rebuilds once when it is
missing or stale and the graph holds Java or C# nodes.

The rebuild runs the real parser, so there is no second implementation of the
normalization to keep in step. An earlier revision of this PR carried a v10 SQL migration
that rewrote rows from stored text; it was dropped because it duplicated the parser's
logic — including comment handling — and both review findings against it landed on that
seam.

`cpp_errors` becomes `failed_languages`, since only its emptiness was ever used and the
same guard is needed per language: a file that failed to parse keeps its old edges, so
the version must not be marked current for that language. C++ behaviour is unchanged.

Verified on a graph built with pre-PR code, then updated with no file changes: the
rebuild triggers, `inheritors_of GenericBase` returns `FromGeneric`, and a second update
does not rebuild again.

## What the target is (and is not)

The target is a **lookup name**, not a language-level declaration identity. It drops
generic arity (`G<T>` and `G<T, U>` both yield `G`) and carries no namespace or package.

`inheritors_of` already resolves non-generic bases the same way: class nodes are keyed by
bare name, so `G<T>` and `G<T, U>` collapse to one node regardless of this PR, and the
bare-name fallback already unions same-named classes. This brings generic bases to parity
with that path rather than introducing a new resolution model. The constructed spelling is
preserved losslessly in `extra.constructed_types`, so a namespace-aware resolver later has
the evidence it needs — tracked in #940, deliberately not attempted here since it
reproduces on `main` for non-generic bases.

## Safety

- only Java/C# `INHERITS` targets containing type arguments change
- malformed or unreadable type text is left unchanged
- no schema change and no SQL data migration
- the identity rebuild runs once, then records the version

## Testing

- `uv run pytest --tb=short -q --cov=code_review_graph --cov-report=term-missing --cov-fail-under=65` (2993 passed, 9 skipped, 2 xpassed; 85.07% coverage)
- `uv run ruff check code_review_graph/ tests/test_multilang.py`

Regression tests cover nested generic arguments (Java and C#), angle brackets inside
trivia, repeated closed constructions, unbalanced type text, and the upgrade rebuild.

### DIFF
--- a/code_review_graph/incremental.py
+++ b/code_review_graph/incremental.py
@@ -78,6 +78,13 @@ logger = logging.getLogger(__name__)
 CPP_IDENTITY_VERSION = "1"
 _CPP_IDENTITY_METADATA_KEY = "cpp_identity_version"
 
+# Bumped when the parser changes how a Java/C# INHERITS target is spelled, so
+# graphs built by an older version are rebuilt instead of keeping targets that
+# no longer match any node. See #935.
+INHERITS_IDENTITY_VERSION = "1"
+_INHERITS_IDENTITY_METADATA_KEY = "inherits_identity_version"
+_INHERITS_IDENTITY_LANGUAGES = ("java", "csharp")
+
 
 def _run_python_resolver(store: GraphStore) -> Optional[dict]:
     """Run repository-wide Python import resolution without failing a build."""
@@ -1273,7 +1280,7 @@ def full_build(
     total_nodes = 0
     total_edges = 0
     errors = []
-    cpp_errors: set[str] = set()
+    failed_languages: set[str] = set()
     file_count = len(files)
 
     use_serial = os.environ.get("CRG_SERIAL_PARSE", "") == "1"
@@ -1291,13 +1298,11 @@ def full_build(
                 total_edges += len(edges)
             except (OSError, PermissionError) as e:
                 errors.append({"file": rel_path, "error": str(e)})
-                if parser.detect_language(full_path) == "cpp":
-                    cpp_errors.add(str(rel_path))
+                failed_languages.add(parser.detect_language(full_path) or "")
             except Exception as e:
                 logger.warning("Error parsing %s: %s", rel_path, e)
                 errors.append({"file": rel_path, "error": str(e)})
-                if parser.detect_language(full_path) == "cpp":
-                    cpp_errors.add(str(rel_path))
+                failed_languages.add(parser.detect_language(full_path) or "")
             if i % 50 == 0 or i == file_count:
                 logger.info("Progress: %d/%d files parsed", i, file_count)
     else:
@@ -1315,8 +1320,9 @@ def full_build(
                 if error:
                     logger.warning("Error parsing %s: %s", rel_path, error)
                     errors.append({"file": rel_path, "error": error})
-                    if parser.detect_language(repo_root / rel_path) == "cpp":
-                        cpp_errors.add(str(rel_path))
+                    failed_languages.add(
+                        parser.detect_language(repo_root / rel_path) or "",
+                    )
                     continue
                 full_path = repo_root / rel_path
                 store.store_file_nodes_edges(
@@ -1332,8 +1338,10 @@ def full_build(
 
     store.set_metadata("last_updated", time.strftime("%Y-%m-%dT%H:%M:%S"))
     store.set_metadata("last_build_type", "full")
-    if not cpp_errors:
+    if "cpp" not in failed_languages:
         store.set_metadata(_CPP_IDENTITY_METADATA_KEY, CPP_IDENTITY_VERSION)
+    if failed_languages.isdisjoint(_INHERITS_IDENTITY_LANGUAGES):
+        store.set_metadata(_INHERITS_IDENTITY_METADATA_KEY, INHERITS_IDENTITY_VERSION)
     _store_vcs_metadata(repo_root, store)
     store.commit()
 
@@ -1375,12 +1383,24 @@ def incremental_update(
     parser = CodeParser(repo_root)
     ignore_patterns = _load_ignore_patterns(repo_root)
 
+    stale_identity = None
     if (
         store.get_metadata(_CPP_IDENTITY_METADATA_KEY) != CPP_IDENTITY_VERSION
         and store.has_nodes_for_language("cpp")
     ):
+        stale_identity = "C++"
+    elif (
+        store.get_metadata(_INHERITS_IDENTITY_METADATA_KEY) != INHERITS_IDENTITY_VERSION
+        and any(
+            store.has_nodes_for_language(language)
+            for language in _INHERITS_IDENTITY_LANGUAGES
+        )
+    ):
+        stale_identity = "Java/C# inheritance"
+    if stale_identity:
         logger.info(
-            "C++ identity format changed; rebuilding the graph before incremental update",
+            "%s identity format changed; rebuilding the graph before incremental update",
+            stale_identity,
         )
         rebuilt = full_build(repo_root, store)
         return {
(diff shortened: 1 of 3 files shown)


merlincat11:
Parking this as a draft rather than merging. The review findings hold up, and the last round showed the approach itself is the problem, not the details.

## What I verified

All three findings reproduce, and all three are introduced by this branch. Each was built with the same source under `main` and under this branch:

**Spring resolution picks the wrong implementation.** `spring_resolver.py` builds its implementation map straight from `INHERITS.target_qualified`. INJECTS already reduces `Store<Integer>` to `Store`, so once INHERITS erases to `Store` too, they meet:

```
main : CALLS Consumer.run -> Store.get
here : CALLS Consumer.run -> StringStore.get     # implements Store<String>
```

This rewrites a CALLS edge, so it propagates into impact radius, flows and review context.

**C# arity.** With `interface I {}` and `interface I<T> {}` in one namespace, querying the non-generic `I`:

```
main : 0 results
here : ['Child']        # Child implements I<int>
```

Distinct from #940, which is namespace ambiguity already present on `main`.

**Identity-version rebuild loop.** One Java file that always fails to parse makes every incremental update a full rebuild, permanently. Filed as #944, since the same loop is reachable on `main` for C++.

## Why parking rather than patching

The root cause is that this branch makes `INHERITS.target_qualified` lossy, and that field has more than one consumer. I audited `inheritors_of` and never checked `spring_resolver.py`, which was my mistake — it is exactly the "audit all consumers before changing a shared identity" point raised earlier in review, which I treated as architectural preference rather than concrete risk.

Fixing both regressions inside the current design means changing Spring resolution and the arity handling as well, which grows this PR into three subsystems. The better move is the separation asked for repeatedly in review: keep the raw spelling, derive a canonical declaration reference that carries arity, and have resolution consume that — rather than reshaping stored data so a bare-name fallback happens to match.

Captured in #943 with the reproductions above as constraints and #935 as the motivating test case.

## State of the branch

The parser work here is still useful input to #943: base names are derived from the Tree-sitter nodes rather than by scanning angle brackets, which is what a character counter gets wrong on `class C : I</* < */ string> {}`. History also records a v10 SQL migration that was dropped because it reimplemented parser normalization against stored text.

Not closing, so the branch and discussion stay available for #943.

tirth8205:
Changes required: replacing a constructed inheritance target with its bare name loses arity and permits incorrect binding. Build `I.cs` containing `interface I {}`, `interface I<T> {}` and `class Child : I<int> {}` with `full_build(root, store)`; this PR changes the INHERITS target from `I<int>` to `I`, which also names the non-generic declaration. Preserve construction and declaration identity together, including the downstream generic-injection case described in the discussion.

tirth8205:
This no longer merges into `staging`. The one conflict is `code_review_graph/incremental.py`, and the branch is 224 commits behind.

```bash
git fetch origin
git merge origin/staging
# resolve, then
uv run pytest tests/ -q
git push
```

PR #949, opened 2026-09-01
fix(opencode): use current plugin hook API

### LINKED ISSUE #947: [Bug]: opencode plugin template uses removed `app.on` API — fails on every startup
### code-review-graph version

2.3.8

### Operating system

macOS

### Python version

3.12.4

### AI platform

opencode

```shell
`code-review-graph install --platform opencode` writes
`~/.config/opencode/plugins/crg-plugin.ts` whose factory registers hooks via
`app.on(...)` (EventEmitter style). opencode's current plugin API calls the
factory with `{ project, client, $, directory, worktree }` and expects a hooks
object back — there is no `app`, so the loader fails on every startup and the
plugin never runs:

    level=ERROR message="failed to load plugin" path=.../crg-plugin.ts
    error="app.on is not a function"

Affected template: `_opencode_plugin_content()` in `code_review_graph/skills.py`
(all three hooks: `file.edited`, `session.created`, `tool.execute.before`).
```

### Steps to reproduce

the `run($, cmd)` helper uses
`$`+"`"+`${cmd}`+"`"+`` — in Bun shell, interpolating a full command string as
the whole template escapes it into ONE executable name and throws (verified).
Even if `app.on` worked, every `run()` call would silently no-op through the
swallowed catch. Commands must be literal tagged templates.

(verified with routing tests):

    import type { Plugin } from "@opencode-ai/plugin"

    export default (async ({ $ }: { $: any }) => {
      return {
        event: async ({ event }: { event: { type: string } }) => {
          if (event.type === "file.edited") {
            try { await $`code-review-graph update --skip-flows`.quiet() } catch {}
          }
          if (event.type === "session.created") {
            try {
              const result = await $`code-review-graph status`.quiet()
              const output = result.stdout?.toString().trim()
              if (output) console.log("[code-review-graph]", output)
            } catch {}
          }
        },
        "tool.execute.before": async (input: { tool: string; args?: { command?: string } }) => {
          try {
            if (/^git\s+commit/i.test(input?.args?.command ?? "")) {
              const result = await $`code-review-graph detect-changes --brief`.quiet()
              const output = result.stdout?.toString().trim()
              if (output) console.log("[code-review-graph] Pre-commit analysis:\n" + output)
            }
          } catch {}
        },
      }
    }) satisfies Plugin

Note: `install_opencode_plugin()` unconditionally overwrites the file
("If the file already exists it is overwritten"), so hand-fixes get
clobbered by any re-run of `install`.

### Additional context

_No response_

logan683:
# Pull Request

## Linked issue

Closes #947

## What & why

OpenCode's current plugin API calls a plugin factory with its context and expects the factory to
return a hooks object. The generated CRG plugin instead called the removed `app.on(...)` API, so
OpenCode failed to load it on every startup and none of its three behaviors registered.

This change:

- routes `file.edited` and `session.created` through OpenCode's generic `event` hook;
- returns the named `tool.execute.before` hook and reads the current
  `input.tool` / `output.args.command` arguments;
- invokes each CRG command as a literal Bun shell template, removing the unused dynamic-command
  helper that would have escaped a whole command as one executable name;
- passes OpenCode's active `directory` through `--repo` on every command, so a global plugin cannot
  update or query the process cwd or another repository;
- preserves graceful failure behavior and the existing console-visible status/pre-commit output;
- adds a regression test that forbids `app.on(` and pins the returned-hooks shape.

`console.log` status remains user-visible console output; this PR does not claim or add model-context
injection.

Because the plugin now loads for the first time, its existing per-file update behavior and awaited
pre-commit analysis are also live for the first time. Debouncing updates and bounding pre-commit
latency are follow-up hardening work, not changes to this compatibility fix; the anchored
`git commit` matcher is likewise preserved from the existing generator.

## How it was tested

```text
uv run pytest tests/test_skills.py -k OpenCodePlugin -q
18 passed, 193 deselected

uv run pytest tests/test_skills.py -q -k "not test_post_tool_use_command_handles_large_stdin_payload"
208 passed, 1 skipped, 1 deselected

uv run --with "ruff>=0.3,<1" ruff check code_review_graph/
All checks passed!

uv run --with "mypy>=1.10,<3" mypy code_review_graph/ --ignore-missing-imports --no-strict-optional
Success: no issues found in 72 source files
```

Runtime checks on OpenCode 1.18.25:

- imported the generated TypeScript with Bun and invoked all three hooks; observed exactly
  `status`, `update --skip-flows`, and `detect-changes --brief`;
- regenerated the user plugin from the patched installer and started a fresh OpenCode/Qwen session;
  no `app.on is not a function` load error occurred;
- replaced CRG on `PATH` with a status canary and confirmed the status line appears in the
  `opencode run` console;
- captured that canary's real arguments and confirmed OpenCode 1.18.25 supplies the active directory
  as `status --repo <active-directory>`.

The full suite on Windows reported `2,914 passed, 60 failed, 21 skipped, 2 xpassed`. The failures are
outside the touched OpenCode generator/tests and are Windows/environment cases (path separator
expectations, unavailable symlink privilege, WSL `/bin/bash`, and file-mode behavior). The focused
OpenCode suite, package lint, type check, and live client checks are green.

## Checklist

- [ ] All tests pass: `uv run pytest tests/ --tb=short -q` — Windows/environment failures disclosed above
- [x] Docs updated where behavior changed (`CHANGELOG.md` and generator docstring/comments)


tirth8205:
Blocked pending a released OpenCode run: passing generated-string tests does not establish that the returned hooks are discovered or invoked. Install the plugin, open two repositories, then trigger a session, file edits and a commit in each; verify the active directory, single refresh invocation and bounded completion of the awaited update and pre-commit commands. The local 18-test result establishes template output only.

tirth8205:
This no longer merges into `staging`. Conflicts in `CHANGELOG.md`.

```bash
# resolve, then
uv run pytest tests/ -q
git push
```

When resolving:
- Rebase or merge origin/staging into fix/opencode-plugin-api and resolve the CHANGELOG.md [Unreleased] conflict (keep both staging's entries and the PR's 6-line OpenCode fix entry).
- After the conflict is resolved, gates (pytest on 3.13, ruff, mypy) and edge-case coverage of the new OpenCode plugin generator (`_opencode_plugin_content`) still need to be assessed; they were not evaluated in this run.

tirth8205:
This no longer merges into `staging`. The one conflict is `CHANGELOG.md`, and the branch is 224 commits behind.

```bash
git fetch origin
git merge origin/staging
# resolve, then
uv run pytest tests/ -q
git push
```

PR #951, opened 2026-09-01
feat(mcp): let a tool call name its own graph data directory

### LINKED ISSUE #950: [Feature]: let an MCP tool call name its own graph data directory
### Problem

One MCP server answering for several repositories cannot give each one its own graph data
directory.

Moving the database off the working tree is already supported, and I use it: the repos I work in
live on a Samba mount where SQLite access is unreliable, so every graph has to live under
`~/.crg-graphs/<repo>/`. Two mechanisms exist and both are process- or machine-wide:

| Mechanism | Scope | Why it does not cover this |
|---|---|---|
| `CRG_DATA_DIR` | whole process, one path used verbatim | eight repos would share one directory |
| registry entry | one repo, durable | only a CLI run writes one; an MCP-only client cannot |

So today the MCP server resolves the right directory only if someone has already run the CLI once
per repo to seed `~/.code-review-graph/registry.json`. That works on a developer laptop. It does
not work for a client that speaks only MCP, or in an ephemeral or read-only environment with
nowhere durable to keep the registry.

The practical symptom is smaller than it sounds and I want to state it plainly rather than
oversell it: the tools are usable, they just cannot be told where to look. Resolution depends on
ambient global state the call cannot see or set.

### Proposed solution

An optional `data_dir` on the graph-backed MCP tools, checked ahead of the registry and
`CRG_DATA_DIR`:

```jsonc
{ "repo_root": "/repos/api", "data_dir": "/var/graphs/api" }
{ "repo_root": "/repos/web", "data_dir": "/var/graphs/web" }
```

It would write no registry entry and affect only the call that passes it. Omitted, resolution stays
exactly as it is now, so nothing changes for existing clients.

Plus `serve --data-dir` as the default for calls that pass none, for the ordinary single-repository
server. Unlike `build --data-dir`, it would not write a registry entry.

This mirrors `repo_root` and `_resolve_repo_root` one for one, which is what makes it cheap to
review: the same precedence shape, an explicit argument beating a server flag beating the existing
fallback.

I have this working and will open a PR against `main` if you are open to the idea. It also covers a
small gap I found on the way: `detect-changes` is already in the CLI's `_data_dir_cmds` and already
resolves a read-only database path from `args.data_dir`, but its parser never defines `--data-dir`,
so that branch is currently unreachable.

### Affected area

MCP tools / server

### Alternatives considered

- **`CRG_DATA_DIR` in the MCP server env.** Correct for one repo per server, and what I would
  recommend to anyone in that situation. It is one path for the process, so it cannot express per
  repo directories.
- **Seed the registry from the CLI once per repo.** This is what I do today and it works. It makes
  a CLI run a prerequisite for correct MCP behaviour, and puts the answer in machine-global state
  rather than in the call.
- **One MCP server per repository.** Sidesteps it, at the cost of a server process per repo.
- **A `set_data_dir` tool.** Rejected: it is the same global mutable state, reachable from a tool
  call, which seems worse than an argument.

### Additional context

This is a  #155's follow-up, the `--data-dir` CLI work you merged in `e46341b`. This is the MCP half
of the same idea; I kept it local at the time and should have sent it then.

Unrelated, and I will file it separately if you would rather, but I hit it while running your suite:
`test_explicit_repo_without_markers_errors_cleanly` fails on a clean checkout for me. The upward
marker walk finds `~/.code-review-graph`, code-review-graph's own registry directory, and treats it
as a project root, so any temp directory under `$HOME` resolves instead of erroring.

felipecaldas:
Closes #950.

Adds an optional `data_dir` to the graph-backed MCP tools, naming where that call reads and writes
the graph database.

## Why

Moving the database off the working tree is already supported. Selecting it *per repository from
MCP* is not:

| Mechanism | Scope | Gap |
|---|---|---|
| `CRG_DATA_DIR` | whole process, one path used verbatim | several repos would share one directory |
| registry entry | one repo, durable | only a CLI run writes one; an MCP-only client cannot |

So the server resolves the right directory only when someone has already run the CLI once per repo
to seed `~/.code-review-graph/registry.json`. Fine on a laptop. Not fine for a client that speaks
only MCP, or an ephemeral or read-only environment with nowhere to keep the registry.

I am not claiming the tools are broken today. They are usable; they just cannot be told where to
look, and resolution depends on ambient global state the call cannot see or set.

## Shape

Mirrors `repo_root` one for one, so the precedence is the shape already in the file:

```
_resolve_data_dir(data_dir)   ->  explicit argument
                              ->  serve --data-dir  (_default_data_dir)
                              ->  None, and get_data_dir falls through unchanged:
                                    registry -> CRG_DATA_DIR -> <repo>/.code-review-graph
```

`data_dir` writes no registry entry and affects only the call that passes it.

```jsonc
{ "repo_root": "/repos/api", "data_dir": "/var/graphs/api" }
{ "repo_root": "/repos/web", "data_dir": "/var/graphs/web" }
```

```bash
code-review-graph serve --repo /repos/api --data-dir /var/graphs/api
```

## Backwards compatibility

Omitting `data_dir` leaves resolution byte-identical. Every parameter defaults to `None`, and a
test asserts that across the whole tool surface. The read-only guarantees from #777, #782 and #803
still hold: read-only resolution with an explicit directory creates nothing, and no command
materializes a `graph.db` for a repo that has no graph.

## Three tools deliberately do not take it

`get_docs_section_tool` serves documentation files and never opens the database. `list_repos_tool`
and `cross_repo_search_tool` span many registered repos, where one directory is meaningless. A test
pins that exception set so a future tool cannot quietly join it.

`apply_refactor_tool` accepts it but forwards it only to the provenance read: it rewrites source
files from a stored plan and never opens the database itself. mypy caught the over-eager version of
that.

## Also in here

`detect-changes --data-dir`. The command was already listed in the CLI's `_data_dir_cmds` and
already resolved a read-only database path from `args.data_dir`, but its parser never defined the
flag, so the branch was unreachable. Happy to split this into its own PR if you would rather keep
the feature clean.

## Verification

|  | before | after |
|---|---|---|
| `uv run pytest tests/` | 2984 passed, 1 failed | 3010 passed, 1 failed |
| `uv run ruff check code_review_graph/` | clean | clean |
| `uv run mypy code_review_graph/ --ignore-missing-imports --no-strict-optional` | clean | clean |

26 net new tests cover the precedence order, that a blank value falls through unchanged, that no
registry entry is written, that read-only resolution creates nothing, that two repos in one session
resolve separately, and that a call argument beats the `serve` flag.

The one failure is identical in both columns and is not from this branch:
`test_explicit_repo_without_markers_errors_cleanly` fails on a clean checkout on my machine because
the upward marker walk finds `~/.code-review-graph`, the tool's own registry directory, and treats
it as a project root. Noted at the end of #950; I left it alone here.

## Context

I wrote the `--data-dir` CLI work you merged in `e46341b`. This is the MCP half of the same idea,
which I kept local at the time.


tirth8205:
Outside this sweep's mission of demonstrated coding-session token savings and necessary correctness repairs: this introduces per-call storage routing across 27 MCP tool signatures and related CLI paths. Existing tools already work through the established storage configuration, and no selected failure requires this API expansion.

PR #952, opened 2026-09-01
fix: refresh graph freshness after no-op updates

anusbutt:
## Summary

Fixes a freshness inconsistency where successful incremental updates with zero structural graph changes could leave the stored Git SHA at the previous commit.

This created a contradiction where the update path reported the graph as up to date while `get_minimal_context()` still returned `stale_graph`.

## What changed

* Advance VCS freshness metadata after successful authoritative Git synchronization, even when zero graph mutations occur.
* Keep explicit `changed_files`, watch batches, and non-authoritative bases conservative.
* Do not advance freshness when discovery or parsing fails.
* Return an error instead of `"Graph is up to date"` for incomplete incremental updates.
* Add regression coverage for empty commits, zero-structural-change updates, explicit file lists, parse failures, and strict Git discovery.

## Validation

* `140 passed` for incremental + Git integration tests
* Full CI pytest: `2,990 passed, 9 skipped, 2 xpassed`
* Coverage: `85.02%`
* Ruff: passed
* mypy: passed
* Bandit: passed
* `git diff --check`: passed


tirth8205:
Integrated on `integration/token-efficiency-hardening` as `66309ec`. Authoritative no-op synchronisation now advances the stored Git anchor after empty or unsupported-file commits, avoiding repeated stale-graph reports. Explicit batches and parse or discovery failures preserve the previous anchor, and the corresponding incremental and Git regressions passed.

tirth8205:
Merged into `staging` through the integration PR #988 (merge commit 3030ca8). Your change is carried by commit 66309ec, which keeps you as the author (anusbutt). It will be promoted staging → testing → main and released from there. Thank you for the contribution! Closing this PR since its content is now on staging.

PR #955, opened 2026-09-03
fix(parser): extract Python aliased module imports

### LINKED ISSUE #920: [Bug]: Python `import X as Y` emits no IMPORTS edge — importers_of and impact silently miss aliased module imports
### code-review-graph version

2.3.8 (also present on `main` — see root cause below)

### Operating system

Windows

### Python version

3.14.4

### AI platform

claude-code

```shell
Nodes: 6
Edges: 12
Files: 5
Languages: python
Last updated: 2026-08-25T04:12:34
Built on branch: master
Built at commit: b196b575a2c9
```

### Steps to reproduce

Minimal repro — a fresh git repo with five files:

```python
# b.py
def hi():
    return 1
```
```python
# plain.py
import b
print(b.hi())
```
```python
# aliased.py
import b as B
print(B.hi())
```
```python
# fromimp.py
from b import hi
print(hi())
```
```python
# fromalias.py
from b import hi as H
print(H())
```

```bash
git init . && git add -A && git commit -m init
code-review-graph build
code-review-graph query importers_of b.py
```

**Expected:** 4 importers — `plain.py`, `aliased.py`, `fromimp.py`, `fromalias.py`.

**Actual:** 3 importers. `aliased.py` is missing.

```
found:  ['fromalias.py', 'fromimp.py', 'plain.py']
```

`import X as Y` produces **no IMPORTS edge at all**. Note the asymmetry: `from X import a as b` (`fromalias.py`) is handled correctly — only the aliased *module* import form is dropped.

This is not cosmetic for real repos. On a 373-file Python codebase where `import mod as ALIAS` is the house style for a shared helper, `importers_of` returned 6 of the 12 actual importers — **50% recall** — and every one of the 6 missing files used the `as` form (12/12 correlation between import form and whether the file was found). `impact --files <that file>` correspondingly under-reports the blast radius. The failure is silent: the result still reads `"summary": "Found 6 result(s)"`, so a caller with no independent grep has no signal that half the graph is missing.

Installing `code-review-graph[enrichment]` (jedi 0.20.0) and rebuilding does not change the result.

### Additional context

**Root cause** — `_extract_import()` in `code_review_graph/parser.py`, the Python branch (`main`, around line 15488):

```python
if language == "python":
    if node.type == "import_from_statement":
        for child in node.children:
            if child.type == "dotted_name":
                imports.append(child.text.decode("utf-8", errors="replace"))
                break
    else:
        for child in node.children:
            if child.type == "dotted_name":            # <-- direct children only
                imports.append(child.text.decode("utf-8", errors="replace"))
```

The `else` branch (i.e. `import_statement`) scans **direct children** for `dotted_name`. But tree-sitter-python nests the module name one level deeper when an alias is present:

```
import b            import_statement → dotted_name → identifier «b»
import b as B       import_statement → aliased_import → dotted_name → identifier «b»
                                                      → as → identifier «B»
import a.b as C     import_statement → aliased_import → dotted_name → identifier «a» . identifier «b»
```

So for `import b as B` the loop finds no `dotted_name` child, appends nothing, and no edge is emitted. `from X import a as b` works because that path resolves the module from the `import_from_statement`'s own `dotted_name`, which is a direct child there.

`_collect_import_names()` (same file, ~line 13394) has the mirrored gap: it handles `import_from_statement` only, so `import X as Y` also contributes nothing to `import_map` for call resolution.

**Suggested fix** — handle the `aliased_import` wrapper in the `import_statement` branch:

```python
    else:
        for child in node.children:
            if child.type == "dotted_name":
                imports.append(child.text.decode("utf-8", errors="replace"))
            elif child.type == "aliased_import":
                for sub in child.children:
                    if sub.type == "dotted_name":
                        imports.append(sub.text.decode("utf-8", errors="replace"))
                        break
```

This also covers `import a.b as C` and the multi-target form `import x, y as z`.

I applied exactly this patch to the installed 2.3.8 and rebuilt:

- the minimal repro above now returns all four importers, and also picks up `import a.b as C` and `import os, b as Z`;
- on the 373-file Python repo mentioned earlier, `importers_of` went from 6 to 12 results — matching the 12 files that actually import the module, with no false positives.

Suggested regression fixtures: `import b as B`, `import a.b as C`, `import x, y as z` — asserting an IMPORTS edge for each, since the current tests appear to cover only the plain and `from ... import` forms.

This looks like the same class of defect already fixed for other languages (#574 for PHP, #310 for C#, #65 for unresolved import paths) — the Python `as` form seems to have been missed.

zimo-xiao-zheng:
# Pull Request

## Linked issue

Closes #920

## What & why

Python `import X as Y` statements are represented as `aliased_import`
nodes by tree-sitter. `_extract_import()` only inspected direct
`dotted_name` children, so aliased module imports emitted no
`IMPORTS_FROM` edge. As a result, `importers_of` and impact analysis
silently missed those files.

This change extracts the original module name from `aliased_import`
nodes and adds regression coverage for:

- `import b as B`
- `import a.b as C`
- `import x, y as z`
- end-to-end `importers_of` and impact results

## How it was tested

```bash
uv run pytest tests/test_parser.py::TestCodeParser::test_parse_python_aliased_module_imports tests/test_parser.py::TestCodeParser::test_python_aliased_module_is_reported_by_queries -q --tb=short
# 2 passed

# All checks passed

git diff --check
# Passed
```

The full local Windows suite produced 2906 passed, 19 skipped, 2 xpassed,
and 71 failures. Every failing node ID was reproduced unchanged on an
untouched `origin/main` worktree, confirming this change adds no failures.

## Checklist

- [ ] All tests pass: `uv run pytest tests/ --tb=short -q` — see the reproduced baseline failures above
- [x] Docs updated where behavior changed — no documentation change is needed for this parser bug fix


tirth8205:
Integrated on `integration/token-efficiency-hardening` as `053bace`. Python aliased-module imports now contribute the missing IMPORTS_FROM edges used by importer and impact queries. Both new regressions fail on main and pass with the small extraction change; the targeted parser suite passed.

tirth8205:
Merged into `staging` through the integration PR #988 (merge commit 3030ca8). Your change is carried by commit 053bace, which keeps you as the author (zimo-xiao-zheng). It will be promoted staging → testing → main and released from there. Thank you for the contribution! Closing this PR since its content is now on staging.

PR #956, opened 2026-09-03
feat(parser): extract call-registration HTTP endpoints for JS/TS and Go

geeknik:
## What

Extract HTTP endpoints from **call-registration** web frameworks so their handlers become addressable like decorator/annotation routes:

- **JS/TS** — Express/Koa/Fastify `app.get('/x', handler)`, `router.post('/x', handler)`
- **Go** — net/http `http.HandleFunc('/x', handler)`, gin/echo/chi `r.GET('/x', handler)`

Each emits an `Endpoint` node and a `HANDLES` edge (handler → endpoint), mirroring the existing `_emit_spring_webflux_endpoint`. The existing `handlers_of` / `endpoints_for` queries work with **zero changes**, so Spring MVC, Spring WebFlux, Express, Koa, Fastify, and Go are all handled uniformly.

## Why

Decorator/annotation routes (FastAPI `@app.get`, Spring `@GetMapping`) are already captured, but call-registration routes were not, so their handlers were invisible as HTTP entrypoints. `flows.detect_entry_points` even carries `(app|router)\.(get|post|...)` patterns, but they only fire on a stamped decorator string, which call-registration never produces.

## How

- New `_call_registration_arguments` + `_emit_call_registration_endpoint` next to the WebFlux emitter.
- Wired into `_extract_calls` at the JS `member_call` branch and the Go typed-call branch.
- Constants `_JS_ROUTE_VERBS` / `_GO_ROUTE_VERBS`.

## Scope / follow-up

Named identifier handlers defined in the file. **Inline/anonymous** handlers (`app.get('/x', (req,res)=>{...})`) are a documented follow-up: they are not their own graph nodes today and need synthetic node creation in the same emitter.

## Safety

No schema change, no migration — `Endpoint`/`HANDLES` already exist and node/edge kinds are free-form TEXT. Detection requires a literal path starting with `/` plus an identifier handler, so non-route calls like `cache.get('key')` are ignored.

## Tests

Adds `tests/test_call_registration_endpoints.py` (Express, Go net/http, gin, `handlers_of`/`endpoints_for` round-trip, negative cases). Existing parser/graph/flows suites pass unchanged locally (774 + 4).


tirth8205:
Changes required: method names alone create false HTTP endpoints and inline handler identities collide. Parse `const cache = new Map(); const data = cache.get('/key', () => fallback());` using `CodeParser(repo_root).parse_file(path)` and this PR emits an HTTP endpoint; parsing `app.get('/a', () => first()); app.get('/b', () => second());` on one line assigns both routes the same handler identity and leaves body calls attached to the file. Require framework evidence, unique handler identities and correct call attribution.

PR #958, opened 2026-09-05
Made the window object which considers child and parent daemon jobs t…

### LINKED ISSUE #925: crg-daemon stop on Windows terminates the daemon without stopping its watchers, leaving orphans that duplicate on the next start
Surfaced while reviewing #918 (which is test-only and not the cause). Verified by reading the source; not reproduced on a Windows host.

## What happens

`crg-daemon stop` sends `os.kill(pid, signal.SIGTERM)` (`code_review_graph/daemon_cli.py:73`). On Windows, `os.kill` with any signal other than `CTRL_C_EVENT`/`CTRL_BREAK_EVENT` calls `TerminateProcess`, which kills the process unconditionally. Nothing in the daemon's shutdown path gets to run:

- `_handle_sigterm` (`daemon.py:1155`), `WatchDaemon.stop()` (`daemon.py:808`) and the `finally: daemon.stop()` in `_handle_start` (`daemon_cli.py:52`) are all unreachable;
- `_terminate_child()` (`daemon.py:1245`) is the only caller of `clear_watch_health()` on the stop path, so the health file is never cleared;
- `_start_watcher()` (`daemon.py:1229`) spawns watchers with a bare `subprocess.Popen` — no `creationflags`, no Job Object — so on Windows the children are **not** killed with the parent.

## Consequence

After `crg-daemon stop` on Windows the watcher children survive as orphans, still holding `graph.db` open and still writing to it. The CLI's own `finally` still calls `clear_pid()` (`daemon_cli.py:95`), so the next `crg-daemon start` sees no daemon and spawns a **second** watcher per repository on top of the orphans.

A related dead path: `_install_sigterm_interrupt()` (`incremental.py:2339`) cannot fire on Windows, because SIGTERM is never delivered from another process there. A watcher killed with Task Manager or `taskkill` therefore leaves a frozen health file that `read_watch_health` marks `stalled=True` indefinitely.

## Suggested fix

Use a Windows-appropriate stop: create the watcher processes in a Job Object (or with `CREATE_NEW_PROCESS_GROUP` plus `CTRL_BREAK_EVENT`) so the daemon can ask them to exit and so orphans die with the parent. Failing that, have `stop` reap known child PIDs explicitly and clear the health file itself before clearing the PID file.

Bakul2006:
# Pull Request

## Linked issue #925

Closes #925

### What

Fixes Windows daemon shutdown so that watcher processes cannot remain
orphaned when `crg-daemon stop` terminates the daemon before its Python
signal handler can run.

On Windows, watcher processes are now assigned to a standard-library
Windows Job Object configured with `JOB_OBJECT_LIMIT_KILL_ON_JOB_CLOSE`.
This ensures that if the daemon process is terminated unexpectedly, its
watcher processes are also terminated automatically.

The existing graceful shutdown path through `WatchDaemon.stop()` and
`_terminate_child()` remains unchanged for normal shutdowns and Unix
platforms.

The stop command was also updated to preserve the daemon PID file when
both graceful and forced termination attempts fail, preventing a future
`crg-daemon start` from incorrectly creating a duplicate daemon.

### Why

On Windows, sending `SIGTERM` with `os.kill()` can terminate the daemon
without executing its Python SIGTERM handler. This bypasses
`WatchDaemon.stop()`, leaving watcher processes running while the daemon
PID file is removed.

Those orphaned watchers can continue accessing `graph.db` and a subsequent
`crg-daemon start` can create duplicate watchers for the same repository.

The Job Object ties the watcher lifecycle to the daemon and prevents this
orphan-process scenario.

## How it was tested

Full test suite:

```bash

### DIFF
--- a/code_review_graph/daemon.py
+++ b/code_review_graph/daemon.py
@@ -107,6 +107,101 @@ _RESTART_BACKOFF_BASE = float(os.environ.get("CRG_RESTART_BACKOFF", "30"))
 _RESTART_BACKOFF_MAX = float(os.environ.get("CRG_RESTART_BACKOFF_MAX", "900"))
 _RESTART_HEALTHY_SECONDS = float(os.environ.get("CRG_RESTART_HEALTHY_AFTER", "600"))
 
+_PROCESS_TERMINATE = 0x0001
+_PROCESS_SET_QUOTA = 0x0100
+_JOB_OBJECT_EXTENDED_LIMIT_INFORMATION = 9
+_JOB_OBJECT_LIMIT_KILL_ON_JOB_CLOSE = 0x2000
+
+
+class _WindowsJob:
+    """Own watcher processes so Windows cannot orphan them on daemon exit."""
+
+    def __init__(self) -> None:
+        import ctypes
+        from ctypes import wintypes
+
+        self._ctypes = ctypes
+        self._kernel32 = ctypes.WinDLL("kernel32", use_last_error=True)
+        kernel32 = self._kernel32
+        kernel32.CreateJobObjectW.argtypes = (wintypes.LPVOID, wintypes.LPCWSTR)
+        kernel32.CreateJobObjectW.restype = wintypes.HANDLE
+        kernel32.SetInformationJobObject.argtypes = (
+            wintypes.HANDLE,
+            wintypes.DWORD,
+            wintypes.LPVOID,
+            wintypes.DWORD,
+        )
+        kernel32.SetInformationJobObject.restype = wintypes.BOOL
+        kernel32.OpenProcess.argtypes = (wintypes.DWORD, wintypes.BOOL, wintypes.DWORD)
+        kernel32.OpenProcess.restype = wintypes.HANDLE
+        kernel32.AssignProcessToJobObject.argtypes = (wintypes.HANDLE, wintypes.HANDLE)
+        kernel32.AssignProcessToJobObject.restype = wintypes.BOOL
+        kernel32.CloseHandle.argtypes = (wintypes.HANDLE,)
+        kernel32.CloseHandle.restype = wintypes.BOOL
+
+        class IoCounters(ctypes.Structure):
+            _fields_ = [(f"value_{index}", ctypes.c_ulonglong) for index in range(6)]
+
+        class BasicLimitInformation(ctypes.Structure):
+            _fields_ = [
+                ("per_process_user_time_limit", ctypes.c_longlong),
+                ("per_job_user_time_limit", ctypes.c_longlong),
+                ("limit_flags", wintypes.DWORD),
+                ("minimum_working_set_size", ctypes.c_size_t),
+                ("maximum_working_set_size", ctypes.c_size_t),
+                ("active_process_limit", wintypes.DWORD),
+                ("affinity", ctypes.c_size_t),
+                ("priority_class", wintypes.DWORD),
+                ("scheduling_class", wintypes.DWORD),
+            ]
+
+        class ExtendedLimitInformation(ctypes.Structure):
+            _fields_ = [
+                ("basic_limit_information", BasicLimitInformation),
+                ("io_info", IoCounters),
+                ("process_memory_limit", ctypes.c_size_t),
+                ("job_memory_limit", ctypes.c_size_t),
+                ("peak_process_memory_used", ctypes.c_size_t),
+                ("peak_job_memory_used", ctypes.c_size_t),
+            ]
+
+        self._handle = kernel32.CreateJobObjectW(None, None)
+        if not self._handle:
+            raise ctypes.WinError(ctypes.get_last_error())
+
+        limits = ExtendedLimitInformation()
+        limits.basic_limit_information.limit_flags = _JOB_OBJECT_LIMIT_KILL_ON_JOB_CLOSE
+        if not kernel32.SetInformationJobObject(
+            self._handle,
+            _JOB_OBJECT_EXTENDED_LIMIT_INFORMATION,
+            ctypes.byref(limits),
+            ctypes.sizeof(limits),
+        ):
+            error = ctypes.get_last_error()
+            kernel32.CloseHandle(self._handle)
+            raise ctypes.WinError(error)
+
+    def assign(self, proc: subprocess.Popen[bytes]) -> None:
+        """Assign *proc* to this job before it can outlive the daemon."""
+        process_handle = self._kernel32.OpenProcess(
+            _PROCESS_SET_QUOTA | _PROCESS_TERMINATE,
+            False,
+            proc.pid,
+        )
+        if not process_handle:
+            raise self._ctypes.WinError(self._ctypes.get_last_error())
+        try:
+            if not self._kernel32.AssignProcessToJobObject(self._handle, process_handle):
+                raise self._ctypes.WinError(self._ctypes.get_last_error())
+        finally:
+            self._kernel32.CloseHandle(process_handle)
+
+    def close(self) -> None:
+        """Release the job; kill-on-close handles abnormal daemon exits."""
+        if self._handle:
+            self._kernel32.CloseHandle(self._handle)
+            self._handle = None
+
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by tirth8205** [comment]
Blocked: the Windows Job Object lifecycle is unverified, current Windows/type checks fail, and successful signal delivery is still treated as proof that the daemon stopped. In an isolated unit test, simulate a recorded running daemon and reproduce the stop defect by making `os.kill` return successfully while `daemon.pid_alive` remains true, then invoke the daemon stop command; this PR clears the PID and reports success without establishing termination. Retain the PID until death is confirmed and verify parent termination and child cleanup with real Windows processes.

**Comment by Bakul2006** [comment]
> Blocked: the Windows Job Object lifecycle is unverified, current Windows/type checks fail, and successful signal delivery is still treated as proof that the daemon stopped. In an isolated unit test, simulate a recorded running daemon and reproduce the stop defect by making `os.kill` return successfully while `daemon.pid_alive` remains true, then invoke the daemon stop command; this PR clears the PID and reports success without establishing termination. Retain the PID until death is confirmed and verify parent termination and child cleanup with real Windows processes.

For sure  @tirth8205   I will take up the tests and update you shortly!!

**Comment by Bakul2006** [comment]
Yes @tirth8205 I just got your point and yet got to know that just os.kill() execution doesn't retrieve that daemon is stopped successfully correct me if I got wrong somewhere ..

**Comment by Bakul2006** [comment]
-> 135 daemon/watch/process tests passed, 2 skipped on Linux.
      mypy, ruff, and git diff --check passed.
->The Windows Job Object integration tests could not be run locally because the current environment is Linux; they will need       to be verified by Windows CI.

**Comment by tirth8205** [comment]
The portable stop correction from d651370 is integrated on `integration/token-efficiency-hardening` as `660b7ac`, with nine additional tests for exit checks, PID retention and restart refusal. The Windows Job Object portion of #925 remains blocked: native lifecycle verification is missing and the current Windows CI job fails. Its new child test generates invalid Python (`...; while ...`); add `compile(parent_code, "<parent>", "exec")` immediately before `Popen` in `test_windows_job_close_terminates_watcher_children` to reproduce the SyntaxError without launching a process.

**Comment by Bakul2006** [comment]
> The portable stop correction from [d651370](https://github.com/tirth8205/code-review-graph/commit/d651370fc098080de3335e7157e37bb7116d8ec0) is integrated on `integration/token-efficiency-hardening` as `660b7ac`, with nine additional tests for exit checks, PID retention and restart refusal. The Windows Job Object portion of #925 remains blocked: native lifecycle verification is missing and the current Windows CI job fails. Its new child test generates invalid Python (`...; while ...`); add `compile(parent_code, "<parent>", "exec")` immediately before `Popen` in `test_windows_job_close_terminates_watcher_children` to reproduce the SyntaxError without launching a process.

yeah @tirth8205  finally all the test passed including windows daemon!!

**Comment by tirth8205** [comment]

- tests/test_daemon.py (TestWatchDaemon, cross-platform, mock job): set daemon._windows_job = MagicMock() with job.assign.side_effect = OSError(...) and mock_proc.poll.return_value = None; assert mock_proc.terminate() and mock_proc.wait() were called, the alias is absent from daemon._children, and the...
- Same test family: make mock_proc.wait raise subprocess.TimeoutExpired on the first call so the proc.kill() branch is exercised; assert kill() was called.
- Variant: job.assign raises but mock_proc.poll() returns a non-None returncode (child died before assignment) -> assert terminate() NOT called and alias not tracked.
- tests/test_daemon.py: patch WatchDaemon._terminate_child with side_effect=RuntimeError, set daemon._windows_job = MagicMock(), call daemon.stop() inside pytest.raises; assert job.close.assert_called_once().
- Either guard the second proc.wait(timeout=5) after proc.kill() in _start_watcher (daemon.py ~1336) so TimeoutExpired cannot escape and leak log_fd / kill the health thread, and add a test for it , or document why it cannot time out.

**Comment by Bakul2006** [comment]
image
@tirth8205  facing some issues with mypy tests but yet passed all in the end!!

**Comment by tirth8205** [comment]
`f6d5bcf` does all five, including the fifth, which you fixed rather than documented around. Good call.

Gates on `c86fe2e` merged into `staging` 9c13f4c: `3341 passed, 11 skipped, 2 xpassed`, ruff clean on `code_review_graph/`, mypy clean. The five new tests all pass cross-platform here on macOS. The only two skips in that area are the pre-existing native Windows Job Object integration tests, and `Windows daemon and file handles` is green on CI at 9m48s, so the Windows job that was failing is no longer failing.

I mutation-checked the fifth item because it is the one that matters most. Reverting the try/except around the second `proc.wait(timeout=5)` in `_start_watcher` makes `test_start_watcher_assignment_failure_handles_kill_wait_timeout` fail with the `TimeoutExpired` escaping, and nothing else fails. That is a real guard, not a test that passes either way.

On the mypy trouble in your screenshot: mypy is clean here with the exact gate command, `uv run mypy code_review_graph/ --ignore-missing-imports --no-strict-optional`. If you saw errors, you were almost certainly pointing mypy at `tests/` as well, which is not part of the gate and is not type-clean. Nothing for you to fix.

One nit, not blocking. `tests/test_daemon.py:614` is 101 characters. CI only lints `code_review_graph/`, so it will not fail, but the repo convention is 100. Wrap it if you push again for any other reason.

Queued for merge.

Removing `needs-tests`.

**Comment by tirth8205** [comment]
The cleanup inside `except Exception:` in `_start_watcher` is itself unguarded, so on Windows one assign failure leaves the daemon permanently dead. `proc.terminate()` and `proc.kill()` are called bare, and `log_fd.close()` sits after them. TerminateProcess can fail with ERROR_ACCESS_DENIED against a process that is already exiting, and `Popen.terminate()` raises OSError when it does.

Repro on the head: set `daemon._windows_job` to a stub whose `assign()` raises OSError, hand `_start_watcher` a process object wrapping a real child whose `terminate()` raises PermissionError, then call it.

```
head 9abf853: escaped=PermissionError: Access is denied  child_still_alive=True  fd_leak=1
base ca23a73: escaped=None                               fd_leak=0
```

Three things break at once. The original assign error is swallowed and replaced, the log handle leaks, and the exception escapes `_start_watcher` into `_check_health`. `_health_loop` calls `_check_health` with no exception guard, so the health thread dies silently and no watcher is restarted again for the life of the daemon. Nothing covers this path: `grep -n 'terminate.side_effect\|kill.side_effect' tests/test_daemon.py` returns nothing, and only `wait` raising TimeoutExpired is exercised. About four lines fix it: `except OSError` around the terminate/kill block, and `log_fd` closed in a `finally`.

All five items from the last round are in the head and each one is individually mutation-covered. Four reverts were run, including dropping only the inner guard around the post-kill `proc.wait(timeout=5)`, and every one failed a test. The ctypes layout matches the Win32 x64 ABI exactly at 144/64/48 bytes with offsets 16/64/112, info class 9 and flag 0x2000 are right, and the job handle is non-inheritable so kill-on-close can fire. Delta against staging is exactly +6 passed and +2 skipped.

`tests/test_daemon.py:614` is still 101 characters. CI only lints `code_review_graph/` so it will not fail, but I asked for it last round.

Three Windows-only items, none blocking:

- `_initial_build` runs the build through `subprocess.run` and never assigns it to the job, so a daemon killed during a fresh-repo build orphans that process. Same in `reconcile()` for a newly added repo.
- `tests/test_watch_robustness.py` constructs `WatchDaemon` bare at lines 1346, 1396, 1427 and 1468 and never closes the job. That file runs in the windows-native CI job and `_WindowsJob` has no finalizer, so each one leaks a handle. The `daemon_env` fixture in `test_daemon.py` got the `_windows_job = None` treatment and these did not.
- Assignment happens after `Popen` returns, so a grandchild spawned in that window escapes the job for good. `subprocess` exposes no thread handle, so CREATE_SUSPENDED is not an option. The docstring should say the job almost always owns them rather than that Windows cannot orphan them.

Mergeable once the terminate/kill block is guarded with a test that sets `terminate.side_effect`, and line 614 is wrapped.

PR #959, opened 2026-09-05
fix(watch): drop events on paths the OS cannot stat

### LINKED ISSUE #897: watch: an event on a path the OS cannot stat (name too long) kills the watch loop
Found while reviewing #893. Pre-existing on `main`, unrelated to that branch.

## What happens

`WatchBatchProcessor._relative_path` (`code_review_graph/incremental.py:2143`) walks up to the nearest existing ancestor:

```python
existing = candidate
while not existing.exists() and existing != lexical_root:
    existing = existing.parent
```

`Path.exists()` propagates `OSError` for a path the OS cannot even stat. A component longer than `NAME_MAX` is the easy case:

```
>>> Path('/tmp/' + 'a'*300 + '/x.py').exists()
OSError: [Errno 63] File name too long
```

That escapes `_relative_path`, is caught by `process()`'s `except BaseException`, sets `self.failure`, and `raise_if_failed()` ends the watch loop with `RuntimeError: watch update failed`. The watcher exits; under the daemon it is restarted (with backoff), pays for a full initial update, and dies again the next time the same event arrives.

## Why it matters

Any tool that writes a too-long name inside a watched repo — a build cache, a generated fixture, a test that constructs deep paths — takes the watcher down. `collect_all_files` already guards for exactly this at `incremental.py:1054-1060` ("Skip paths that would exceed OS filename limits"), so the concern is acknowledged elsewhere in the codebase; the watch path just doesn't have it.

## Suggested fix

Guard the `exists()` walk (and the `realpath`/`resolve` calls near it) so an unstattable path is dropped like any other irrelevant event rather than killing the loop. Reuse the length check from `collect_all_files` if that reads better than catching `OSError`.

## Reproduction

macOS, Python 3.13. Any watched repo; write a file whose name exceeds 255 bytes in a watched directory and the loop exits with `watch update failed`.

Chang-Jin-Lee:
# Pull Request

## Linked issue

Closes #897

## What & why

`WatchBatchProcessor._relative_path` walks up to the nearest existing ancestor with a bare `Path.exists()`. That call answers `False` for a path that is merely absent, but it *raises* `OSError` for a path the OS cannot stat at all. A component past `NAME_MAX` is the easy case:

```
>>> Path('/tmp/' + 'a'*300 + '/x.py').exists()
OSError: [Errno 36] File name too long
```

The exception escaped `_relative_path`, was caught by `process()`'s `except BaseException` into `self.failure`, and the next `raise_if_failed()` ended the watch loop with `RuntimeError: watch update failed`. Under the daemon the watcher then restarted with backoff, paid for a full initial update, and died again the next time the same event arrived — so any tool writing a too-long name inside a watched repo takes the watcher down repeatedly.

`collect_all_files` already skips paths past the OS limits, so the concern is acknowledged elsewhere in this file; the watch path just did not carry the guard.

The `exists()` walk, the `resolve()` call and the symlink scan now sit in one guarded block. `OSError` drops the event the way an ignored path is dropped and logs at debug, matching the sibling guard in `collect_all_files`. The pre-existing `ValueError` branch keeps its meaning, that a path resolving outside the root is not ours; widening it over the walk costs nothing, since a path Windows rejects as malformed should be dropped for the same reason.

I took the `OSError` route rather than reusing the length check because it also covers the cases a length check cannot see — a broken mount, a parent that has become unreadable — and every one of them wants the same answer.

The two tests are the reproduction from the issue. The first asserts the loop survives the event; the second puts a real file in the same batch and asserts it is still indexed, so the guard drops the bad event rather than the whole batch.

## How it was tested

```bash
$ uv run pytest tests/test_watch_robustness.py::TestUnstattablePaths -q
2 passed

$ uv run pytest tests/test_watch_robustness.py -q
58 passed

$ uv run pytest tests/ --tb=short -q
2987 passed, 9 skipped, 2 xpassed in 35.00s

All checks passed!

$ uv run mypy code_review_graph/incremental.py --ignore-missing-imports --no-strict-optional
Success: no issues found in 1 source file
```

Both new tests fail on `main` with `RuntimeError: watch update failed`, which is the symptom in the issue.

`ruff format` reports `incremental.py` and `test_watch_robustness.py` as unformatted, but it reports the same on unmodified `main`, so I left formatting alone rather than bury this change in an unrelated reflow.

Verified on Linux (Python 3.12). The issue reports macOS/3.13; the failing call is `Path.exists()` raising `ENAMETOOLONG`, which is POSIX behaviour rather than platform-specific, and the repro reproduces here unchanged.

One note in case it saves someone else the detour: I first ran the suite with `TMPDIR` on a tmpfs path that happened to contain the substring `test`, and `test_parser.py::TestCodeParser::test_non_test_file_describe_not_special` failed because the temp file it writes then sits under a test-looking path. Nothing to do with this change, but the test is sensitive to where `TMPDIR` points.

## Checklist

`docs/TROUBLESHOOTING.md` describes watcher stalls in terms of a dead observer thread; this failure ended the loop through `raise_if_failed` instead, and now does not happen at all, so I did not find a line there that became wrong. Happy to add one if you would rather it be named.


tirth8205:
Integrated on `integration/token-efficiency-hardening` as `33a96b4`. Unstattable watch events are contained to the offending path, allowing a valid source event in the same batch to update the graph. Both path/batch regressions fail on main and pass with the change; the native FSEvents whole-module run remained unavailable in the sandbox.

Chang-Jin-Lee:
Thanks for taking it.

On the FSEvents gap: the guard sits in `_relative_path`, which every backend feeds after the fact, so FSEvents changes which paths arrive rather than whether an unstattable one is dropped. What stays untested there is the batching shape — a whole-module event whose path stats fine is a different question from this one, and nothing here claims to answer it.

tirth8205:
Merged into `staging` through the integration PR #988 (merge commit 3030ca8). Your change is carried by commit 33a96b4, which keeps you as the author (Chang-Jin-Lee). It will be promoted staging → testing → main and released from there. Thank you for the contribution! Closing this PR since its content is now on staging.

Issue #943, opened 2026-08-30
Design: separate base-type extraction, canonical declaration identity, and resolution for INHERITS edges


merlincat11:
Raised out of #939, which is parked in favour of doing this deliberately. #935 is the motivating bug and should be a test case for whatever design lands.

## Problem

`INHERITS.target_qualified` currently serves three incompatible roles at once:

| role | example |
| --- | --- |
| raw source reference | `Models.GenericBase<string>` |
| canonical declaration reference | `Models.GenericBase` + arity |
| resolved graph target | `/repo/GenericBase.cs::GenericBase` |

#935 is a symptom: a generic base is stored as `GenericBase<string>`, which never matches
the node `GenericBase`, so `inheritors_of` returns zero and asserts the absence.

#939 attempted to fix this by normalizing the stored target to the erased declaration
name. That makes a **shared** field lossy, and the field has more than one consumer. Two
new false positives resulted, both verified against `main`.

## Verified regressions from the erase-in-place approach

These are the constraints any design must satisfy. Both were reproduced by building the
same source with `main` and with the #939 branch.

**1. Spring resolution picks the wrong implementation.**

```java
public interface Store<T> { T get(); }

@Component
public class StringStore implements Store<String> { public String get() {...} }

@Service
public class Consumer {
    @Autowired Store<Integer> store;
    public void run() { store.get(); }
}
```

`spring_resolver.py` builds its implementation map directly from `INHERITS.target_qualified`
and resolves when a key has exactly one implementor. INJECTS already reduces
`Store<Integer>` to `Store`, so once INHERITS also erases to `Store` the two meet:

```
main : CALLS Consumer.run -> Store.get          (conservative)
#939 : CALLS Consumer.run -> StringStore.get    (wrong implementation)
```

This rewrites a CALLS edge, so the error propagates into impact radius, flows and review
context. Generic parameters are deliberately autowiring qualifiers in Spring, so
`Store<String>` and `Store<Integer>` are distinct candidates.

**2. C# generic arity is part of declaration identity.**

```csharp
// I0.cs
public interface I { }
// I1.cs
public interface I<T> { }
// Child.cs
public class Child : I<int> { }
```

Querying the **non-generic** `I0.cs::I`:

```
main : 0 results
#939 : ['Child']        (Child implements I<int>, not I)
```

`inheritors_of` falls back from the qualified node to `node.name` filtered only by
language, and `extra.constructed_types` is not consulted on that path. This is within one
[...]

PR #964, opened 2026-09-08
fix(csharp): preserve namespace identity and lexical call binding

### LINKED ISSUE #946: [Bug]: C# namespaces are absent from the graph identity, making namespace-sensitive call resolution undecidable
### Summary

C# namespaces are not represented in the graph's C# identity. `parent_name` holds only the containing-*type* path, and the only namespace evidence is `csharp_namespaces_by_file` — a set of every namespace a file declares. The resolver therefore cannot establish which namespace a given type actually sits in, which makes namespace-sensitive call resolution undecidable in both directions: it fabricates edges for types the caller cannot name, and it cannot safely reject them without also rejecting valid calls.

This is long-standing behaviour, not a regression. It became visible during review of #937 (nested type identity, #934), where four consecutive rounds each closed one instance of it. It should be fixed at the cause.

### Why the current evidence is insufficient

Given:

```csharp
// Indexed.cs
namespace Other;

public class App
{
    public class Report
    {
        public class ExportHandler
        {
            public void Run() {}
        }
    }
}
```

the method is stored as:

```
parent_name = App.Report.ExportHandler
file        = Indexed.cs
namespaces  = {"Other"}          # per file, not per node
```

`Other` has been erased from the key. A caller meaning an entirely different `App.Report.ExportHandler` — typically from a referenced assembly that was never indexed — matches this node exactly.

### Failing cases

All of these reproduce on `main`. Each was verified during #937 review.

1. **Exact path, wrong namespace.** Receiver `App.Report.ExportHandler`, candidate above. `main` resolves it (on the bare name `ExportHandler`); #937 narrows it to requiring the whole containing path to coincide, but still resolves it. Should be unresolved.

2. **Same file, different namespaces.** File co-location does not imply visibility in C#; a type in `namespace Other` is not nameable from `namespace ConsumerNamespace` in the same file without a `using`. Currently resolves.

3. **Multi-namespace file.** A file declaring both `App` and `Other` cannot show which namespace encloses a candidate. #937 handles this conservatively — it requires the dropped namespace to be the file's *only* namespace — so genuine matches in such files are left unresolved. A false negative, chosen over a fabricated edge.

4. **`using` scoped to a namespace body.** `imports_by_file` collapses every `IMPORTS_FROM` edge in a file into one set, but the spec scopes an ordinary `using` to its immediately containing compilation unit or namespace body. A `using Other;` inside `namespace A` is currently indistinguishable from one that applies to `namespace B` in the same file.

5. **Project-wide `global using`.** A `global using Other;` in a separate file applies to every compilation unit. Any visibility check keyed on the caller's own file will wrongly reject these.

Cases 1, 2 and 4 over-permit; 3 and 5 under-permit. They share one cause.

### Prior attempt

`_csharp_namespace_visible()` (2fb7b6d, reverted in 01371fe) tried to decide visibility from file-level evidence: same file, global namespace, a `using` in the caller's file, or a namespace the caller declares. It simultaneously

- **regressed case 5** — a valid `global using` call resolved on `main` and became unresolved, and
- **failed to fix cases 1, 2 and 4** — still over-permitting.

Being wrong in both directions at once is the evidence that the granularity, not the rule, is the problem. It was reverted rather than extended.

### Proposed direction

Persist the declaring namespace per type/method rather than per file, and resolve against `(namespace, containing-type path)` instead of the containing-type path alone.

Open design questions worth settling before implementation:

- **Identity vs. sidecar.** Whether the namespace joins `qualified_name` (a C# identity-format change, so it needs the `CSHARP_IDENTITY_VERSION` rebuild gate that #937 already uses) or lands as a separate indexed column resolved against. The sidecar is cheaper and avoids re-spelling every C# node; the identity change makes wrong-namespace collisions impossible by construction.
- **Scoped usings.** `IMPORTS_FROM` needs to record the namespace body a `using` applies to, otherwise case 4 stays undecidable.
- **Global usings.** These need to be collected project-wide and unioned into every compilation unit's visible set, otherwise fixing cases 1–4 will regress case 5 exactly as the reverted attempt did.
- **Conservative default.** When evidence is still insufficient, leave the call unresolved rather than guess — consistent with the resolver's existing policy.

### Existing coverage

Two tests on the #937 branch already pin the boundary and should be honoured by any fix:

- `test_exact_path_match_should_check_the_candidate_namespace` — currently `xfail`, documents case 1, flips to passing when this is fixed
- `test_global_using_target_still_resolves` — guards case 5 against a repeat of the reverted approach

### Context

- #934 — nested type identity (the bug that surfaced this)
- #937 — nested type identity fix; held pending this work

merlincat11:
C# declarations in different namespaces can collapse into the same graph node, and call resolution can select types the caller cannot name. This change stores the namespace and complete containing-type path in C# qualified names and resolves calls using the declaration and lexical using scope.

Fixes #946. Includes the nested-type identity needed by #934/#937; it does not change inheritance target spelling or implement #943.

- Preserve namespace-body identity for ordinary usings, including reopened and file-scoped namespaces; collect explicit global usings across files in the same inferred project.
- Retain raw call references, re-evaluate bindings on incremental updates, and keep `TESTED_BY` mirrors aligned when bindings change or declarations are deleted. Persist flow invalidation when a binding changes, and reuse full tracing so unchanged callers and old/new entry points stay consistent, including after deferred postprocessing.
- Preserve each call's lexical containing type and byte offset, including initializers and same-line receivers; bind nullable receivers while retaining their raw annotations.
- Resolve `Alias::Type` through namespace aliases only, and allow unqualified calls to enclosing static methods while respecting nearer method groups and recorded callable shadowing. `this.Run()` stays within the immediate type.
- Reparse legacy C# identities through a version gate, retrying failed files individually instead of repeating a full rebuild. No new dependency or schema migration.

Project ownership uses the nearest unambiguous `.csproj`; loose C# files share the review root. MSBuild compile-item evaluation, generated/implicit usings, generic binding, overload selection, and inherited-member resolution remain outside this change. The data model and boundaries are documented in `docs/architecture.md`.

Validation: `uv run pytest --tb=short -q` — 3,048 passed, 9 skipped, 2 xpassed. Ruff, the CI mypy command, and wheel build passed. The 63 C# namespace tests and three real-observer watcher tests also pass in an isolated Python 3.11 environment.

A temporary 2,000-caller smoke check changed only a global alias: all callers rebound in 1.085 seconds, the full flow refresh took 0.104 seconds, and all 2,001 stored flows matched a fresh trace. A repeated binding pass made no changes. These are local smoke timings, not a performance guarantee.

All 13 GitHub checks passed for the `ced4f45` code state, including Python 3.10–3.13 and Windows daemon/file-handle tests; `b7e713d` adds only the changelog entry.

`CHANGELOG.md` records the change under `[Unreleased]`. Graphs that carry embeddings should re-run `code-review-graph embed` after the identity upgrade: the renamed C# nodes leave their previous vectors orphaned until a refresh purges them.

Two known gaps are left as follow-ups rather than widened into this change. Same-named types declared in several projects stay unresolved even where the caller's own `.csproj` owns exactly one of them, so a repository holding many copies of one namespace binds fewer calls than before. And a file whose grammar parse fails yields no type node, which now leaves it unbindable in both directions where bare-name matching previously guessed.

Local measurement on two real corpora, not a guarantee: serilog, 216 files, 1389 bound C# calls before and 1352 after; dotnet/samples, 3042 files, 7110 before and 4572 after. In a sample of the differences most were references this change declines to guess — `XElement.Load`, `Stopwatch.GetTimestamp` and `((ICollection)x).CopyTo` each previously bound to a same-named declaration in the repository — alongside the two gaps above. A 1500-file and a 6000-file synthetic C# repository showed single-file `update` at 1.5s and 3.2s against 1.8s and 3.0s on `main`, with the full build faster in both.

### DIFF
--- a/code_review_graph/csharp_resolver.py
+++ b/code_review_graph/csharp_resolver.py
@@ -36,6 +36,69 @@ def _join(parent: str, name: str) -> str:
     return f"{parent}.{name}" if parent else name
 
 
+_CLOSING = {"<": ">", "(": ")", "[": "]"}
+
+
+def _split(text: str, separator: str) -> list[str] | None:
+    """Split on *separator* at nesting depth zero, or ``None`` if unbalanced."""
+    parts: list[str] = []
+    depth: list[str] = []
+    current = ""
+    for character in text:
+        if character in _CLOSING:
+            depth.append(_CLOSING[character])
+        elif depth and character == depth[-1]:
+            depth.pop()
+        elif character == separator and not depth:
+            parts.append(current)
+            current = ""
+            continue
+        current += character
+    if depth:
+        return None
+    parts.append(current)
+    return parts
+
+
+def _arity(arguments: str) -> int | None:
+    """Return how many type arguments *arguments* supplies, ``<`` excluded."""
+    if not arguments.endswith(">"):
+        return None
+    supplied = _split(arguments[:-1], ",")
+    if supplied is None or not all(argument.strip() for argument in supplied):
+        return None
+    return len(supplied)
+
+
+def _type_key(reference: str) -> str | None:
+    """Return the declaration key a type reference names, or ``None``.
+
+    Generic declarations are keyed by arity — ``App.Box`1`` — because that is
+    what distinguishes ``Box`` from ``Box<T>``. Erasing the arguments instead
+    would let a reference select a declaration it does not name.
+    """
+    segments = _split(reference, ".")
+    if segments is None:
+        return None
+    keys = []
+    for index, segment in enumerate(segments):
+        name, separator, arguments = segment.partition("<")
+        if not separator:
+            if not segment.isidentifier():
+                return None
+            keys.append(segment)
+            continue
+        # A constructed containing type carries its own arity, which one key
+        # cannot describe; ``Outer<T>.Inner`` stays unresolved.
+        if index != len(segments) - 1 or not name.isidentifier():
+            return None
+        arity = _arity(arguments)
+        if arity is None:
+            return None
+        keys.append(f"{name}`{arity}")
+    return ".".join(keys)
+
+
 def resolve_csharp_calls(store: GraphStore, repo_root: Path | None = None) -> dict:
     conn = store._conn
     nodes = conn.execute("SELECT * FROM nodes WHERE language = 'csharp'").fetchall()
@@ -79,8 +142,12 @@ def resolve_csharp_calls(store: GraphStore, repo_root: Path | None = None) -> di
         qualified = node["qualified_name"]
         parent = node["parent_name"] or ""
         parents[qualified] = parent
-        if node["kind"] == "Class" and not extra.get("csharp_generic"):
-            types.setdefault(_join(parent, node["name"]), []).append(qualified)
+        if node["kind"] == "Class":
+            arity = extra.get("csharp_arity")
+            name = node["name"]
+            if isinstance(arity, int) and arity > 0:
+                name = f"{name}`{arity}"
+            types.setdefault(_join(parent, name), []).append(qualified)
             type_files[qualified] = node["file_path"]
             if extra.get("csharp_partial"):
                 partial_types.add(qualified)
@@ -136,7 +203,8 @@ def resolve_csharp_calls(store: GraphStore, repo_root: Path | None = None) -> di
         raw = raw.removeprefix("global::").removesuffix("?")
         alias, separator, reference = raw.partition("::")
         if separator:
-            if not alias.isidentifier() or not all(p.isidentifier() for p in reference.split(".")):
+            reference_key = _type_key(reference)
+            if not alias.isidentifier() or reference_key is None:
                 return []
             # Unlike '.', '::' searches only namespace aliases, even when a
             # local, type, or namespace has the same name as the qualifier.
@@ -147,12 +215,14 @@ def resolve_csharp_calls(store: GraphStore, repo_root: Path | None = None) -> di
                 if len(aliases) != 1:
                     return []
                 target = import_target(*aliases[0])
-                return types.get(_join(target, reference), []) if target in namespaces else []
+                return types.get(_join(target, reference_key), []) if target in namespaces else []
             return []
-        # Keep generic arguments lossless; erasure could select a different
-        # declaration (I vs I<T>). Generic binding belongs to #943.
-        if not all(part.isidentifier() for part in raw.split(".")):
+        # Generic arguments are kept, not erased: the reference resolves by
+        # arity, so ``I<int>`` reaches ``I<T>`` and never the separate ``I``.
+        key = _type_key(raw)
+        if key is None:
             return []
+        raw = key
         first, _, tail = raw.partition(".")
         if absolute:
             return types.get(raw, [])
--- a/code_review_graph/csharp_resolver.py
+++ b/code_review_graph/csharp_resolver.py
@@ -1,0 +1,284 @@
+"""Bind C# calls using declaration identity and lexical namespace evidence.
+
+Raw references survive binding: every pass can revoke a previously inferred
+edge when declarations or usings change. This is a structural binder, not a
+C# compiler; unsupported type expressions and ambiguous declarations stay raw.
+"""
+
+from __future__ import annotations
+
+import json
+import os
+from functools import lru_cache
+from pathlib import Path
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from .graph import GraphStore
+
+
+def _extra(raw: str | None) -> dict:
+    try:
+        value = json.loads(raw or "{}")
+    except (ValueError, TypeError):
+        return {}
+    return value if isinstance(value, dict) else {}
+
+
+def _parents(name: str):
+    while name:
+        yield name
+        name = name.rpartition(".")[0]
+    yield ""
+
+
+def _join(parent: str, name: str) -> str:
+    return f"{parent}.{name}" if parent else name
+
+
+def resolve_csharp_calls(store: GraphStore, repo_root: Path | None = None) -> dict:
+    conn = store._conn
+    nodes = conn.execute("SELECT * FROM nodes WHERE language = 'csharp'").fetchall()
+    files = {n["file_path"] for n in nodes}
+    if not files:
+        return {"files_indexed": 0, "calls_resolved": 0}
+    root = repo_root or Path(os.path.commonpath([str(Path(f).parent) for f in files]))
+
+    @lru_cache(maxsize=None)
+    def project(directory: Path) -> str | None:
+        try:
+            projects = list(directory.glob("*.csproj"))
+        except OSError:
+            return None
+        if projects:
+            return str(projects[0]) if len(projects) == 1 else None
+        if directory == root or directory.parent == directory:
+            # ponytail: loose .cs files share the review root; MSBuild compile
+            # item evaluation is needed for linked/conditional source ownership.
+            return str(root)
+        return project(directory.parent)
+
+    projects = {f: project(Path(f).parent) for f in files}
+    types: dict[str, list[str]] = {}
+    methods: dict[tuple[str, str], list[str]] = {}
+    parents: dict[str, str] = {}
+    partial_types: set[str] = set()
+    type_files: dict[str, str] = {}
+    namespaces = {""}
+    for node in nodes:
+        extra = _extra(node["extra"])
+        namespace = extra.get("csharp_namespace")
+        if node["kind"] == "File":
+            for name in extra.get("csharp_namespaces", []):
+                namespaces.update(_parents(name))
+        if not isinstance(namespace, str):
+            continue  # Legacy nodes do not establish namespace identity.
+        namespaces.update(_parents(namespace))
+        qualified = node["qualified_name"]
+        parent = node["parent_name"] or ""
+        parents[qualified] = parent
+        if node["kind"] == "Class" and not extra.get("csharp_generic"):
+            types.setdefault(_join(parent, node["name"]), []).append(qualified)
+            type_files[qualified] = node["file_path"]
+            if extra.get("csharp_partial"):
+                partial_types.add(qualified)
+        elif node["kind"] in ("Function", "Method", "Test"):
+            method_key = (f"{node['file_path']}::{parent}", node["name"])
+            methods.setdefault(method_key, []).append(qualified)
+
+    imports: dict[tuple[str, int], list[tuple[str, dict]]] = {}
+    global_imports: dict[str, list[tuple[str, dict]]] = {}
+    csharp_files_sql = "SELECT file_path FROM nodes WHERE kind = 'File' AND language = 'csharp'"
+    for row in conn.execute(
+        f"SELECT * FROM edges WHERE kind = 'IMPORTS_FROM' AND file_path IN ({csharp_files_sql})",
+    ):
+        extra = _extra(row["extra"])
+        scopes = extra.get("csharp_scopes")
+        if row["file_path"] not in files or not scopes:
+            continue
+        directive = (row["target_qualified"], extra)
+        scope = scopes[0][1]
+        if extra.get("csharp_global") and scope == -1:
+            # Ambiguous project ownership never exports a using to other files.
+            key = projects[row["file_path"]] or row["file_path"]
+            global_imports.setdefault(key, []).append(directive)
+        else:
+            imports.setdefault((row["file_path"], scope), []).append(directive)
+
+    def import_target(raw: str, extra: dict) -> str | None:
+        absolute = raw.startswith("global::")
+        raw = raw.removeprefix("global::")
+        if not all(part.isidentifier() for part in raw.split(".")):
+            return None
+        first = raw.split(".", 1)[0]
+        namespace = extra["csharp_scopes"][0][0]
+        for parent in ([""] if absolute else _parents(namespace)):
+            prefix = _join(parent, first)
+            if prefix in namespaces or prefix in types:
+                target = _join(parent, raw)
+                return target if target in namespaces or target in types else None
+        return None
+
+    def directives(file: str, scope: int) -> list[tuple[str, dict]]:
+        local = imports.get((file, scope), [])
+        if scope == -1:
+            return local + global_imports.get(projects[file] or file, [])
+        return local
+
+    def type_targets(raw: str, owner: str, scopes: list, file: str) -> list[str]:
+        absolute = raw.startswith("global::")
+        raw = raw.removeprefix("global::")
+        # Keep generic arguments lossless; erasure could select a different
+        # declaration (I vs I<T>). Generic binding belongs to #943.
+        if not all(part.isidentifier() for part in raw.split(".")):
+            return []
+        first, _, tail = raw.partition(".")
+        if absolute:
+            return types.get(raw, [])
+        namespace = scopes[0][0]
+        for enclosing in _parents(owner):
+            if enclosing == namespace:
+                break
+            if _join(enclosing, first) in types:
+                return types.get(_join(enclosing, raw), [])
+        scope_by_namespace = {name: scope for name, scope in scopes}
+        for enclosing in _parents(namespace):
+            visible = (
+                directives(file, scope_by_namespace[enclosing])
+                if enclosing in scope_by_namespace else []
+            )
+            aliases = [
+                (target, extra) for target, extra in visible
+                if extra.get("csharp_alias") == first
+            ]
+            prefix = _join(enclosing, first)
+            if prefix in namespaces or prefix in types:
+                return [] if aliases else types.get(_join(enclosing, raw), [])
+            if aliases:
+                if len(aliases) != 1:
+                    return []
+                alias = import_target(*aliases[0])
+                return types.get(_join(alias, tail) if tail else alias, []) if alias else []
+            imported = set()
+            for target, extra in visible:
+                kind = extra.get("csharp_using_kind")
+                if kind not in ("namespace", "static"):
+                    continue
+                imported_scope = import_target(target, extra)
+                if imported_scope is None:
+                    continue
+                # Using a namespace imports its types, never child namespaces.
+                if kind == "namespace" and imported_scope not in namespaces:
+                    continue
+                if kind == "static" and imported_scope not in types:
+                    continue
+                candidate = _join(imported_scope, first)
+                if candidate in types:
+                    imported.add(candidate)
+            if imported:
+                if len(imported) != 1:
+                    return []
+                prefix = next(iter(imported))
+                return types.get(_join(prefix, tail) if tail else prefix, [])
+        return []
+
+    def targets(row, extra: dict) -> list[str]:
+        scopes = extra.get("csharp_scopes")
+        if not scopes:
+            return []
+        owner = parents.get(row["source_qualified"], "")
+        kind = extra.get("csharp_call_kind")
+        method = extra["csharp_raw_target"].rsplit("::", 1)[-1]
+        receiver = extra.get("receiver_scope", "")
+        if extra.get("receiver") == "this" or kind == "unqualified":
+            own_type = f"{row['file_path']}::{owner}"
+            candidates = [own_type] if own_type in parents else []
+            if own_type in partial_types:
+                candidates = [
+                    c for c in types.get(owner, []) if c in partial_types
+                    and projects[type_files[c]] == projects[row["file_path"]]
+                ]
+        else:
+            candidates = type_targets(receiver, owner, scopes, row["file_path"])
+        if not candidates:
+            return []
+        if len(candidates) > 1:
+            owners = {projects[type_files[c]] for c in candidates}
+            if (
+                not all(c in partial_types for c in candidates)
+                or len(owners) != 1 or None in owners
+            ):
+                return []
+        if kind == "constructor":
+            return candidates
+        return [method_qn for c in candidates for method_qn in methods.get((c, method), [])]
+
+    resolved = 0
+    changed = False
+    for row in conn.execute(
+        f"SELECT * FROM edges WHERE kind = 'CALLS' AND file_path IN ({csharp_files_sql})",
+    ).fetchall():
+        extra = _extra(row["extra"])
+        raw = extra.get("csharp_raw_target")
+        if row["file_path"] not in files or not isinstance(raw, str):
+            continue
+        candidates = targets(row, extra)
+        target = candidates[0] if len(candidates) == 1 else raw
+        for key in ("unresolved_targets", "scoped_resolved", "scoped_via"):
+            extra.pop(key, None)
+        if len(candidates) == 1:
+            extra.update(scoped_resolved=True, scoped_via="csharp_namespace")
+        else:
+            extra["unresolved_targets"] = []
+        if _set_call(store, row, target, extra):
+            resolved += len(candidates) == 1
+            changed = True
+    if changed:
+        conn.commit()
+        store._invalidate_cache()
+    return {"files_indexed": len(files), "calls_resolved": resolved}
+
+
+def _set_call(store: GraphStore, row, target: str, extra: dict) -> bool:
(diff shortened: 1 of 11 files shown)


code_review_graph/csharp_resolver.py
```diff
@@ -0,0 +1,284 @@
...
+                    return []
+                prefix = next(iter(imported))
+                return types.get(_join(prefix, tail) if tail else prefix, [])
+        return []
+
+    def targets(row, extra: dict) -> list[str]:
+        scopes = extra.get("csharp_scopes")
+        if not scopes:
+            return []
+        owner = parents.get(row["source_qualified"], "")
```
merlincat11: **[P2] Retain the lexical containing type for initializer calls**

C# field initializers are emitted with the file as `source_qualified`, so this lookup makes `owner` empty even inside a namespaced class. `type_targets()` then tests the global namespace in its enclosing-type loop before checking the call's actual namespace.

Reproduce with these three files:

```csharp
// Global.cs
public class Service { public static int Run() => 1; }
// App.cs
namespace App; public class Service { public static int Run() => 2; }
// Caller.cs
namespace App; class C { static int Value = Service.Run(); }
```

`full_build` on this PR binds the initializer to `Global.cs::Service.Run`; `main` binds it to `App.cs::Service.Run`, which is the declaration visible in `App`. This creates an incorrect cross-namespace call edge. Preserve the call's lexical containing-type context independently of its graph caller identity, and cover a namespaced initializer in the regression tests.

code_review_graph/csharp_resolver.py
```diff
@@ -0,0 +1,284 @@
...
+            return local + global_imports.get(projects[file] or file, [])
+        return local
+
+    def type_targets(raw: str, owner: str, scopes: list, file: str) -> list[str]:
+        absolute = raw.startswith("global::")
+        raw = raw.removeprefix("global::")
+        # Keep generic arguments lossless; erasure could select a different
+        # declaration (I vs I<T>). Generic binding belongs to #943.
+        if not all(part.isidentifier() for part in raw.split(".")):
+            return []
```
merlincat11, line 155: **[P2] Preserve binding for nullable reference receivers**

The new lossless `_store_typed_binding()` passes `Service?` here, and `isidentifier()` rejects the trailing `?`. As a result, a previously resolved call on an annotated reference becomes unresolved even when its declaration and `using` are unambiguous:

```csharp
// Service.cs
namespace Other; public class Service { public void Run() {} }
// Caller.cs
#nullable enable
using Other; namespace App;
class C { void Go(Service? value) { value?.Run(); } }
```

After `full_build`, this PR leaves `Service?::Run` with `unresolved_targets: []`; `main` resolves the call to `Service.cs::Service.Run`. Nullable reference annotations do not change the underlying runtime type ([Microsoft documentation](https://learn.microsoft.com/en-us/dotnet/csharp/fundamentals/null-safety/nullable-reference-types#express-intent-with-annotations)). Retain the raw spelling while handling reference nullability separately from the deliberately deferred generic binding, so this supported receiver form keeps its call edge.

code_review_graph/parser.py
```diff
@@ -2765,6 +2811,15 @@ def parse_bytes(self, path: Path, source: bytes) -> tuple[list[NodeInfo], list[E
...
 
         edges = self._apply_typed_call_targets(edges, typed_call_targets, language)
+        if language == "csharp":
+            for edge in edges:
+                if edge.kind != "CALLS":
+                    continue
+                receiver = edge.extra.get("receiver_scope") or edge.extra.get("receiver")
+                if receiver:
+                    # Distinct receivers on one line must survive edge upsert.
+                    edge.target = f"{receiver}::{edge.target}"
```
merlincat11, line 2821: **[P2] Use call offsets when collecting and applying typed receiver evidence**

The new per-call edge preservation runs after `_apply_typed_call_targets()`, but `_collect_typed_call_targets()` still stores evidence by `(line, receiver, method)` (lines 5242–5243), and application uses that same key (line 5601). Two valid lexical bindings for the same receiver name on one line overwrite each other before `source_offset` can distinguish the edges.

Define public `A.Service.Run()` and `B.Service.Run()` in separate files, then keep this caller on one line:

```csharp
class C { void Go() { { A.Service s = new A.Service(); s.Run(); } { B.Service s = new B.Service(); s.Run(); } } }
```

`full_build` on this PR stores two distinct calls, but both have `receiver_scope = B.Service` and both resolve to `B.cs::B.Service.Run`; the first must target `A.cs::A.Service.Run`. On `main`, the receiver-type collision exists but the bare `Service` candidates remain unresolved, so the newly preserved qualification now turns that lost context into an incorrect resolved edge. Carry the byte offset through the typed-evidence collection and lookup as well as edge storage.

merlincat11 (commented):
Update for a6c2573: all three findings below are fixed, covered by regression tests, and resolved in their inline threads. The full local suite passes (3,026 tests), as do all 41 C# namespace tests and three real-observer watcher tests on Python 3.11. The previously failing Python 3.11 CI job also passed on this commit. The review below records the original findings against 7696119.

Reviewed `7696119b1e60a60e6fa3b0ae234f2e00b4a2473a` against `main` at `b58668751ab0c7670c078cf7cbd4d1f5b8e54f81`. **Keep this PR in draft: three correctness findings need fixes before merge.** Each was reproduced through `full_build` and the graph endpoint-resolution passes, then compared with the same fixtures on `main`.

The namespace identity and lexical-using design addresses #946 at the right boundary. The findings below concern lost call-site context and a nullable-receiver regression within that design.

| Area | Blast radius | Review result |
| --- | --- | --- |
| C# parser and binder | Call targets, impact traversal, and test mirrors | Three confirmed P2 findings in inline comments |
| Shared edge storage | Edge upserts for every language | Existing suite passes; C# offsets are exercised by storage tests |
| Incremental updates | Stored identity upgrades, per-file retries, deletion/rebinding | Existing migration and lifecycle regressions pass |

The graph reports 8 changed files, 64 changed functions/classes, overall risk score **0.40**, and 33 additional impacted files at the configured depth. It found no indexed affected flows; that is a graph-coverage limitation, not evidence that these changes cannot affect execution paths. I checked the source and tests directly.

Validation at this commit: **3,016 passed, 9 skipped, 2 xpassed**; Ruff, mypy, and wheel build passed. The three additional review fixtures expose behavior the current suite does not cover. Generic binding, inherited members, and MSBuild project evaluation remain outside this PR's stated scope.

CI follow-up: the [Python 3.11 test job](https://github.com/tirth8205/code-review-graph/actions/runs/34228066761/job/102067000333) failed only `tests/test_watch_robustness.py::TestRealObserver::test_recreated_directory_survives_and_is_reindexed` (the recreated Python file was not indexed). All other reported checks passed. The isolated test passes locally on Python 3.13.5; Python 3.11 is unavailable locally, so the cause is not established and I have not classified it as an introduced defect. CI is not fully green.

code_review_graph/csharp_resolver.py
merlincat11: Fixed in [a6c2573](https://github.com/tirth8205/code-review-graph/commit/a6c2573ca74e82125789737eae2b052d48e32974). Each C# call now retains its lexical containing type independently of its graph caller. The binder uses that context for field/property initializers, nested types, and unqualified initializer calls. The C# format gate is now 3 so cached versions 1 and 2 are reparsed, with the existing per-file retry behavior retained.

The original full-build repro now targets App.Service.Run. Added initializer and cached-context upgrade regressions. Validation: 3,026 tests passed; all 41 C# namespace tests also passed on Python 3.11.

code_review_graph/csharp_resolver.py
merlincat11: Fixed in [a6c2573](https://github.com/tirth8205/code-review-graph/commit/a6c2573ca74e82125789737eae2b052d48e32974). Member lookup removes the terminal nullable annotation while preserving the complete raw receiver spelling on the edge. Namespace qualifications stay intact, and constructed generic receivers such as Service<int>? remain unresolved rather than selecting the plain Service declaration.

Added coverage for nullable parameters, fields, and locals, normal and conditional calls, and fully qualified receivers. The original full-build repro now resolves to Other.Service.Run. Validation: 3,026 tests passed; all 41 C# namespace tests also passed on Python 3.11.

code_review_graph/parser.py
merlincat11: Fixed in [a6c2573](https://github.com/tirth8205/code-review-graph/commit/a6c2573ca74e82125789737eae2b052d48e32974). The C# typed-receiver collection and application phases now both key evidence by the AST byte offset, so the two s.Run() calls retain A.Service and B.Service before edge storage. The format gate forces cached calls to be reparsed with that corrected evidence.

The regression verifies distinct targets and TESTED_BY mirrors for calls on the same line and separate lines, with a UTF-8 prefix to exercise byte offsets. The original full-build repro now resolves to A.Service.Run and B.Service.Run respectively. Validation: 3,026 tests passed; all 41 C# namespace tests also passed on Python 3.11.

code_review_graph/incremental.py
```diff
@@ -1539,8 +1573,11 @@ def incremental_update(
     temporal_stats = _run_temporal_resolver(store) if spring_changed else None
     hcl_changed = any(rp.endswith((".tf", ".hcl")) for rp in all_files)
     hcl_stats = _run_hcl_resolver(store) if hcl_changed else None
-    scoped_changed = any(rp.endswith((".php", ".rs", ".cs")) for rp in all_files)
-    scoped_stats = _run_scoped_resolver(store) if scoped_changed else None
+    scoped_changed = any(
+        rp.endswith((".php", ".rs", ".cs", ".csproj"))
+        for rp in set(all_files) | set(stale_files) | missing_paths
+    )
+    scoped_stats = _run_scoped_resolver(store, repo_root) if scoped_changed else None
```
merlincat11, line 1580: **[P1] Invalidate derived flows when C# rebinding changes unchanged callers**

This resolver can rewrite `CALLS` edges in C# files that were not reparsed, but the result reports only counts and `changed_files` still contains only the edited using file. The incremental postprocessor therefore cannot identify the caller flows that must be retraced.

Reproduce with indexed `A.Service` and `B.Service`: build with `global using S = A.Service;` and `Caller.Go()` calling `S.Run()`, then change only the alias to `B.Service`. After the incremental build, the live edge is `Caller.Go -> B.Service.Run`, but the persisted flow remains `Caller.Go -> A.Service.Run -> A.Service.Finish`; recomputing flows from the live graph yields the B path. Even explicitly retracing with the changed `Imports.cs` path updates zero flows because that file is not a flow member.

Please return or otherwise propagate the callers/endpoints whose bindings changed and invalidate the corresponding derived flows (old and new paths). Until that affected set is available, a full flow refresh when C# rebinding changes an edge is the safe fallback. This is a logical graph inconsistency even though SQLite integrity remains `ok`.

code_review_graph/csharp_resolver.py
```diff
@@ -0,0 +1,289 @@
...
+        return local
+
+    def type_targets(raw: str, owner: str, scopes: list, file: str) -> list[str]:
+        absolute = raw.startswith("global::")
+        # Nullable receivers use the underlying declaration for member lookup;
+        # the original annotation remains in the call's stored evidence.
+        raw = raw.removeprefix("global::").removesuffix("?")
+        # Keep generic arguments lossless; erasure could select a different
+        # declaration (I vs I<T>). Generic binding belongs to #943.
+        if not all(part.isidentifier() for part in raw.split(".")):
```
merlincat11, line 154: **[P2] Preserve binding for namespace aliases used with `::`**

`type_targets()` recognizes only the special `global::` prefix. Any ordinary alias-qualified receiver reaches the identifier check with `::` still present and is rejected before the lexical alias metadata below can be consulted.

```csharp
using Alias = App;
namespace App { class Service { public static void Run() {} } }
class C { void Go() { Alias::Service.Run(); } }
```

The latest head leaves this valid call as unresolved `Alias::Service::Run`; the base resolves it to the indexed `Service.Run`. C# defines `::` specifically for namespace aliases, and its alias-only meaning can matter when the same spelling is also a type or namespace. Please parse the leading alias qualifier and resolve it through the existing directives for the call's lexical scope rather than rejecting the receiver.

code_review_graph/csharp_resolver.py
```diff
@@ -0,0 +1,289 @@
...
+        if not scopes:
+            return []
+        # Initializers have a File caller but still belong to a lexical type.
+        owner = extra.get("csharp_containing_type") or parents.get(
+            row["source_qualified"], scopes[0][0],
+        )
+        kind = extra.get("csharp_call_kind")
+        method = extra["csharp_raw_target"].rsplit("::", 1)[-1]
+        receiver = extra.get("receiver_scope", "")
+        if extra.get("receiver") == "this" or kind == "unqualified":
```
merlincat11: **[P2] Search enclosing types for unqualified static calls**

This branch treats an unqualified call like a `this` call and considers only `own_type`, so it drops valid lexical lookup through containing types:

```csharp
namespace App {
  class Outer {
    static void Run() {}
    class Inner { void Go() { Run(); } }
  }
}
```

The latest head leaves `Inner.Go -> Run` unresolved; the base resolves it to `Outer.Run`. A nested type can access static members of its containing type. Please walk the containing-type chain for `kind == "unqualified"` while keeping `this.Run()` restricted to the immediate instance type. This is lexical containing-type lookup, separate from the explicitly deferred inherited-member resolution.

code_review_graph/incremental.py
merlincat11: Verified and fixed in ced4f45. This is in scope because the PR deliberately rebinds calls in unchanged files.

The shared C# edge update now persists a flow-dirty marker with the binding change. The existing incremental flow tracer consumes it through a full retrace, covering both caller paths and changed entry points. The marker survives `minimal`/`none` postprocessing and a later update with no parsed files; successful full flow replacement clears it in the same transaction.

The regression switches A → B, removes/re-adds the alias, and deletes the callee file, comparing persisted entry points and paths with a fresh trace. It also verifies recovery after a failed flow replacement. This uses the existing full-trace path without adding an affected-endpoint framework.

code_review_graph/csharp_resolver.py
merlincat11: Verified and fixed in ced4f45. `Alias::Service` now uses the existing lexical directives and project-scoped global aliases, searching namespace aliases only.

Regressions cover the reported call, a same-named type/local, typed receivers, reopened namespace isolation, and ambiguous project ownership. A type alias used with `::` remains unresolved. This follows C#'s [namespace alias qualifier rules](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/namespaces#149-namespace-alias-qualifiers) without expanding generic or project-reference resolution.

code_review_graph/csharp_resolver.py
merlincat11: Verified and fixed in ced4f45. Unqualified calls now walk containing types, stop at the nearest method group, and bind an enclosing method only with parsed static evidence. `this.Run()` still searches the immediate type.

The tests cover the reported call, partial types across files, nearer methods, enclosing instance methods, and callable parameters/locals/fields/properties/local functions that hide the method. The latter reuse the parser's binding evidence to avoid introducing false edges. The existing cache-version gate now reparses older C# nodes for this evidence and retains its per-file retry behavior. Inherited-member lookup remains outside the PR.

Validation for this round: 3,048 full-suite tests passed; Ruff, CI mypy, wheel build, and the isolated Python 3.11 C# tests passed.

tirth8205:
Changes required: namespace-aware resolution drops a supported unambiguous generic receiver call. Build `Box.cs` containing `namespace App; public class Box<T> { public void Run() {} }` and `Caller.cs` containing `using App; class Caller { void Go(Box<int> value) { value.Run(); } }` with `full_build(root, store)`; main resolves the call to `Box.Run`, while this PR leaves raw `Box<int>::Run`. Preserve the supported generic case with arity evidence alongside the namespace repair.

merlincat11:
Fixed in 6031800. Your fixture now resolves:

```
Caller.Go -> Box.cs::App.Box.Run
```

The receiver is keyed by arity rather than erased, so the case #943's regression 2 warns about stays closed. Generic declarations are indexed as ``App.Box`1``; the parser records the type-parameter count, and a receiver's key is derived from its syntax rather than by scanning for angle brackets, so `Pair<string, int>` counts 2 and `Box<List<int>>` counts 1.

```
I<int>          -> Generic.cs::A.I.Run     the generic declaration
I               -> Plain.cs::A.I.Run       the separate non-generic one
Pair<int>       -> Pair<int>::Run          raw; only Pair<K, V> is declared
Outer<int>.Inner-> Outer<int>.Inner::Run   raw; a second arity one key cannot describe
```

Six cases added to the binding matrix. Two existing tests changed with the behaviour rather than around it: the `I<int>` case now asserts it reaches `I<T>` instead of asserting it stays raw, and the nullable `Service<int>?` case binds to the generic declaration while keeping its raw spelling in the call's evidence. The version gate moves to `"5"` so graphs built without arity evidence reparse, and the upgrade test now covers `"4"`.

Type argument substitution, constraints and overload selection remain out of scope, and the receiver's spelling is still retained on the call — only the lookup key is derived from it.

`uv run pytest tests/ -q` — 3,054 passed, 9 skipped, 2 xpassed; ruff and the CI mypy command pass. Rebuilding serilog gives an identical bound-call count (1,352), so the change is inert where the generic receivers are BCL types and only adds bindings for repository-local generic declarations.

tirth8205:
This no longer merges into `staging`. Conflicts in `CHANGELOG.md`, `code_review_graph/incremental.py`, `code_review_graph/tools/build.py`, `docs/architecture.md`.

```bash
# resolve, then
uv run pytest tests/ -q
git push
```

When resolving:
- Rebase PR onto current origin/staging and resolve conflicts in CHANGELOG.md, code_review_graph/incremental.py (6 hunks), code_review_graph/tools/build.py (1 hunk), docs/architecture.md
- Re-run full gates after rebase; edge-case coverage of tests/test_csharp_namespaces.py and tests/test_multilang.py not yet assessed

merlinhu1:
Merged the latest staging and pushed the conflict resolution in fc34671; GitHub confirms the PR is mergeable. Validation passed with 4,038 tests, including C# namespace and multilanguage coverage, plus Ruff and mypy.

tirth8205:
This does not merge into `staging` any more. The conflicts are `CHANGELOG.md`, two hunks in `code_review_graph/incremental.py`, and two hunks in `code_review_graph/parser.py`. fc34671 was mergeable when you pushed it. Staging moved a few hours later, the branch is 42 commits behind, and the API reports `mergeable_state: dirty`.

`parser.py` is the one to be careful with, and it is not really textual. Staging replaced `_extract_import(...)` with `_extract_import_specs(...)` returning `_ImportSpec(module, line, form)` plus `_resolve_import_target(...)`, and it now writes `extra={IMPORT_SCOPE_KEY: scope} if scope else {}` on every IMPORTS_FROM edge. Your branch writes the C# using metadata on that same `extra` field. Taking either side whole silently drops the other. Staging's side removes the lexical `using` evidence this PR is built on. Your side removes import scope for every language. Merge both dicts inside staging's per-spec loop. That is safe for C# because it yields one spec per node, unlike go and ruby. Nothing in the suite asserts import scope and C# using metadata on the same edge today, so a bad resolution still goes green.

The `incremental.py` hunk at the tail of `full_build` needs both sides too. Staging added `advance_to_postprocess_pending(store)`, you added the `_CSHARP_IDENTITY_METADATA_KEY` and `_CSHARP_PENDING_METADATA_KEY` writes, and the C# metadata has to be written before the checkpoint. The second hunk is just the import block.

The fixes from the last round hold up. I rebuilt fixture repos against plain `origin/staging` and dumped the CALLS edges rather than reading the diff. `namespace Deep.Ns; class Widget` now gives `Apple.cs::Deep.Ns.Widget.Make` where staging gave `Apple.cs::Widget.Make`, the generic receiver binds to `Box.cs::App.Box.Run`, `Svc? value; value?.Run()` binds direct, two distinct receivers on one line produce two edges instead of one unresolved one, and `using Al = A; Al::Service x; x.Run()` resolves. Build time is within noise.

Two costs the CHANGELOG does not cover.

A full build over a shallow serilog clone, 216 `.cs` files, moves direct C# CALLS from 1330 to 1353 net, but that is 108 gained and 88 lost. The losses concentrate as 30 in `Log.cs`, 20 in `LoggerTests.cs`, 15 in `ILogger.cs`. `ILogger.cs` produces no type node at all on either branch. The `#if FEATURE_DEFAULT_INTERFACE` guards defeat the parse, every method lands with `parent_name=None`, and there is a spurious `Function|if` node. Staging's short-name matching still bound `Log.Logger.ForContext()` to the concrete `Core/Logger.cs::Logger.ForContext`. This branch records receiver type `ILogger`, cannot find that type, and leaves the call unresolved, so serilog's public static entry point loses its edges into the concrete logger. A fallback for receivers whose type is simply not indexed, as opposed to not visible, would keep those.

The other is the read path. Unresolved C# targets now carry the receiver spelling.

```csharp
public class Base { void Shared() {} }
public class Derived : Base {}
void Go(Derived d) { d.Shared(); }
```

Before, the target is `Shared` and `query_graph("callers_of", "Base.Shared")` returns 1 call site. After, the target is `Derived::Shared`, `callers_of` on `ProjA.Base.Shared` returns 0, and `Base.Shared` returns `not_found`. Across serilog, unresolved targets that exactly match an indexed node name fall from 1313 to 532. Much of that is noise you were right to drop, since staging bound `(_targetSink as IDisposable)?.Dispose()` to `BatchingSink.Dispose`, which is wrong. It is still a user-visible reduction in `callers_of` and `references_to` answers for C#, and it belongs in the CHANGELOG.

What would make this mergeable:

- Re-merge `origin/staging` with `parser.py` resolved so both `extra` dicts survive.
- A test that asserts import scope and C# using metadata on the same IMPORTS_FROM edge.
- One CHANGELOG line for the unresolved-target spelling change.
- Re-run the gates on the merged tree. The clean run I measured, 4039 passed with 0 failed plus clean ruff and mypy, is for the unmerged head, and staging's parser refactor is exactly what the merge touches.

The `ILogger.cs` fallback is worth looking at, but I would not hold the merge on it.

merlinhu1:
Re-merged `origin/staging` in f6c833e. The branch is level with staging and the PR reports `mergeable_state: clean`.

**`parser.py`** — resolved inside staging's per-spec loop so both dicts survive:

```python
edges.append(EdgeInfo(
    kind="IMPORTS_FROM",
    ...
    extra={**csharp_extra, **({IMPORT_SCOPE_KEY: scope} if scope else {})},
))
```

`csharp_extra` is built once per `using` directive, before the loop over specs — which is safe here for the reason you gave, C# yields one spec per node. Staging's `_extract_import_specs(...)` / `_resolve_import_target(...)` path is otherwise untouched: the only staging line this merge removes from `_extract_imports` is the old single-dict `extra=`.

**`incremental.py`** — both sides kept at the tail of `full_build`. The `_CSHARP_IDENTITY_METADATA_KEY` and `_CSHARP_PENDING_METADATA_KEY` writes precede `advance_to_postprocess_pending(store)`. The second hunk is the import block, taken as you described.

**Test for the shared edge** — `tests/test_csharp_namespaces.py::test_import_scope_and_csharp_using_metadata_survive_together`, parametrised over scope `None`, `package` and `tree`. It reads the edges back out of the store and asserts `import_scope` alongside `source_offset`, `csharp_scopes`, `csharp_global` and the alias/namespace `csharp_using_kind` on the same two `IMPORTS_FROM` edges, so taking either side whole fails it rather than going green. I added `test_csharp_retry_metadata_precedes_build_checkpoint` for the other hunk; it captures the metadata at the checkpoint and fails if the C# writes move after it.

**CHANGELOG** — one line under `Unreleased / Fixed` for the read-path change:

> Unresolved C# member-call targets retain the receiver spelling (for example, `Derived::Shared` instead of `Shared`), so `callers_of` can return fewer matches for unsupported bindings, including inherited members and receiver types missing from the index (#946).

The C# entries stayed under `Unreleased` after staging's 2.3.9 release.

**Gates** — re-run on the merged tree rather than the old head: all checks green on f6c833e, including test on 3.10–3.13, lint, type-check, security, schema-sync and the Windows and E2E jobs.

Issue #940, opened 2026-08-30
inheritors_of can report a same-named class from another namespace as an inheritor


merlincat11:
Split out of #939 (review discussion) to keep that PR scoped to the parser bug in #935.

## Problem

`inheritors_of` resolves the requested node, and when exact qualified-target lookup finds
nothing it falls back to the node's plain `name`. That fallback matches edges by bare
target plus source language, which does not establish namespace or package identity.

```
A/GenericBase.cs:  namespace A; class GenericBase<T> {}
B/GenericBase.cs:  namespace B; class GenericBase<T> {}
B/Child.cs:        namespace B; class Child : GenericBase<string> {}
```

Querying the `A` declaration by its qualified graph name returns `B.Child`, with no
caveat attached:

```
$ code-review-graph query inheritors_of /repo/A/GenericBase.cs::GenericBase
count: 1  results: ['Child']   confidence: None
```

## Not specific to generics

Reproduces on `main` with plain `class Child : PlainBase {}` and two same-named
`PlainBase` declarations. This is the bare-name fallback, not the generic handling.

## Why it matters

#935 objected to the graph making unjustified affirmative claims about inheritance. A
false positive asserted without qualification is the same failure pointing the other way.

## Options

1. Resolve the base name against namespace/package/import context before treating the
   relationship as resolved — the real fix, and the one worth doing.
2. Attach a caveat when the fallback matched an ambiguous bare name, so the result is
   returned but not asserted. A prototype existed in #939 (restricting the ambiguity
   count to `Class`/`Type` kinds) and was removed to keep that PR scoped.

Option 2 is a mitigation; option 1 is the fix.

---

## Status

**#966 (open) implements option 2.** Ambiguous bare-name matches are retained and
marked `inferred_by: "bare_name"`, in both standard and `detail_level="minimal"`
output, scoped to `inheritors_of` so `tests_for` is unaffected. The false positive
in the repro above is still returned — it is caveated rather than asserted. This
issue stays open for option 1.

## Negative result: option 1 does not work at the query layer

The first implementation of #966 (`e7f5f05`) attempted option 1 at query time,
resolving the base name from file-level namespace and import metadata: same file,
an `IMPORTS_FROM` edge to the declaring file, or a namespace the declaring file
declares, walking outwards to enclosing namespaces.

Measured against `dotnet/eShop` (547 C# files; 8 genuinely ambiguous inheritance
targets; ground truth read from the namespaces and `using` directives in source):

[...]

PR #966, opened 2026-09-09
fix(query): caveat ambiguous bare-name inheritor candidates

merlincat11:
When several indexed declarations share a base name, `inheritors_of` can return a child of another declaration without a caveat. For example, querying `A/Base.cs::PlainBase` can return `B.Child` through a bare `PlainBase` edge. This change retains those candidates and marks them with `inferred_by: "bare_name"` in both standard and minimal output.

This is the query-side mitigation described in #940. Full declaration binding remains tracked by #943; #940 should remain open for that work.

- Count exact-name `Class`/`Type` declarations within the same language family used by the existing bare-edge lookup. Methods and unrelated languages do not create ambiguity; JavaScript, TypeScript, and TSX share the count.
- Stream candidates through the existing bounded result collector, preserving result counts, omitted counts, and aligned edges.
- Preserve the marker in compact inheritance results. Other query projections retain their current behavior.

File-level namespaces, imports, and same-file placement cannot establish lexical binding or prove that a base is inaccessible. This query therefore keeps every ambiguous bare-name candidate marked for verification. Exact-target lookup and the existing single-indexed-declaration fallback retain their current behavior; the latter remains a heuristic. Parser output, stored edge targets, generic handling, and resolvers are outside this change.

Regression coverage includes C# global/relative usings, type aliases, global bases, enclosing namespaces and shadowing, multiple namespaces in one file, Java package/import candidates, compact output, bounded edge retention, exact targets, Dart member names, and JS-family declaration counts. Fourteen regression cases fail against the original PR implementation and pass with this version.

A local `tracemalloc` smoke check with 10,000 inheritors and `max_results=1` measured about 43 KB peak Python allocation, versus 12.7 MB with the original PR and 42 KB on the base commit. Both single- and multiple-declaration cases retained all 10,000 in the total count while returning one result.

Validation: `uv run pytest --tb=short -q --cov=code_review_graph --cov-report=term --cov-fail-under=65` — 3,002 passed, 9 skipped, 2 xpassed; 85.04% coverage. Ruff, the CI mypy command, Bandit, and `uv build --wheel` passed locally.

### DIFF
--- a/code_review_graph/tools/query.py
+++ b/code_review_graph/tools/query.py
@@ -604,8 +604,10 @@ def query_graph(
             # (e.g. "Animal") while qn is fully qualified
             # (e.g. "sample.dart::Animal"). Search by plain name too. See: #87
             if total_results == 0 and node:
-                # ponytail: indexed-name ambiguity only; binding belongs in #943.
-                # File namespaces/imports cannot prove which declaration a base names.
+                # Ambiguity is measured on the indexed name alone: several
+                # declarations answer to this bare base name. Which one it
+                # binds to is #943's work — file namespaces and imports cannot
+                # prove it — so caveat the matches instead of guessing.
                 languages = (
                     _compatible_edge_languages(node.language) if node.language else (None,)
                 )
--- a/code_review_graph/tools/query.py
+++ b/code_review_graph/tools/query.py
@@ -116,103 +116,6 @@ def _rank_disambiguation_candidates(
     return [node_to_dict(node) for node in sorted(candidates, key=score)]
 
 
-def _declared_namespaces(store: GraphStore, file_path: str) -> set[str]:
-    """Return the namespaces a file declares, when the parser records them."""
-    file_node = store.get_node(file_path) if file_path else None
-    extra = getattr(file_node, "extra", None)
-    declared = extra.get("csharp_namespaces") if isinstance(extra, dict) else None
-    if not isinstance(declared, list):
-        return set()
-    return {namespace for namespace in declared if isinstance(namespace, str)}
-
-
-def _namespace_scope(namespaces: set[str]) -> set[str]:
-    """Return *namespaces* plus the enclosing namespaces they can see into.
-
-    A declaration in ``A.Sub`` names a type in ``A`` without a using
-    directive, so lookup has to walk outwards. The reverse does not hold:
-    ``A`` cannot see ``A.Sub`` unqualified.
-    """
-    scope = set()
-    for namespace in namespaces:
-        while namespace:
-            scope.add(namespace)
-            namespace = namespace.rpartition(".")[0]
-    return scope
-
-
-def _declaration_visibility(
-    store: GraphStore, node: GraphNode, child: GraphNode, declared: set[str],
-) -> str:
-    """Return whether *child* can name the declaration in *node*.
-
-    ``"visible"`` on evidence, never proximity: the same file, an import of
-    the declaring file, or — for C#, whose ``using`` imports a namespace
-    rather than a file — a namespace the declaring file declares.
-
-    ``"hidden"`` only when both files name a namespace and they disagree,
-    which is positive evidence that this declaration is out of reach.
-    Everything else is ``"unknown"``: absent evidence is not counter-evidence,
-    and a language whose packages the graph does not record must not have its
-    correct answers deleted on that silence.
-    """
-    if not node.file_path or not child.file_path:
-        return "unknown"
-    if node.file_path == child.file_path:
-        return "visible"
-    imported = {
-        edge.target_qualified
-        for edge in store.iter_edges_by_source(child.file_path)
-        if edge.kind == "IMPORTS_FROM"
-    }
-    if node.file_path in imported:
-        return "visible"
-    if not declared:
-        return "unknown"
-    child_namespaces = _declared_namespaces(store, child.file_path)
-    if declared & (imported | _namespace_scope(child_namespaces)):
-        return "visible"
-    return "hidden" if child_namespaces else "unknown"
-
-
-def _resolve_bare_inheritors(
-    store: GraphStore, node: GraphNode, candidates: list[tuple[GraphNode, Any]],
-) -> list[tuple[GraphNode, Any, bool]]:
-    """Attach identity to bare-name inheritor matches. See: #940.
-
-    ``INHERITS``/``IMPLEMENTS`` targets are bare base names, so the fallback
-    that matches them by name cannot tell two same-named declarations apart
-    and would report a class from another namespace as an inheritor without
-    any caveat. One declaration of the name is unambiguous and stays exactly
-    as before; several means each match is judged on its own visibility. A
-    match the graph cannot place is kept and flagged as inferred from the bare
-    name — dropping it would delete the correct answer whenever a sibling
-    happens to carry stronger evidence.
-
-    Returns ``(child, edge, inferred)`` triples.
-    """
-    languages = _compatible_edge_languages(node.language) if node.language else (None,)
-    declarations = sum(
-        store.count_nodes_by_name(node.name, language=language, kinds=("Class", "Type"))
-        for language in languages
-    )
-    if declarations <= 1:
-        return [(child, edge, False) for child, edge in candidates]
-    declared = _declared_namespaces(store, node.file_path)
-    # Visibility depends only on the child's file, and one file commonly holds
-    # several inheritors, so each file is placed once per query.
-    by_file: dict[str, str] = {}
-    resolved = []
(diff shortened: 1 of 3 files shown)


code_review_graph/tools/query.py
```diff
@@ -109,6 +116,103 @@ def score(node: GraphNode) -> tuple[int, str]:
...
+        if edge.kind == "IMPORTS_FROM"
+    }
+    if node.file_path in imported:
+        return "visible"
+    if not declared:
+        return "unknown"
+    child_namespaces = _declared_namespaces(store, child.file_path)
+    if declared & (imported | _namespace_scope(child_namespaces)):
+        return "visible"
+    return "hidden" if child_namespaces else "unknown"
```
merlincat11: [P1] Different file namespaces do not prove that the requested base is inaccessible, so this branch drops valid inheritors. With `A.PlainBase` and `B.PlainBase`, a `global using A;` in another file makes `C.Child : PlainBase` valid, but querying the A declaration returns `Child` on the base commit and zero on this PR, labelled a "real absence". I reproduced the same regression with a type alias, a relative namespace import, and a global base class sharing its file with an unrelated namespace declaration. Please retain and mark these candidates as `inferred_by: "bare_name"` when declaration identity cannot be established, rather than treating file-level namespace disagreement as sufficient evidence to exclude them.

code_review_graph/tools/query.py
```diff
@@ -597,13 +701,21 @@ def add_result(result: dict[str, Any], edge: Any | None = None) -> None:
...
             if total_results == 0 and node:
+                bare_matches: list[tuple[GraphNode, Any]] = []
                 for kind in ("INHERITS", "IMPLEMENTS"):
                     for e in store.iter_edges_by_target_name(
                         node.name, kind=kind, language=node.language or None,
                     ):
                         child = store.get_node(e.source_qualified)
                         if child:
-                            add_result(node_to_dict(child), e)
+                            bare_matches.append((child, e))
```
merlincat11: [P2] This accumulation removes the existing bounded-memory behavior: the query previously streamed matches through `add_result`, retaining only the requested prefix while counting all results. It now retains every node/edge pair in `bare_matches` and constructs a second complete list in `_resolve_bare_inheritors`, even for an unambiguous base that needs no visibility filtering. On a synthetic graph with 10,000 inheritors and `max_results=1`, I measured peak Python allocation with `tracemalloc` increasing from 42,672 bytes on the base commit to 13,688,100 bytes on this PR, although both returned just one result. Please stream the classified matches through `add_result` instead of materializing both lists, preserving the existing total-count and response-limit behavior.

code_review_graph/tools/query.py
```diff
@@ -109,6 +116,103 @@ def score(node: GraphNode) -> tuple[int, str]:
...
+        for edge in store.iter_edges_by_source(child.file_path)
+        if edge.kind == "IMPORTS_FROM"
+    }
+    if node.file_path in imported:
+        return "visible"
+    if not declared:
+        return "unknown"
+    child_namespaces = _declared_namespaces(store, child.file_path)
+    if declared & (imported | _namespace_scope(child_namespaces)):
+        return "visible"
```
merlincat11: [P2] Visibility does not establish which declaration the base name binds to.
With both `A.PlainBase` and `A.Sub.PlainBase`, I reproduced `A.Sub.Child : PlainBase` being returned for the outer A declaration without a caveat, although the nearer declaration wins.
The namespace union accepts the outer scope without checking shadowing, so this portion of #940 remains unfixed.
Please account for binding precedence before asserting the match, or retain `inferred_by: "bare_name"` when the selected declaration cannot be established.

code_review_graph/tools/query.py
```diff
@@ -109,6 +116,103 @@ def score(node: GraphNode) -> tuple[int, str]:
...
+    declarations = sum(
+        store.count_nodes_by_name(node.name, language=language, kinds=("Class", "Type"))
+        for language in languages
+    )
+    if declarations <= 1:
+        return [(child, edge, False) for child, edge in candidates]
+    declared = _declared_namespaces(store, node.file_path)
+    # Visibility depends only on the child's file, and one file commonly holds
+    # several inheritors, so each file is placed once per query.
+    by_file: dict[str, str] = {}
```
merlincat11: [P2] Visibility is lexical, so a file-wide namespace union and cache cannot establish it for each child.
I reproduced a file containing an unrelated `namespace A { class Marker {} }` and `namespace B { class Child : PlainBase {} }` returning Child for a separate `A.PlainBase`, without a caveat.
The presence of namespace A elsewhere in the file does not put the B child in that scope, and caching by file propagates the same classification to every child there.
Please use declaration/lexical scope evidence, such as that introduced by #964, or keep these matches uncertain rather than treating file metadata as proof.

code_review_graph/tools/query.py
```diff
@@ -597,13 +701,21 @@ def add_result(result: dict[str, Any], edge: Any | None = None) -> None:
...
                         child = store.get_node(e.source_qualified)
                         if child:
-                            add_result(node_to_dict(child), e)
+                            bare_matches.append((child, e))
+                for child, e, inferred in _resolve_bare_inheritors(
+                    store, node, bare_matches,
+                ):
+                    child_result = node_to_dict(child)
+                    if inferred:
+                        child_result["inferred_by"] = "bare_name"
```
merlincat11: [P2] Preserve this ambiguity marker in `detail_level="minimal"` as well.
I verified that the two same-named Java base declarations produce caveated matches in standard output, but the projection later in `query_graph` removes `inferred_by` from both compact results.
Although the PR acknowledges this limitation, compact callers receive neither this marker nor a replacement warning, leaving #940's unsupported affirmative answer unchanged in that mode.
Please retain `inferred_by` for inheritance results and add a compact-output regression check; this can be scoped to `inheritors_of` without redesigning uncertainty fields or changing `tests_for`.

merlincat11 (commented):
Three additional verified gaps in the mitigation: binding precedence, lexical scope, and preserving uncertainty in compact results.

code_review_graph/tools/query.py
merlincat11: Confirmed independently: with `using Lib;` living in a GlobalUsings.cs, querying the A declaration returned zero and labelled it a real absence — and eShop ships 10 such files, so this is mainstream .NET layout rather than a corner case. fd1bad7 removes the visibility filter outright; ambiguous bases are now retained and marked `inferred_by: "bare_name"` instead of excluded, so no correct inheritor can be dropped.

code_review_graph/tools/query.py
merlincat11: Fixed in fd1bad7 — `bare_matches` and the second list built in `_resolve_bare_inheritors` are both gone, and matches stream through `add_result` again, so the total-count and response-limit behaviour is byte-for-byte the prior one. `test_inheritors_of_streams_bare_matches_and_keeps_counts` guards the streaming path and the counts.

code_review_graph/tools/query.py
merlincat11: Taken the second of your two options: fd1bad7 stops asserting a binding at all, so `A.Sub.Child` queried against the outer `A.PlainBase` now comes back marked `inferred_by: "bare_name"` rather than silently asserted. Binding precedence stays with #943 instead of being approximated here.

code_review_graph/tools/query.py
merlincat11: Agreed — the file-level namespace union and the per-file classification cache are both removed in fd1bad7, so no file metadata is treated as proof of lexical scope. Rather than adopt the #964 machinery and couple this PR to an unmerged one, ambiguous matches are simply kept uncertain and caveated.

code_review_graph/tools/query.py
merlincat11: Done in fd1bad7, scoped exactly as you suggested: the minimal projection now carries `inferred_by` for `inheritors_of` only, leaving `tests_for` and its `naming_convention` marker untouched. The regression check is built into the new tests, which assert marker parity across both `standard` and `minimal`.

tirth8205:
Integrated on `integration/token-efficiency-hardening` as `c84531d`. Ambiguous bare-name inheritor candidates remain available with `inferred_by='bare_name'` in both standard and minimal output, rather than being asserted as proven bindings. The bounded query and token-budget regressions passed; this is an uncertainty repair, not complete namespace or inheritance resolution.

tirth8205:
Merged into `staging` through the integration PR #988 (merge commit 3030ca8). Your change is carried by commit c84531d, which keeps you as the author (MerlinH). It will be promoted staging → testing → main and released from there. Thank you for the contribution! Closing this PR since its content is now on staging.

PR #972, opened 2026-09-10
fix(parser): index JS/TS components wrapped in forwardRef, memo and other HOCs

### LINKED ISSUE #971: `export const X = forwardRef(...)` produces no node, so every JSX use of a wrapped component dangles
## Summary

In JS/TS, `_extract_js_var_functions` only creates a `Function` node when the
variable declarator's value is *directly* an `arrow_function`,
`function_expression` or `function`. A component wrapped in a higher-order
function puts that function inside a `call_expression`, so the declaration is
skipped entirely — **no node is created**.

Because JSX uses still resolve to `<file>::<Name>`, every use of such a component
becomes a `CALLS` edge pointing at a node that does not exist. `callers_of`,
`tests_for` and `get_impact_radius` all return empty for it, with no indication
that the target was never indexed.

This affects the most common way to declare a React component that needs a ref —
`forwardRef` — plus `memo`, `styled`, `observer` and `withRouter`.

## Reproduction

```tsx
// Button.tsx
import { forwardRef } from 'react';
export const Button = forwardRef<HTMLButtonElement, Props>((props, ref) => (
  <button ref={ref} {...props} />
));

// Toolbar.tsx
import { Button } from './Button';
export const Toolbar = () => <Button>Save</Button>;
```

```
code-review-graph build
```

**Expected:** a `Function` node `Button.tsx::Button`, and `callers_of` it returns `Toolbar`.

**Actual:** no node for `Button`. The `CALLS` edge from `Toolbar` targets
`Button.tsx::Button`, which matches nothing. `callers_of('...Button.tsx::Button')`
returns `not_found`.

A plain `export const Toolbar = () => ...` in the same file *is* indexed, which is
what makes this hard to notice — the file has nodes, just not the one you wanted.

## Impact, measured on a real codebase

A React 19 / TypeScript app, 581 files, ~27k `CALLS` edges. Counting `CALLS`
edges whose target is a qualified name with no matching node:

| | before | after fix |
| :--- | ---: | ---: |
| dangling qualified `CALLS` targets | **1797 / 27090 (6.6%)** | 477 / 27057 (1.8%) |
| nodes | 5134 | 5167 |
| duplicate `qualified_name`s | 0 | 0 |
| `callers_of(Button)` | **0** | **109** |

The three worst single targets were the app's three most-used primitives —
`Text` (473 edges), `Button` (287), `Heading` (79) — all `forwardRef`.

## Suggested fix

Unwrap a *recognised* wrapper when the declarator's value is a `call_expression`.
Patch against 2.3.8 attached below; it adds `_unwrap_hoc_function` and one
`elif` branch in `_extract_js_var_functions`.

The wrapper list is deliberately closed rather than "any call with a function
argument": `const x = useMemo(() => ..., [])` also holds a function inside a
call, but the const is that call's **result**, not the function, so indexing it as
a function would be wrong.

```python
_JS_HOC_WRAPPERS = frozenset({
    "forwardRef", "memo", "observer", "styled", "withRouter",
    "React.forwardRef", "React.memo",
})
```

Generic parameters are stripped before the lookup, so
`forwardRef<HTMLButtonElement, ButtonProps>(...)` matches. Nesting is handled
recursively for `memo(forwardRef(fn))`.

<details>
<summary>Patch against 2.3.8</summary>

```diff
--- a/code_review_graph/parser.py
+++ b/code_review_graph/parser.py
@@ -8622,6 +8622,48 @@
         {"arrow_function", "function_expression", "function"},
     )
 
+    # Higher-order wrappers whose argument is the real component/function.
+    # `export const Button = forwardRef((props, ref) => {...})` puts the arrow
+    # function inside a call_expression, so the declarator has no direct
+    # function child and the declaration produces no node at all — every JSX
+    # use of it then dangles on a target that does not exist.
+    _JS_HOC_WRAPPERS = frozenset({
+        "forwardRef", "memo", "observer", "styled", "withRouter",
+        "React.forwardRef", "React.memo",
+    })
+
+    def _unwrap_hoc_function(self, value_node):
+        """Return the function inside a known HOC call, or None.
+
+        Only unwraps recognised wrappers: `useMemo(() => ..., [])` also holds a
+        function in a call, but the const is its RESULT, not the function, and
+        indexing that as a function would be wrong.
+        """
+        if value_node is None or value_node.type != "call_expression":
+            return None
+
+        callee = None
+        for ch in value_node.children:
+            if ch.type in ("identifier", "member_expression"):
+                callee = ch.text.decode("utf-8", errors="replace")
+                break
+        if callee is None:
+            return None
+        callee = callee.split("<", 1)[0].strip()
+        if callee not in self._JS_HOC_WRAPPERS:
+            return None
+
+        for ch in value_node.children:
+            if ch.type != "arguments":
+                continue
+            for arg in ch.children:
+                if arg.type in self._JS_FUNC_VALUE_TYPES:
+                    return arg
+                inner = self._unwrap_hoc_function(arg)  # memo(forwardRef(fn))
+                if inner is not None:
+                    return inner
+        return None
+
     def _extract_js_var_functions(
         self,
         child,
@@ -8659,6 +8701,10 @@
                     var_name = sub.text.decode("utf-8", errors="replace")
                 elif sub.type in self._JS_FUNC_VALUE_TYPES:
                     func_node = sub
+                elif func_node is None:
+                    # const X = forwardRef(...) / memo(...) — the function is an
+                    # argument of the wrapper, not a direct child.
+                    func_node = self._unwrap_hoc_function(sub)
 
             if not var_name or not func_node:
                 continue
```

</details>

## Three related cases the patch does NOT address

Raised together because they share a shape — a JSX member expression resolving to
a name that does not match the declared node — but each needs a different fix.

**1. `Object.assign` compound components.** 

```tsx
export const ListItemCard = Object.assign(ListItemCardRoot, {
  Header: ListItemCardHeader,
  Body: ListItemCardBody,
});
```

`<ListItemCard.Header>` resolves to `ListItemCard.tsx::Header`; the declared node
is `ListItemCardHeader`. `_resolve_jsx_component_target` treats `Base.Member` as a
module export, which is right for `import * as X` and wrong here. Fixing it needs
the defining module's `Object.assign` mapping, which the consumer's parse does not
have — so it probably belongs in the existing post-processing pass that resolves
bare edge endpoints, rather than in the parser.

**2. Property-assignment compounds.**

```tsx
export const Alert: React.FC<Props> & Sub = (...) => ...;
Alert.Icon = ({ children }) => <span>{children}</span>;
```

Here `_extract_js_member_functions` *does* create a node, qualified by its full
member path: `Alert.tsx::Alert.Icon`. But the JSX consumer emits
`Alert.tsx::Icon`, so they still miss each other. Emitting the `Base.Member` form
from the consumer would fix this but break genuine namespace imports — and
`import_map` is a flat `dict[str, str]` of name → module, so the resolution site
cannot currently tell a namespace import from a named one. Recording that
distinction would fix this case cleanly.

**3. Library context members** — `<AuthContext.Provider>` resolves to
`AuthContext.tsx::Provider`. `Provider` belongs to React, not to the file. Leaving
this unresolved is arguably correct; noting it only so it is not mistaken for the
same bug.

## Note on the `confidence` string

Every query response carries:

> `js/ts callbacks land on REFERENCES not CALLS, and obj[name]() is unresolved, so callers can be missing`

That is accurate but did not lead me here — it points at dynamic access, whereas
these targets *were* statically resolved and simply had nothing to match. If the
first fix lands, it might be worth having queries distinguish "target not indexed"
from "no results", since the two mean very different things to a caller. The
`not_found` status already does this for a directly-queried node; the gap is only
for a node reached through an edge.

L4XB:
# Pull Request

## Linked issue

Closes #971

## What & why

`_extract_js_var_functions` only created a `Function` node when a declarator's value was directly an arrow function or function expression. A component wrapped in a higher-order call, like the issue's `forwardRef<HTMLButtonElement, Props>((props, ref) => ...)`, has its function inside a `call_expression`, so no node was created. JSX uses still resolved to `<file>::<Name>`, so every use became a `CALLS` edge to a node that does not exist, and `callers_of`, `tests_for` and `get_impact_radius` came back empty for these components.

**What counts as a wrapped component.** `_js_wrapped_function` accepts a call only when all of these hold:

- The wrapper name is one of `forwardRef`, `memo`, `observer`, `withRouter`, `withStyles`, `withTheme`, `connect`, `styled`, `inject`, `withTranslation`, `withErrorBoundary` (`_JS_COMPONENT_WRAPPERS`). The name is the callee itself, the last segment of a member callee (`React.memo`), or for a curried call the name of the inner callee (`styled(Base)(fn)`, `connect(mapState)(fn)`).
- The call takes exactly one argument. `memo` is the one exception: it also takes the props comparator, `memo(Component, arePropsEqual)`, and nothing more. Comments between the arguments are not counted.
- That argument is a function literal, or another wrapper call that passes these checks, nested at most three levels below the outer call (`memo(forwardRef(fn))`).

The name is needed because the shape alone does not show that the result is callable: `const result = evaluate(() => compute())` has the same shape as `memo(() => paint())` and returns a number. Calls that fail a check are handled exactly as before. Examples: `useMemo(() => x, [dep])`, `useCallback(fn, deps)`, `createClient({...})`, ``styled.h1`...` ``, `memo(wrap(fn))` with an unknown `wrap`, and wrappers nested past the limit.

A wrapped declaration gets the same treatment as a directly assigned function:

- a `Function` node with the declaration's line range, and parameters and return type from the wrapped function;
- a `CONTAINS` edge;
- a walk of the function body, so calls inside the component are attributed to it.

**What stays in the enclosing scope.** Only the wrapped function is the component. The following parts run where the declaration is, so their edges keep the enclosing function (or the file, at module scope) as the source, as on staging:

- every wrapper call in the chain (`setup -> memo`, `setup -> forwardRef`);
- the inner call of a curried wrapper, with the calls and references in it (`styled(Base)`, `connect(mapState, ...)`);
- type arguments (`forwardRef<HTMLButtonElement, Props>`) and the type annotation of the declaration (`const Card: FC<Props> = memo(...)`);
- memo's comparator: the calls in an inline one, and the reference to a named one.

**Other declarators in the same statement** are now walked in the enclosing scope, both when a declarator is wrapped and when a function is assigned directly. This is a change from staging, which skipped the whole statement whenever one declarator was a directly assigned function: `const a = () => x(), b = y()` lost the call to `y`. The type annotation of a directly assigned function is still not walked, as on staging.

**Known false positives.** Wrappers are recognised by name, and for a member callee by its last segment, without checking where the name comes from. So a non-React call with the same name is also read as a wrapper if it takes a single callback (for `memo`, a callback plus at most one more argument). Examples are `pool.connect(cb)` and `cache.memo(fn)`: the variable becomes a `Function` node, and the calls inside the callback are attributed to it. The corpus below has no such case: all 644 wrapped declarations there are `memo`, `React.memo`, `forwardRef`, `React.forwardRef`, `ReactModule.forwardRef` or `withErrorBoundary`. Restricting member callees to `React.*` would remove these false positives, but it would also miss namespace imports under other names (`ReactModule.forwardRef` above), so I have left that choice to you.

**Corpus check.** I parsed 2058 React source files from cherry-studio, dify, LibreChat and grafana with staging (`f0e4eb7`) and with this branch, and compared their nodes and edges.

- **Lost:** one staging edge, a `TESTED_BY` edge in grafana's `validators.test.tsx`. That test renders `<Component />` inside `memo(() => ...)`. The call now belongs to the new `wrapped` component rather than to the test, and `TESTED_BY` follows the call. Staging does the same when the function is assigned directly.
- **Added:** 706 new nodes. 644 of them are wrapped components; calls moved into them or now resolve to them. Two of those are named `TestRunMenu` and `TestInfoTab`, so the existing name rule makes them `Test` nodes and 36 new `TESTED_BY` edges point at them. Staging does the same for a directly assigned function with those names.
- **Monaco bundles:** the other 62 new nodes are in two minified Monaco bundles under `dify/web/public/vs`. There the branch now walks the other declarators of a statement that assigns a function directly, which also adds 280 call edges that staging did not record.

## How it was tested

These are the CI commands, run at `d26609b` in a Python 3.10 virtualenv (`pip install -e ".[dev]" pytest-cov`, plus mypy, types-networkx and bandit):

```bash
ruff check code_review_graph/                                          # All checks passed!
ruff format --check tests/test_js_wrapped_functions.py                 # 1 file already formatted
mypy code_review_graph/ --ignore-missing-imports --no-strict-optional  # mypy 2.3.1 and 1.15.0: no issues found in 75 source files
bandit -r code_review_graph/ -c pyproject.toml                         # No issues identified
pytest --tb=short -q -m "not browser and not upgrade" \
  --cov=code_review_graph --cov-report=term-missing --cov-fail-under=65
# 3983 passed, 728 skipped, 46 deselected, 2 xfailed, 2 xpassed; total coverage 86.39%
```

The skips are mostly the opt-in suites: 715 of 728 (cli_surface, platform_lifecycle, action_e2e, exports, concurrency, corpus, surface, determinism). The other 13 are igraph not installed (5), Playwright not installed (2), four token-budget worst cases that the suite reports instead of asserting, one Windows-only path test, and one interpreter probe that needs user-site packages.

On Python 3.13:

```bash
pytest -q tests/test_js_wrapped_functions.py tests/test_callback_context_preservation.py   # 23 passed
```

`tests/test_js_wrapped_functions.py` (21 tests) covers:

- the issue's `forwardRef<HTMLButtonElement, Props>` reproduction (node, line range, `CONTAINS` edge);
- nested and curried wrappers;
- `React.memo` and `React.forwardRef<T, P>` callees;
- the two-file `Toolbar -> Button` case, where the `CALLS` target now exists;
- calls inside a component belong to it, while wrapper calls, the curried inner call, type arguments, the declaration's annotation and memo's comparator belong to the enclosing function;
- values that are not definitions: `choose(() => left(), () => right())`, `useMemo(fn, [dep])`, `evaluate(() => compute())`, `memo(wrap(fn))`, wrappers nested past the depth limit, `forwardRef(fn, extra)` and `memo(fn, eq, extra)`;
- `memo(Component, arePropsEqual)` with a named and an inline comparator;
- comments among the arguments;
- the other declarators of a statement, in both the wrapped and the direct form;
- that a directly assigned function's annotation is still not walked.

I also mutated the wrapper code in 25 ways (argument counts, depth limit, name check, member callees, comment filtering, which scope each part is walked in). Every mutation fails at least one test.

## Checklist

- [x] All tests pass (ran the CI test job's command above)
- [x] Linting passes (`ruff check code_review_graph/`)
- [x] Type checking passes (`mypy code_review_graph/ --ignore-missing-imports --no-strict-optional`)
- [x] Docs updated where behavior changed (docstrings of `_js_wrapped_function`, `_js_wrapper_name` and `_extract_js_var_functions`; no README or `docs/` page covers this)


tirth8205:
Changes required: stopping traversal at the first callback removes calls already present on main. Parse `function setup() { const result = choose(() => left(), () => right()); return result; }` with `CodeParser(repo_root).parse_file(path)`; this PR retains only `left` under a new `result` Function and loses `choose` and `right`. Preserve wrapper and other-argument calls, and avoid treating ordinary values returned by callback-taking functions as callable definitions.

L4XB:
Fixed at `c0668b0`, and you were right on both halves. I reproduced your snippet first:

```
main:        funcs=[setup]          calls=[setup->choose, setup->left, setup->right]
this PR:     funcs=[setup, result]  calls=[result->left]
now:         funcs=[setup]          calls=[setup->choose, setup->left, setup->right]
```

**Ordinary values are no longer treated as definitions.** A wrapper call now has to take the function as its *only* argument. That is the shape of the cases this PR is for, `forwardRef(fn)`, `memo(fn)`, `observer(fn)`, `memo(forwardRef(fn))`, `styled(Base)(fn)`, and it excludes every call that takes a callback among others, `choose(() => left(), () => right())` and `useMemo(() => compute(), [dep])`, which assign whatever the call returns. Those go back to the generic path, exactly as on `main`. The argument count is read from `arguments.named_children`, so `forwardRef<HTMLButtonElement, Props>(fn)` still matches: type arguments are a separate node.

**The wrapper and its siblings are preserved.** The body walk started at the wrapped function, which is why the wrapper call itself went missing even in the cases the PR handles. It starts at the declarator now, so everything in the initializer is visited, while the signature still comes from the function:

```
export const Card = memo(forwardRef((props, ref) => { paint(); return null; }));
  funcs=[Card], calls=[Card->memo, Card->forwardRef, Card->paint], params=(props, ref)
export const Button = forwardRef((props, ref) => { paint(); return null; });
  funcs=[Button], calls=[Button->forwardRef, Button->paint]
export const Box = styled(Base)((props) => { theme(); return null; });
  funcs=[Box], calls=[Box->styled, Box->theme]
```

Before this commit the first case reported only `Card->forwardRef` and `Card->paint`, and the second only `Button->paint`.

Three tests added for it: your snippet with the exact expected call set, the two-argument hook, and the nested wrapper asserting all three calls. All three fail if the single-argument rule is relaxed. `tests/test_js_wrapped_functions.py` is at 8 passed, the JS and parser suites at 303 passed, and the unit suite at 2952 passed, 9 skipped, 2 xpassed. `ruff check` is clean on both files and `ruff format --check` is clean on the test file (`parser.py` is already unformatted on `main` under ruff 0.14, so I left it as it is rather than reformat 8000 lines).

tirth8205:
Rechecked `c0668b0`: the original two-callback regression is fixed, but ten preservation checks still fail despite 392 existing tests passing. With `const result = evaluate(() => compute())` inside `setup`, the graph invents a `result` Function and makes `callers_of(compute)` return `result` instead of `setup`. A sibling declaration such as `const Button = memo(() => paint()), token = nextToken()` also drops `nextToken`; copy `tests/test_callback_context_preservation.py` from `integration/token-efficiency-hardening` and run `python -m pytest -q tests/test_callback_context_preservation.py` to reproduce both regressions. Please establish that the result is callable, preserve every declarator, and keep wrapper initialisation in its enclosing scope.

L4XB:
Both reproduced and fixed in `1d0dbb6`. Your file runs clean against it:

```
$ python -m pytest -q tests/test_callback_context_preservation.py
..                                                                       [100%]
2 passed
```

**"Establish that the result is callable."** That was the real defect, and the shape check could never do it: `evaluate(() => compute())` and `memo(() => paint())` are the same tree, and one returns a number. The **name** is what establishes it, so unwrapping is now limited to wrappers documented to return a component (`forwardRef`, `memo`, `observer`, `withRouter`, `withStyles`, `withTheme`, `connect`, `styled`, `inject`, `withTranslation`, `withErrorBoundary`), resolved through `React.memo` and the curried `styled(Base)(fn)` / `connect(map)(fn)` forms. A wrapper outside the set is simply not a definition here — which is what the parser did before wrapped components were indexed at all, so the failure direction is a component that stays unindexed rather than a function that does not exist.

**"Keep wrapper initialisation in its enclosing scope."** The component's body walk is now the wrapped function alone, and every wrapper in the chain is recorded against the enclosing function through `_extract_calls` rather than by walking the node, which would descend into the wrapped function again and credit its calls to the enclosing function as well:

```
function setup() { const Card = memo(forwardRef(() => paint())); return Card; }

  CALLS  setup -> memo
  CALLS  setup -> forwardRef
  CALLS  Card  -> paint
```

**"Preserve every declarator."** The caller skips its generic recursion over the whole declaration as soon as this pass extracts anything, so a declarator it does not own was being dropped. Those are now walked here, in the enclosing scope:

```
const Button = memo(() => paint()), token = nextToken();

  CALLS  setup  -> memo
  CALLS  setup  -> nextToken
  CALLS  Button -> paint
```

**One of my own tests was the defect written down.** `test_a_wrapped_component_keeps_the_wrapper_calls` asserted `("Card", "memo")` — the component credited with a call made before it existed. At module scope that reads as harmless; inside a function it is exactly the misattribution you found. It now puts the declaration inside `setup` and asserts the wrappers belong to `setup`, with a comment saying what it used to claim.

I added the two cases to `tests/test_js_wrapped_functions.py` rather than committing your file, so the branch does not collide with `integration/token-efficiency-hardening` when both land; your version is what I ran to verify.

```
pytest -q tests/         2997 passed, 9 skipped, 2 xpassed
```

tirth8205:

pytest (py3.13, merged origin/staging + PR head 1d0dbb6): "3233 passed, 9 skipped, 2 xpassed, 1 warning in 115.23s". ruff: "All checks passed!". mypy 2.3.0: "code_review_graph/parser.py:8827: error: Need type annotation for "chain" (hint: "chain: list[<type>] = ...") [var-annotated] / Found 1 error in 1 file (checked 72 source files)". Same error reproduced with mypy==1.15.0, so it is not version-dependent.

To fix:
- Fix mypy: annotate `chain: list = []` (or `list[Any]`) at code_review_graph/parser.py:8827 - CI type-check job fails as-is.
- Test for member-expression callee `React.memo(fn)` / `React.forwardRef<T, P>(fn)` (the member_expression branch in _js_wrapper_name is untested).
- Test documenting behaviour of `memo(Component, arePropsEqual)` (documented React API, currently NOT indexed - either support it for memo or add a test pinning the limitation and mention it in the PR body).
- Test for the failure branch of nested unwrapping: `memo(wrap(fn))` with an unknown inner wrapper, and for wrapper depth > _JS_WRAPPER_MAX_DEPTH.

L4XB:
Thanks for the list. I merged `origin/staging` first (`73053ca`), so the branch is up to date with `f0e4eb7`. Everything below is at `d26609b`.

**mypy.** `3b690d8` annotates the chain as `chain: list = []`, which after the merge sits at `code_review_graph/parser.py:10083`. `mypy code_review_graph/ --ignore-missing-imports --no-strict-optional` finds no issues in 75 source files, with both mypy 2.3.1 and 1.15.0.

**Member-expression callees.** `7fcf2a3` adds `test_a_member_expression_wrapper_is_unwrapped`. It checks three things for `React.memo(fn)` and `React.forwardRef<HTMLInputElement, Props>(fn)`: both are unwrapped, the parameters and line range come from the wrapped function and the declaration, and both wrapper calls stay on the enclosing function.

**`memo(Component, arePropsEqual)` is now supported** (`558690e`). `memo` also accepts the comparator as a second argument; every other wrapper still takes exactly one. The comparator is not part of the component, so calls in an inline comparator and the reference to a named one stay in the enclosing scope. `forwardRef(fn, extra)` and `memo(fn, eq, extra)` remain plain calls, and both are tested (`d886d0b` adds the second).

**Failure branches of nested unwrapping** (`1bad097`). `memo(wrap(fn))` with an unknown inner wrapper is not a definition, and every call stays on the enclosing function. The same goes for wrappers nested one level past `_JS_WRAPPER_MAX_DEPTH`. The test also checks the boundary itself, and that each wrapper call is recorded once.

While comparing the branch with staging edge by edge, I found and fixed a few more cases:

- `58d79eb`: the inner call of a curried wrapper (`styled(Base)(fn)`, `connect(mapState, ...)(fn)`), with the calls and references in it, and the type arguments of `forwardRef<El, Props>(fn)` are walked in the enclosing scope again. Before this, the branch lost the `styled`/`connect` calls and the references to `Base`, `mapState` and `Props`.
- `50c3308`: `const Card: FC<Props> = memo(...)` keeps its `Props` reference, as on staging. A directly assigned function still does not get one, again as on staging, and `d26609b` pins that.
- `84c4d2d`: tree-sitter lists comments among a call's arguments, so a `// eslint-disable-next-line` line above the function made `forwardRef(fn)` look like a two-argument call. Comments are no longer counted. In the corpus below, this indexes one real component, LibreChat's `OriginalDialog.tsx`.
- `3c4ad67`: pins that the other declarators of a statement are also walked in the enclosing scope when a function is assigned directly. Staging lost `y` in `const a = () => x(), b = y()`.

**Corpus differential.** I parsed 2058 React files from cherry-studio, dify, LibreChat and grafana with staging and with the branch.

- **Before these commits:** the branch lost 219 staging edges (218 `REFERENCES`, 1 `TESTED_BY`).
- **Now:** it loses one, the `TESTED_BY` edge in grafana's `validators.test.tsx`. That test renders `<Component />` inside `memo(() => ...)`. The call now belongs to the new `wrapped` component rather than to the test, and `TESTED_BY` follows the call, exactly as staging does when the function is assigned directly.
- **Other differences:**
  - 644 of the new nodes are wrapped components; calls moved into them or now resolve to them.
  - Two of those are named `TestRunMenu` and `TestInfoTab`. The existing name rule makes them `Test` nodes, as it does for a directly assigned function, and 36 `TESTED_BY` edges point at them.
  - The other 62 new nodes are in two minified Monaco bundles under `dify/web/public/vs`. There the branch now walks the other declarators of a statement that assigns a function directly.

**One correction to my comment of 13 Sep.** I wrote that the failure direction is a component that stays unindexed rather than a function that does not exist. That is not true for member callees. The wrapper name is the last segment of the callee, so a non-React call with the same name that takes one callback, such as `pool.connect(cb)` or `cache.memo(fn)`, is also read as a wrapper and becomes a `Function` node. It did not occur in the corpus: all 644 wrapped declarations there are `memo`, `forwardRef`, `React.memo`, `React.forwardRef`, `ReactModule.forwardRef` or `withErrorBoundary`. Restricting member callees to `React.*` would remove the false positive, but it would also miss namespace imports such as `ReactModule`. The PR body describes this, and I have left the choice to you.

**Checks, as CI runs them.**

- Python 3.10:
  - `ruff check code_review_graph/` and `ruff format --check` on the test file: clean.
  - bandit: no issues.
  - `pytest --tb=short -q -m "not browser and not upgrade" --cov=code_review_graph --cov-report=term-missing --cov-fail-under=65`: 3983 passed, 728 skipped (715 of them in the opt-in suites), 46 deselected, 2 xfailed, 2 xpassed; coverage 86.39%.
- Python 3.13: `tests/test_js_wrapped_functions.py` plus your `tests/test_callback_context_preservation.py`: 23 passed.

tirth8205:
The PR drops 63 TESTED_BY edges on material-ui across 22 test files, not the one edge in grafana's `validators.test.tsx` that the body reports, and three production symbols lose all of their test coverage in the graph.

Build a repo from material-ui's own `packages/mui-utils/src/useForkRef` and `getReactElementRef`, commit it, then run the CLI in a staging worktree and in this branch:

```
# staging
$ uv run code-review-graph build --repo $R
7 files, 31 nodes, 281 edges
$ uv run code-review-graph query --repo $R tests_for "$R/src/useForkRef/useForkRef.ts::useForkRef"
ok, n=2
  it:forks if only one of the branches requires a ref@L33
  it:does nothing if none of the forked branches requires a ref@L55

# this PR
7 files, 33 nodes, 275 edges
Found 0 result(s)
```

Both tests build their probe as `React.forwardRef(function Component(props, ref) { ... useForkRef(handleOwnRef, ref) ... })` inside the `it()` body. The call moves off the test onto the new component node, and TESTED_BY follows the call. Across all 27,701 JS/TS files of a fresh material-ui clone, TESTED_BY goes from 41,668 to 41,605, and all 22 per-file deltas are negative: `Slide.test.js` -10, `Grow.test.js` -8, `useForkRef.test.tsx` -8, `RadioGroup.test.js` -7, `FocusTrap.test.tsx` -6. `useForkRef`, `getReactElementRef` and `useSlot` drop to zero. The 23 names that gain coverage are all fixtures the PR itself invented: `FakeDiv`, `RealDiv`, `BrokenButton`. `tests_for` is a documented workflow, and no test in the PR parses a test file at all.

The false positive you disclosed is also a wrong answer:

```js
// src/db.js
import { logErr } from './log';
export function handle(pool) {
  const session = pool.connect(function (err) { logErr(err); });
  return session;
}
```

`query callers_of logErr` answers `src/db.js::handle` on staging and `src/db.js::session` here, a local variable holding a connection, and `handle` leaves the caller set entirely. `dead-code` on the same repo reports two dead symbols, `App` and `session`, where staging reports one. Same shape for `const sock = connect(function () { onOpen(); })` and `const cached = utils.memo(() => compute())`. It is rare, 0 new lowercase-named Function nodes across 33,454 files from express, socket.io, strapi and material-ui, but it is silent when it hits.

`_js_wrapper_name` recurses through curried callees with no depth cap, unlike `_js_wrapped_function`. `const S = connect(a)(b)(b)...(() => { paint(); });` first fails at 992 curried steps with `RecursionError: maximum recursion depth exceeded`, repeating `parser.py:10140`. Staging parses the same file and returns one edge. Not realistic source, but it is a crash where there was none.

All four items from the last round are there and they are load-bearing. mypy is clean on 77 files on the merged tree. I mutated the new suite rather than trusting the count: relaxing the `_JS_COMPONENT_WRAPPERS` name check fails 3 tests, dropping the `unowned` sibling-declarator walk fails 3, walking the declarator instead of the wrapped function fails 7. On 5,910 files from excalidraw and strapi the branch adds 97 Function nodes, every one a genuine component, with zero net target-level edge loss and parse time inside run-to-run noise. The branch is behind staging only because staging moved, and it merges without conflict.

To land it:

- Keep calls made inside a wrapped component defined in a test body attributed to the enclosing test, and add a test that parses a test file.
- Restrict member callees to `React.*`. Losing `ReactModule.forwardRef` namespace imports is the better trade than turning `pool.connect(cb)` into a function.
- Apply `_JS_WRAPPER_MAX_DEPTH` in `_js_wrapper_name` too.

L4XB:
@tirth8205 You're right, and the body understated it. One edge in my corpus was one probe in grafana, not the pattern; material-ui builds its probes inside `it()`, and I didn't have it in the set. The description is corrected, and all three items are in `a6d1d4d` and `f20ec1d`:

- **Probes built in a test body stay the test's.** A wrapped component whose enclosing function is a test (`it`, `describe`, ...) is not indexed, so its calls stay attributed to the test; one at module scope in a test file still is. The new test parses a test file shaped like `useForkRef.test.tsx` and checks both the `TESTED_BY` edges and `tests_for` end to end through `full_build`.
- **Member callees are `React.*` only.** `pool.connect(cb)` and `utils.memo(() => compute())` stay plain variables, and `logErr` / `compute` keep `handle` / `open` as their caller. `ReactModule.forwardRef` is no longer indexed, the trade you chose.
- **`_js_wrapper_name` stops at `_JS_WRAPPER_MAX_DEPTH`.** Your `connect(a)(b)…` chain at 1000 steps parses instead of raising `RecursionError`.

Your repro, on material-ui's own `useForkRef` and `getReactElementRef` files, with each tree in its own environment (I checked `__file__` from outside the clone):

```
staging 6b12d11       7 files, 31 nodes, 281 edges   tests_for useForkRef: n=2
previous head 5635db5 7 files, 33 nodes, 275 edges   tests_for useForkRef: n=0
current head          7 files, 31 nodes, 281 edges   tests_for useForkRef: n=2
```

The full suite passes (4263 passed, 858 skipped). `ruff check code_review_graph/` is clean, and your CI's mypy command reports only an `embeddings.py` error in a file this PR doesn't touch. The two new parser tests fail before `f20ec1d`, and the test-file test fails without `a6d1d4d`.

PR #973, opened 2026-09-11
Fix dynamic ignored-directory watcher replanning and degraded status

### LINKED ISSUE #908: watch: ignored trees created after startup are watched recursively, re-creating #811's exposure

## What happens

`_plan_watch_subtree` calls a directory "clean" when the ignored trees under it weigh less than `_WATCH_SPLIT_MIN_DIRS` (4 directories), and a clean directory is covered by one **recursive** watch (`code_review_graph/incremental.py:1665`). The planner runs once, at startup.

So the split that keeps `node_modules` and `target/` out of the kernel's hands only happens if those trees are already big when the watcher starts. Start `watch` on a fresh clone and then run `npm install`, and the new `node_modules` grows underneath a recursive watch on its parent — the resource-exhaustion path #811 is about.

The same applies to the budget fallback: when no plan fits, the root is watched recursively (`:1703`), and everything created later is inside it.

## Why it matters

The release notes say ignored trees are never registered with the OS. That is true for a repository whose ignored trees already exist at startup, and false for one where they arrive afterwards — which is the normal order for CI, containers, and a fresh clone.

## Suggested fix

Re-plan when a shallow watch's children change materially: `sync_watches` already scans the children of every non-recursive watch each tick, so a recursive watch whose subtree has grown an ignored directory above the split threshold could be released and re-planned. Failing that, document the limitation and log once when a recursive watch is installed over a directory that has no ignored children *yet*.

## Related

- The startup budget fallback never sets `degraded = True`, so `crg-daemon status` reports `ok` where `docs/TROUBLESHOOTING.md:189` says it should report degraded coverage. Only the runtime promotion path sets that flag.

Bakul2006:
### Linked issue

Closes #908 

## What & why

Fixes watcher planning when ignored directories are created or grow after the initial watch plan is installed.

Previously, the watch plan was calculated at startup. If an ignored directory such as `node_modules/` or `target/` did not exist at startup, its parent could receive a recursive watch. If the ignored directory was created later and grew beyond the existing split threshold, the watch plan was not reconsidered, allowing the ignored tree to fall under the recursive watch and potentially contribute to watcher/resource exhaustion.

This PR:

* Marks the watch plan dirty when directory changes can affect the ignored-tree split.
* Detects ignored-tree threshold crossings during shallow-watch synchronization.
* Releases obsolete watches before installing the recalculated plan.
* Preserves the existing watch adoption and budget behavior.
* Marks startup budget fallback as `degraded`, so `crg-daemon status` correctly reports degraded coverage.

Regression tests were added for dynamically created/growing ignored directories and the startup degraded-status behavior.

## How it was tested

The following validations were run successfully:

```bash
python -m py_compile code_review_graph/incremental.py
git diff --check
```

Language/static diagnostics reported no errors.

The full pytest/lint/type-check commands could not be run in the current environment because `pytest`, `uv`, and `networkx` are unavailable.

```bash
```

These commands should be run in a fully provisioned development environment before merging.

## Checklist

* [x] Tests added for new functionality
* [ ] All tests pass: `uv run pytest tests/ --tb=short -q`
* [ ] Linting passes: `uv run ruff check code_review_graph/`
* [ ] Type checking passes: `uv run mypy code_review_graph/ --ignore-missing-imports --no-strict-optional`
* [x] Lines are at most 100 characters
* [ ] Docs updated where behavior changed (README, `docs/`, docstrings)


Bakul2006:
Hey @tirth8205 just wanted to add on one note on validation: I added the regression tests, but I wasn't able to run the full pytest suite locally because the current environment is missing `pytest`, `uv`, and `networkx`.

The changes do pass `py_compile`, language diagnostics, and `git diff --check`.

Could you please run the full test suite in the project environment to verify the new watcher regression tests and the existing tests?

tirth8205:
Changes required: pending replans are never processed when the watcher has only a recursive root and no shallow watches. Run `python -m pytest tests/test_watch_robustness.py -k 'new_large_ignored_tree_replaces_recursive_root_watch or ignored_tree_growth_replans_recursive_root'`; both new regressions fail because `_replan_if_requested` is called only inside the empty shallow-watch loop. Process that request independently and retain existing coverage until replacement registration succeeds.

Bakul2006:
Hey @tirth8205  just did the same as you suggested me now the desired test are successfully passed

tirth8205:
008051b fixes the original recursive-root replan tests, but existing source coverage is still lost when replacement registration raises `OSError`. Reproduce by growing `node_modules` beside `src/deep/a.py` and failing `observer.schedule` for `src`, or by removing the ignored tree and failing registration of a recursive root; after `request_replan()` and `sync_watches()`, later deep edits are no longer delivered. Four new preservation tests, including two real PollingObserver cases, pass main and the integration branch but fail this head. `python -m mypy code_review_graph/ --ignore-missing-imports --no-strict-optional` also reports five new `Path`/`str` errors in `incremental.py`, so the source changes still require rework.

Bakul2006:
Hey @tirth8205 just addressed the failed replan coverage regression.

tirth8205:

```bash
# resolve, then
uv run pytest tests/ -q
git push
```

When resolving:
- Rebase the watcher/replan branch onto current origin/staging and resolve code_review_graph/incremental.py: staging (82ef9d0 'match ignore rules on repo-relative update inputs') changed sync_watches() to take an ignore_patterns kwarg and to compute newly_included via _should_ignore(...previous_patter...
- After the rebase, re-run the full gate set (pytest 3.13, ruff, mypy) and re-check that tests/test_watch_robustness.py still exercises the replan + degraded-status paths against the merged sync_watches() semantics.

tirth8205:
This no longer merges into `staging`. The one conflict is `code_review_graph/incremental.py`, and the branch is 144 commits behind.

```bash
git fetch origin
git merge origin/staging
# resolve, then
uv run pytest tests/ -q
git push
```

PR #976, opened 2026-09-13
Scope noninteractive file-operation guidance

woud420:
## Primary changes

The shell guidance recommends `cp -f`, `mv -f`, and `rm -rf` as universal defaults. Replace those examples with alias bypass and explicit path, collision, and authorization checks.

## Reviewer walkthrough

- Use `command cp`, `command mv`, and `command rm` for the documented ordinary-alias case.
- Example: copying a task artifact to a new destination bypasses an interactive alias; an existing destination still requires authorization to replace it.
- Scope force/recursive flags to necessary authorized operations, and keep noninteractive SSH authentication as a bounded example.

## Correctness and invariants

Avoiding a prompt does not grant permission to overwrite or delete. Unexpected collisions stop the affected operation while independent work continues. Software installation still requires task authorization.

## Testing and QA

- Executed all three documented forms with ordinary `cp`, `mv`, and `rm` aliases set to fail, in disposable fixtures under both Bash and Zsh: passed; the original source fixture remained intact.
- Reviewed the diff: only `AGENTS.md` shell-command guidance changed.
- `git diff --check`: passed.
- No user files were copied, overwritten, or deleted by this verification.

## Risk / Rollout

This strengthens permission-sensitive guidance and remains a draft for review. Alias bypass is not an overwrite safeguard by itself; the explicit destination and authorization checks are part of the instructions. No runtime code changes.

Resolves ENG-445


tirth8205:
This no longer merges into `staging`. The one conflict is `AGENTS.md`, and the branch is 224 commits behind.

```bash
git fetch origin
git merge origin/staging
# resolve, then
uv run pytest tests/ -q
git push
```

PR #996, opened 2026-09-15
ci: automate staging to testing, gate main behind the maintainer

tirth8205:
## What this changes

Promotion to `testing` becomes automatic. Promotion to `main` becomes impossible for any automation.

```
PR -> staging        CI green                    (you merge)
staging -> testing   automatic, fast-forward     (no pull request)
testing gauntlet     full matrix + e2e + security + coverage gate
   not green   ->    issue says what tests to add on staging, loop repeats
   green       ->    issue says testing is ready, gives the command
testing -> main      pull request, no bypass for anyone               (only you)
main -> PyPI         tag                                              (only you)
```

## Files

| File | Purpose |
| --- | --- |
| `.github/workflows/promote-to-testing.yml` | Fast-forwards `testing` after CI passes on `staging`. Refuses if the commit is no longer the tip of `staging`, if `testing` already has it, or if `testing` has diverged. |
| `.github/workflows/testing-gauntlet.yml` | Full suite on 3.10 to 3.13, e2e on three operating systems, bandit, coverage gate, then one verdict issue. |
| `scripts/test_gap_gate.py` | Runs `detect_changes` against `origin/main` and fails when a changed function has no test. Writes a markdown report naming each one. |
| `.github/workflows/promote.yml` | Deleted. It could open the pull request into `main`; that is now a manual step. |
| `CONTRIBUTING.md`, `CLAUDE.md` | Describe the pipeline as it now works. |

## Requires one manual setting

`testing` currently blocks every writer. The promotion workflow needs to push to it:

Settings, Rules, Rulesets, `testing-protection`, Bypass list, Add, "GitHub Actions" (Integration), mode Always.

`main` keeps an empty bypass list. That is deliberate and should stay that way.

## How it was tested

```
uv run pytest tests/ -q                       3226 passed, 9 skipped, 2 xpassed
uv run ruff check code_review_graph/ scripts/ All checks passed
python scripts/test_gap_gate.py --base origin/main   exits 0, reports the counts
```

Both workflow files parse as YAML. The gauntlet's end-to-end job treats pytest exit 5 (marker selected nothing) as a pass, so it is safe before every e2e branch has landed.


tirth8205:
Updated to `9abe8f2`: merged `ci/gate-accuracy` (#1000) and turned the gauntlet's coverage job from a blocker into a report.

**Why this branch could not merge as it stood.** Measured on the real `staging`-to-`main` delta (135 changed files, 639 changed nodes), `scripts/test_gap_gate.py` reported **212** changed functions with no test. 82 of those rows were wrong:

- **80 were test code** — 36 `TestX` classes (`parser.py` never set `is_test` on any Class node) and 44 `conftest` fixtures and test helpers.
- **2 had tests.** `changes.py` looked up `TESTED_BY` only under the node's own qualified name, so an edge left under a bare source read as no coverage. `detect_changes_func` itself was on the list, with three tests, because `tests/test_changes.py` reaches it through the `code_review_graph.tools` re-export.

A further 16 rows were closures — `walk`, `record`, `bind`, `sync_tested_by` — which nothing outside their enclosing function can call. Same tree, same base, same graph build, only the analyser swapped: **212 → 114, with no row added**.

Three other defects in the script are fixed: it printed `?` for every location (it read `file_path`/`line` from rows carrying `file`/`line_start`/`line_end`); it rendered 50 rows with a footer computed from a 25-row default; and **it failed open** — with the database deleted it printed `changed_files=135 changed_functions=0 test_gaps=0` and exited 0. It now refuses to answer unless every changed file the indexer collects is in the graph at the hash it has on disk.

**Why the job is now advisory.** Its rows are accurate about what they measure: 14 of 14 hand-checked flagged functions have no test that calls them. But "no test calls this" is not "this is untested" — 47 of the remaining 114 rows name helpers a test does execute two calls away (`_parse_vue` has no test of its own and runs in every Vue parsing test). Failing a promotion on a list that is 41% "run but not directly tested" teaches people to override the gate, which is worse than not having one.

So the verdict now turns on the suite, the end-to-end runs and the security scan. The coverage number and its report still appear on the tracking issue and in the checks list, marked advisory, with `indirect` on the rows a test does reach. A missing or stale graph now reports "the analysis could not run" instead of "no gaps". To enforce it later: drop `continue-on-error` from the `test-gaps` job and require its result in the verdict.

Trivially satisfiable tests — `@pytest.mark.skip`, or `def test_x(): my_func()` with no assertion — are detected and listed in their own section, and `--strict-assertions` makes them fail. They do not block by default because "this test asserts nothing" is a source-text heuristic that would misjudge a test delegating its assertions to a helper. Across 3051 test nodes here it finds 2 unconditionally skipped and 20 assertionless, nearly all language fixtures. `skipif`/`skipUnless` are deliberately not counted.

Gates on the merged branch: `pytest tests/ -q` 3361 passed / 9 skipped / 2 xpassed, `ruff check code_review_graph/` clean, `mypy` clean, coverage 85.5%.

tirth8205:

PR #997, opened 2026-09-15
fix(python): resolve relative imports to the module file, not the imported symbol

tirth8205:
> **Updated after three rounds of review.** A blocker (`forget` no longer
> matching a rebuild on the relative-import path) and five further defects
> were fixed on this branch in `173ad4b`; the CI-guard claim below was
> inverted and is corrected in place. The third round found that the
> precision gain never reached `callers_of` — fixed in `74e626c`, measured
> under "The read path had to change too", with four further disclosures
> under "Disclosures" and one mislabelled row corrected in place. See the
> review-response comments on this PR for the full list.

## The bug

`from .registry import Registry` produced an `IMPORTS_FROM` edge to
`<repo>/code_review_graph/Registry.py` — a file that exists on no filesystem
that tells `R` from `r`. `from .cli import main` produced
`<repo>/code_review_graph/main.py`: a real file, and the wrong one. **0 of 222
relative imports in `code_review_graph/` had a correct target** (207 wrong, 15
emitted no edge at all), and the same confusion in a second place meant **0 of
37 cross-module calls through a relatively imported name resolved to anything**.

tree-sitter-python wraps the module half of a relative import in a
`relative_import` node, never a `dotted_name`. Verified against the repo's own
parser:

```
from .m import a      -> [from, relative_import('.m'), import, dotted_name('a')]   module_name = '.m'
from pkg.m import a   -> [from, dotted_name('pkg.m'),  import, dotted_name('a')]   module_name = 'pkg.m'
```

`_extract_import` (`parser.py:15657`) and `_collect_import_names`
(`parser.py:13484`) both scanned for the **first `dotted_name`**. For an
absolute import that genuinely is the module — which is why absolute imports
were never affected. For a relative import the loop skips `relative_import`
and the first `dotted_name` it reaches is the imported **symbol**. The leading
dots, the only carrier of the package level, were discarded with it, and
`from .m import a as b` / `from .m import *` matched neither branch and emitted
nothing.

Downstream, `_do_resolve_module`'s Python branch had no `level` parameter at
all: it did `module.replace(".", "/")` and walked from the caller's directory
to the **filesystem root**, stat-ing as it went. That walk is what turned the
symbol `Registry` into `Registry.py` (APFS matched `registry.py`
case-insensitively) and the symbol `main` into the sibling `main.py`.
`python_resolver.py`'s post-build recovery could not help: it looks a bare
target up in a dotted-module index, and a symbol name is never a module name —
hence `imports_resolved: 0` on 1138 stamped edges.

## The fix

* Both sites now read `node.child_by_field_name("module_name")`, which is
  correct for the relative and the absolute form alike.
* Relative modules resolve through `_resolve_python_module_in_repo` — which
  already honoured the leading-dot level and already refused to leave the
  repository, and which `_expand_python_star_imports` was already calling two
  hundred lines away in the same file.
* `_collect_import_names` stores the **resolved file** for a relative import.
  `_resolve_imported_symbol` already accepts a path there, and star-import
  expansion already writes one into the same map. This is what makes
  cross-module `CALLS` resolve; fixing only the edge target would have left
  0/37.
* `from . import incremental` and `from .tools import query` additionally emit
  an edge to the submodule they name, because that is the dependency the
  statement creates. `from .graph import A, B` stays **one** edge — the
  statement imports from one module however many names it binds — so edge
  counts do not inflate (+0.25% on this repo, +1.0% on flask).
* The absolute walk-up now stops at the repository root instead of the
  filesystem root: a match above it is another checkout or a site-packages
  copy, not this repository's module.
* Python module probing compares against the real directory listing (cached
  per directory), so a case-insensitive filesystem can no longer answer "yes"
  to a spelling that does not exist.

### Convention for the target, stated explicitly

One edge per **module actually imported**:

| form | target(s) |
| --- | --- |
| `from .m import a` / `a, b` / `a as b` / `*` | `<dir>/m.py` — one edge |
| `from . import m` | `<dir>/__init__.py` **and** `<dir>/m.py` |
| `from .pkg import mod` (mod is a submodule) | `<dir>/pkg/__init__.py` **and** `<dir>/pkg/mod.py` |
| `from .pkg import Thing` (not a submodule) | `<dir>/pkg/__init__.py` |
| `from .m import a`, where **both** `m.py` and `m/__init__.py` exist | `<dir>/m/__init__.py` — the package, as CPython binds it |
| `from .ns import leaf`, where `ns/` has no `__init__.py` (PEP 420) | `<dir>/ns/leaf.py` **only** — a namespace package has no file to name |
| `from ..pkg.m import a` | `<dir>/../pkg/m.py` |
| unresolvable / outside the repo root | the bare module string, never a path |
| `from pkg.m import a`, `import x`, `import x.y as z` | unchanged |

## Before / after

Measured with **`scripts/audit_python_imports.py`** (new, in this PR), which
takes its ground truth from CPython's own `ast` — every `from … import …` is
re-resolved with real import semantics and compared against the edge the parser
wrote at that line. Both columns are the same script over two worktrees,
`origin/staging` (81a1fac) and this branch.

### `code_review_graph/` (this repository)

| measurement | before | after |
| --- | --- | --- |
| relative imports with the correct existing file as target | **0 / 222 (0.0%)** | **222 / 222 (100%)** |
| — of which wrong target / no edge at all | 207 / 15 | 0 / 0 |
| absolute repo-local imports correct (no regression) | 31 / 31 | 31 / 31 |
| cross-module CALLS through a relative import, resolved | **0 / 37 (0.0%)** | **37 / 37 (100%)** |
| Python `IMPORTS_FROM` edges naming a repository file | 670 / 1808 (37.1%) | 961 / 1903 (50.5%) |
| all edges whose target matches no node | 21952 / 54549 (**40.2%**) | 20760 / 54687 (**38.0%**) |
| — `IMPORTS_FROM` alone | 1300 / 1998 (65.1%) | 1097 / 2093 (52.4%) |
| resolved targets naming a file that does not exist | **7** | **0** |
| `imports_resolved` in the build log | 0 | 0 (see below) |
| build wall time, median of 3 cold builds | 3.64 s | 3.66 s |

### `pallets/flask` (`git clone --depth 30`, 94 files)

| measurement | before | after |
| --- | --- | --- |
| relative imports with the correct existing file as target | **0 / 167 (0.0%)** | **167 / 167 (100%)** |
| — of which wrong target / no edge at all | 118 / 49 | 0 / 0 |
| cross-module CALLS through a relative import, resolved | **11 / 67 (16.4%)** | **43 / 67 (64.2%)** |
| Python `IMPORTS_FROM` edges naming a repository file | 118 / 599 (19.7%) | 285 / 663 (43.0%) |
| all edges whose target matches no node | 3870 / 8377 (**46.2%**) | 3660 / 8458 (**43.3%**) |
| — `IMPORTS_FROM` alone | 503 / 604 (83.3%) | 383 / 668 (57.3%) |
| resolved targets naming a file that does not exist | **17** | **0** |
| `imports_resolved` in the build log | 55 | 55 |
| build wall time, median of 3 cold builds | 0.87 s | 0.88 s |

Reproduce either column:

```bash
uv run --python 3.13 code-review-graph build 2>&1 | tee build.log
uv run --python 3.13 python scripts/audit_python_imports.py \
    --repo . --package code_review_graph --build-log build.log
```

### Two numbers that need explaining rather than celebrating

**`imports_resolved` stays at 0 on this repo.** That counter reports only what
the *post-build* suffix index in `python_resolver.py` recovered. Before, it was
0 because a symbol name is never a module name, so nothing could be recovered.
After, it is 0 because there is nothing left to recover — resolution moved
earlier, to parse time, where the caller's package is actually known. Same
number, opposite meaning. The audit therefore also reports the number that
counter was a proxy for — Python `IMPORTS_FROM` edges that name a repository
file — which goes 37.1% → 50.5% here and 19.7% → 43.0% on flask. `imports_updated`
drops 1138 → 942 for the same reason: the relative edges no longer arrive
needing a fix.

**flask's cross-module calls reach 64%, not 100%.** The remaining 24 are calls
through an *attribute* of an imported object (`request.args.get(...)` on a
`LocalProxy` imported from `.globals`). `_resolve_call_target` only looks up the
root name of the call, so those were never in scope for this fix and are
unchanged by it.

**Edge-count growth** is the submodule edges: 54549 → 54687 (+0.25%) here,
8377 → 8458 (+1.0%) on flask. Every added edge names a real file.

## Tests

`tests/test_python_relative_imports.py` (31 tests) was written first and run
against `origin/staging`: **17 failed, and each failed for the diagnosed
reason** — `['GraphStore']` where the module file was expected, `[]` for the
`as`/`*` forms, `.../Registry.py` for the case-collision, `.../main.py` for the
sibling-shadowing case. It covers every form in the table above, plus the
repository-root clamp (`from ...escape import Z` must not emit a path), the
unresolvable case, absolute-import non-regression, and cross-module `CALLS`
for the plain, level-2, aliased and star forms.

Because `_extract_import`, `_collect_file_scope` and `_do_resolve_module` are
shared code paths, the same file pins unchanged behaviour per affected
language: TypeScript, JavaScript (import edge **and** call resolution), Java,
Kotlin, Go and Rust.

### CI guard

`tests/test_import_target_integrity.py` builds a **real graph** over a
repository laid out with every form that used to break — including a lowercase
`registry.py` imported as `Registry`, and a `cli.py`/`main.py` pair that
reproduces the sibling shadowing — then fails if any resolved `IMPORTS_FROM`
target names a file that does not exist, or escapes the repository root.

**Corrected after review — the earlier claim here was inverted.** It said the
guard is "meaningful on Linux (where the wrong spelling simply is not there)".
It is not: where the wrong spelling is not there, `origin/staging` never
*emits* it. Measured on a real case-sensitive APFS volume
(`hdiutil create -fs "Case-sensitive APFS"`), same fixture, same
`origin/staging` parser:

| | case-insensitive APFS (macOS default) | case-sensitive APFS (= Linux, = CI) |
| --- | --- | --- |
| `from .registry import Registry` target on `origin/staging` | `.../app/Registry.py` | `Registry` (bare, no path) |
| `missing_path_targets()` on `origin/staging` | **2 offenders — guard fails** | **[] — guard passes** |
| `audit_imports()` on `origin/staging` | 3/3 relative imports wrong | **3/3 relative imports wrong** |

So, precisely:

* `test_no_import_target_claims_a_file_that_does_not_exist` is a **macOS and
  Windows** guard. A bare module name makes no claim about the filesystem, so
  on a case-sensitive filesystem there is nothing for it to catch. CI's matrix
  is `ubuntu-latest`; this assertion does not prove the fix there.
* `test_every_relative_import_resolves_to_the_right_file` compares against
  `ast`-derived ground truth and therefore bites on **every** platform,
  including the Linux matrix. That is the assertion the CI matrix actually
  proves, and it is the one that also catches `from .cli import main` landing
  on the sibling `main.py` — a wrong target that exists on disk, which the
  missing-file guard cannot see on any filesystem.

The module docstring of `tests/test_import_target_integrity.py` now states
this split rather than the inverted claim.

`tests/import_audit.py` loads the audit script the way
`tests/test_action_render.py` loads `scripts/render_pr_comment.py`, so the CI
guard and the numbers in this table cannot drift apart.

## Existing tests that pinned the wrong behaviour

**None — no existing test was changed.** That is the finding, not a
convenience: `grep -rnF 'from .' tests/` returns exactly one hit in the whole
test tree (`tests/test_python_star_imports.py:74`), and
`tests/fixtures/sample_python.py` contains only `import os` and
`from pathlib import Path`. Python relative imports had essentially zero
`IMPORTS_FROM` coverage, and that absence — not a bad assertion — is what let
0/222 ship. `tests/test_parser.py::test_parse_python_aliased_module_imports`
asserts an exact set (`{"a.b", "b", "x", "y"}`) but only over `import`
statements, so it is untouched and still passes; the same expectation is now
re-pinned in the new file to keep it from drifting.

## Follow-up: attributing a method call to its receiver (commit `3bca5cf`)

A pre-existing bug in `graph.py::_resolve_bare_endpoints` was **amplified** by
the import fix above, and is now fixed here. The resolver bound a bare `CALLS`
target to any same-named node the call-site file imported, never consulting the
`receiver` already stored on the edge, so `some_dict.get(...)` was recorded as a
call into `ConnectionPool.get`. Correct relative imports widen the import
evidence, so this branch produced *more* of them: inbound CALLS to
`registry.py::ConnectionPool.get` went 27 (staging) -> 124 (this branch), where
the true number is 15.

The parser now records what the call-site file itself says the receiver is —
module binding, class (annotation / typed parameter / constructor assignment /
imported symbol), container literal, or unknown — scoped the way Python is
scoped, and the resolver restricts candidates to that. Container and unknown
attribute nothing: the target stays bare and the unresolved path explains it.
No method-name list is consulted anywhere.

### An earlier claim in this PR was wrong

The previous pass said "every withdrawn edge was a misattribution". It was not:
of the 59 edges withdrawn from the four named functions, **9 are genuine**.
Seven come back here (module-qualified `agent_baseline.run` / `.aggregate`,
each hand-verified against the source). The two `GraphUpdateHandler` ones do
not, because `handler = _create_watch_handler(...)` is an unannotated factory
and nothing in the file says what it holds — an honest unresolved beats a guess.

### Measurements, one fixed corpus built three times

The first two rows are **store** measurements: they count CALLS edges and
their distinct source nodes. An earlier revision of this table labelled the
second row "what `callers_of` returns", which it is not — `callers_of` has a
bare-name fallback that these rows do not model. The tool-level number is the
third row, and the section below is where it is measured.

| metric | staging | before this commit | after |
| --- | ---: | ---: | ---: |
| inbound CALLS to `ConnectionPool.get` | 27 | 124 | **15** (true: 15) |
| distinct source nodes of qualified CALLS edges naming `ConnectionPool.get` | 15 | 45 | **5** |
| — of those, true by hand | 5/15 | 0/15 | **5/5** |
| distinct callers `callers_of` actually returns (see "The read path" below) | 421 | 421 | **5** |
| `agent_baseline.py::run` / `::aggregate` | 44 / 8 | 0 / 0 | 3 / 4 |
| path-shaped CALLS targets matching no node | 155 | 84 | 84 |
| ...of those into `__init__.py` | 73 | 0 | 0 |
| resolved CALLS endpoints | 10633 | 11781 | 10407 |
| relative imports correct (this repo) | 0/222 | 222/222 | **222/222** |
| relative imports correct (fresh `pallets/flask`) | 0/167 | 167/167 | **167/167** |

### New metric: attribution precision

`edges with no matching node` only counts a target matching **no** node. It
cannot see a target matching the **wrong** node, and it scores an honest bare
"unresolved" exactly as badly as a confident lie. `audit_python_imports.py`
gains `audit_call_attribution`, which re-reads every attributed Python call
with CPython's `ast` and asks whether the source at that line gives any reason
for the node it was attributed to. The test is deliberately more permissive
than any resolver: a variable's class counts if its annotation or construction
appears anywhere in the file, in any scope, and package re-exports are followed
to their origin.

| corpus | staging | before | after |
| --- | ---: | ---: | ---: |
| this repo | 9235/10252 (90.1%) | 10264/11400 (90.0%) | **10022/10026 (100.0%)** |
| `pallets/flask` | 305/739 (41.3%) | 359/833 (43.1%) | **337/415 (81.2%)** |

This drops 1 132 of the 1 141 unjustified attributions while keeping 10 022 of
the 10 264 justified ones, ending with 787 more justified attributions than
staging. Honest limits: the metric sets aside edges where `ast` finds no call
with the target's leaf name at that line (1 here), and 4 of the 4 remaining
unjustified are two oracle blind spots (a `sys.path`-relative fixture import, a
`scripts/` import) plus two real residual misattributions from an aliased
`resolve = resolve_spring_di_calls`.

### Tests

`tests/test_python_receiver_evidence.py` — 14 cases pinning **exact** targets in
both directions: the module-receiver rule (absolute, relative, function-scoped
import, a same-named function in a sibling module, and `runner.start()` not
reaching into `Runner`), the do-not-attribute rule (dict literal, list literal,
unknown local, expression receiver, contradictory bindings), and the evidence
that does justify an attribution (constructor, annotation, `self.field`
constructed in another method, per-function scoping). No existing test was
weakened or changed.

## The read path had to change too (commit `74e626c`)

Making the graph honest is not the same as making the tool honest, and the two
had come apart. `tools/query.py::callers_of` keeps a bare-name fallback — a
`CALLS` edge whose target is the plain method name still answers for a node of
that name — and that fallback applies exactly the rule `_resolve_bare_endpoints`
had just been taught to refuse. Because the commit above deliberately converts
wrong *qualified* attributions into honest *bare* ones, the fallback re-absorbed
them. Measured through `query_graph` the way `main.py` calls it
(`detail_level="minimal"`, `max_results=500`, reading `result_count`), one fixed
corpus built three times:

| `callers_of` returns | staging | after `3bca5cf` | after `74e626c` |
| --- | ---: | ---: | ---: |
| `ConnectionPool.get` | 421 | 421 | **5** |
| `GraphStore.close` | 343 | 365 | **189** |
| `GraphUpdateHandler.start` | 14 | 29 | **0** |
| `GraphUpdateHandler.stop` | 3 | 10 | **0** |
| `_sanitize_name` (unique-name control) | 28 | 28 | **28** |

`GraphStore.close` was *worse* through the tool after `3bca5cf` than on staging:
the resolver withdrew 113 wrong qualified attributions — staging had resolved
`self.tmp.close()` on a `NamedTemporaryFile` and `conn.close()` on a
`sqlite3.Connection` into `GraphStore.close` — and the fallback took them
straight back as bare ones.

### Hand verification

Every call site below was opened and read. Where a list is longer than the
sample, the sample is stratified across it (head, quartiles, tail).

| true callers / inspected | staging | after `3bca5cf` | after `74e626c` |
| --- | ---: | ---: | ---: |
| `ConnectionPool.get` | 5 / 15 | 5 / 15 | **5 / 5** |
| `GraphStore.close` | 9 / 15 | 15 / 27 | **16 / 16** |
| `GraphUpdateHandler.start` | 1 / 14 (all) | 2 / 29 (all) | **0 / 0** |
| `GraphUpdateHandler.stop` | 1 / 3 (all) | 2 / 10 (all) | **0 / 0** |
| `_sanitize_name` | 28 / 28 | 28 / 28 | **12 / 12** sampled of 28 |

The false positives are what the rule predicts: `qn_to_idx.get(...)`,
`comm.get("level", 0)` and `n.get("kind", "")` for `ConnectionPool.get`;
`plt.close(fig)`, `os.close(devnull)`, `self._conn.close()`, `self.tmp.close()`
and `bare_edges.close()` (a `sqlite3.Cursor`) for `GraphStore.close`;
`thread.start()`, `observer.start()`, `daemon.start()`, `reaper.start()` and six
regex `match.start()` calls for `GraphUpdateHandler.start`.

### What it cost

`GraphUpdateHandler.start` and `.stop` go to **zero, and two of each were
real** — `incremental.py::watch` and
`tests/test_watch_rule_coverage.py::test_polling_observer_delivers_subsequent_deep_edits`,
both of which call `handler.start()` / `handler.stop()` on a receiver bound by
`handler = _create_watch_handler(...)` or `supervisor, handler = create_watch(...)`.
Neither is annotated, so nothing in the file says what `handler` holds, and the
evidence reads `unknown`. Admitting `unknown` would return those two — and also
return 700 more dict `.get()` calls, 179 more file-handle `.close()` calls and
33 more `thread.start()` calls, which is the blocker this commit exists to fix.
An honest zero with a `not-indexed` note beats 29 results with 2 right ones; the
fix for the real loss is inferring a return type for an unannotated factory,
filed as follow-up.

### The fix

`GraphStore.receiver_evidence_admits` runs the single candidate the caller asked
about through the resolver's own `_receiver_backed_candidates`, so there is one
rule in one place rather than two implementations to drift. An edge with **no
receiver at all** — a plain unqualified `foo(...)` — is admitted unchanged,
because that is the case the fallback exists for.

Where the same fallback lives elsewhere, and what was done about each:

* **`callers_of`** — the `CALLS` bare-name fallback. **Changed**; this is the
  measured one.
* **`tests_for`** — `get_transitive_tests` has its own bare-**source** fallback.
  `TESTED_BY` is minted from the test's own `CALLS` edge with its metadata
  copied, so 5 347 bare-source `TESTED_BY` edges in this repo's graph carry
  receiver evidence and the fallback could claim `'  a  '.strip()` in a test as
  coverage of a module-level `strip`. **Changed.**
* **`references_to`** — shares `iter_edges_by_target_name`. **Changed for
  consistency, with no measured effect**: the parser stamps receiver evidence on
  `CALLS` only (`TESTED_BY` inherits it), and no `REFERENCES` edge in this graph
  carries any.
* **`get_impact_radius`** — **not changed, and no fallback to change**. Both
  engines (the SQLite best-score relaxation and the NetworkX one) traverse exact
  `qualified_name` joins, so a bare target such as `get` matches no node.
* `refactor.py`'s rename-preview and dead-code detection also call
  `search_edges_by_target_name`. Left alone deliberately: there a bare match
  keeps a symbol *out* of the dead-code list, so over-inclusion currently fails
  safe, and narrowing it is a behaviour change that wants its own measurement.
  Filed as follow-up.

### Tests

`tests/test_receiver_evidence_read_path.py` — 12 cases, all asserting through
`query_graph`, not through the store, because the store was already right.
**10 of the 12 fail on `3bca5cf`**, each for the diagnosed reason. Seven seeded
cases pin one edge per evidence kind (builtin receiver, matching class, other
class, module, unknown, and no receiver at all) plus the count `detail_level=
"minimal"` reports; five run a real parse and build of a small repository and
cover the dict-receiver case, the unannotated-factory case, the `tests_for`
fallback, and a uniquely named plain call as the control. The two that pass on
both commits are the positive controls — a matching class receiver and the
unique name must keep their callers, and they do.

## Disclosures

Four things the numbers above do not say on their own.

**The ledger is not one-way.** Against staging, `3bca5cf` adds 1 068
attributions, 1 067 of them justified, and withdraws 1 294, 1 014 of them
unjustified. Net: **+787 justified, −1 013 unjustified**. The headline
"attribution precision 90.1% → 100.0%" is that trade, not a pure gain.

**280 previously-correct attributions are withdrawn.** They are the difference
between the 1 294 withdrawn and the 1 014 unjustified ones, and they are the
price of refusing to guess at an `unknown` receiver. The same refusal shows up
in coverage: `TESTED_BY` edges whose source names an indexed node drop from
6 603 to 6 160, **−6.7%**, so some real test coverage is no longer reported.

**The attribution-precision denominator only counts edges that name a node.**
`audit_call_attribution` selects `CALLS` edges where `EXISTS (SELECT 1 FROM
nodes WHERE qualified_name = target)`. Withdrawing an attribution removes the
edge from the denominator instead of scoring against it, so a resolver that
attributes nothing at all scores 100%. The 90.1% → 100.0% figure is therefore
partly an artifact of the metric, and has to be read next to the ledger above
and next to `edges with no matching node`, which does count the withdrawals.

**`IMPORTS_FROM` edges lose an `extra` field, undocumented until now.** On
staging, a Python relative import that `python_resolver.py` could not place was
stamped `{"import_resolution": "unresolved", "python_module": "<symbol>",
"import_candidates": [], "import_candidate_count": 0,
"import_candidates_truncated": false}`. Those imports now resolve at parse time,
so the post-build resolver never sees them and their `extra` is `{}`: **197
`IMPORTS_FROM` edges in this repo** go from carrying those five keys to carrying
none, and the stamped total falls 1 159 → 962. Nothing in-tree reads
`import_resolution` except the ambiguity path in `_resolve_bare_endpoints`,
which only acts on `"ambiguous"`, but any external consumer of that field sees
the change.

## Gates

```
pytest tests/ -q                                 3306 passed, 9 skipped, 2 xpassed
ruff check code_review_graph/                    All checks passed
mypy code_review_graph/ --ignore-missing-imports --no-strict-optional
                                                 Success: no issues in 72 source files
bandit -r code_review_graph/ -c pyproject.toml   0 issues
```


tirth8205:
## Review response — `173ad4b`

Everything below is on this branch; the PR body's convention table, gates block
and CI-guard paragraph are corrected in place.

### Blocker: `forget` no longer matched a rebuild

Reproduced first, as a diff between "build, then `forget pkg/util.py`" and "build
the same repo with `pkg/util.py` deleted", on a package wired with **relative**
imports:

```
DANGLING AFTER FORGET: [('CALLS', '.../pkg/main.py::run', '.../pkg/util.py::helper')]
PARITY: False
  only-after-forget: ('CALLS', 'pkg/main.py::run', 'pkg/util.py::helper', ...)
  only-in-rebuild:   ('CALLS', 'pkg/main.py::run', 'helper',              ...)
```

Cause as diagnosed: `_collect_import_names` (and `_expand_python_star_imports`,
the same hole two hundred lines away) call `_resolve_python_module_in_repo`
directly, so the `exclude_files` filter that lives in `_resolve_module_to_file`
never runs. `tests/test_forget_parity.py` used `from util import helper` — the
absolute path, which was never broken.

Fixed by moving the exclusion check **into `_python_file_exists`**, the probe
every Python-specific resolver already shares, rather than adding a second
filter at each call site. Two consequences that the one-layer-up filter did not
have, and that parity needs:

* every Python caller inherits it, including ones added later;
* an excluded file makes resolution fall through to the **next** candidate,
  which is exactly what its real absence would do.

The process-wide star-export cache is bypassed while exclusions are in force —
its key is only the module's own mtime and size, so it would otherwise hand
`forget` an answer computed against a different set of files on disk.

`tests/test_forget_parity.py` gains `_RELATIVE_IMPORT_FILES` (covering
`from .util import helper`, `from . import util`, `from .sub.deep import
deep_fn`, `from .sub import deep`) and two tests that forget a module and a
sub-package module and diff the whole snapshot — nodes, edges, flows,
communities, risk index, embeddings — against a clean rebuild without the file.
Both **fail on `c4ce127`** and pass now.

### The five other findings

| # | finding | fix | test (all fail on `c4ce127`) |
| --- | --- | --- | --- |
| 1 | `pkg/m.py` probed before `pkg/m/__init__.py`, so the wrong file won | package before module, in the relative resolver *and* the absolute walk-up | `test_package_beats_a_same_named_module_file`, `test_absolute_import_also_prefers_the_package` |
| 2 | PEP 420 namespace packages unresolved — `from .ns import leaf` emitted `.ns` | submodules resolve to their files; the namespace package itself contributes no target | `test_namespace_package_submodule_resolves_to_its_file`, `test_namespace_package_emits_no_target_for_itself` |
| 3 | `from . import sub` → `REFERENCES` to `<package __init__>::sub`, matching no node | a module-bound name resolves to the module's **file node** | `test_reference_to_an_imported_subpackage_names_its_init` |
| 4 | `_extract_import` said `pkg/cli.py`, `_collect_import_names` said `pkg/__init__.py` | both now name the submodule file | `test_import_map_and_import_edge_agree_on_from_dot_import` |
| 5 | precision — see below | package re-exports followed to the defining file | `test_call_through_a_package_reexport_lands_on_the_defining_file` |

On (1), the skeptic's warning was right: **the audit script's own `_module_file`
had the same inverted order**, so ground truth agreed with the bug. Both are
corrected, and the tests take ground truth from the interpreter
(`importlib.machinery.PathFinder.find_spec`, which runs the same `FileFinder`
that `import` runs) rather than from the audit or from prose. Verified:

```
$ python3 -c "import pkg.m; print(pkg.m.__file__)"   #  pkg/m.py  +  pkg/m/__init__.py
.../pkg/m/__init__.py
$ python3 -c "import pkg.both; print(pkg.both.__file__)"  # pkg/both.py + pkg/both/ (no __init__)
.../pkg/both.py
```

The second line is why the namespace fallback is only reached after the file
probe fails, and it is pinned by `test_module_file_beats_a_same_named_namespace_directory`.

### (5) The "precision regression", investigated before acting

Two separate effects were bundled in that finding. They point opposite ways.

**a. The 155 → 239 path-shaped CALLS targets matching no node is not
`_resolve_bare_endpoints`.** It is `_resolve_imported_symbol`: correct relative
targets meant `from .tools import query_graph` now resolves to
`tools/__init__.py`, and then qualifies the symbol against it — but
`query_graph` is defined in `tools/query.py`. The package's own export map
already knows that, so it is now followed. Result on this repository:

| | staging `81a1fac` | PR `c4ce127` | now `173ad4b` |
| --- | --- | --- | --- |
| path-shaped CALLS targets matching no node | 155 | 239 | **105** |
| — of those, into an `__init__.py` | 73 | 130 | **0** |
| resolved CALLS endpoints (total) | 10 794 | 12 055 | **12 146** |

`tools/__init__.py::list_flows` (9 edges, matching nothing) became
`tools/flows_tools.py::list_flows` (11 edges, matching a real node);
`tools/query.py::query_graph` went 91 → 110.

**b. The four functions that lost `callers_of` should have lost them.** I
pulled the actual edges rather than the counts:

| function | staging | PR & now | what staging was counting |
| --- | --- | --- | --- |
| `agent_baseline.py::run` | 44 | 0 | `subprocess.run(...)` in `tests/test_eval.py` |
| `agent_baseline.py::aggregate` | 8 | 0 | `token_efficiency.aggregate()`, `impact_accuracy.aggregate()` |
| `GraphUpdateHandler.start` | 5 | 0 | `thread.start()`, `observer.start()`, `debouncer.start()` |
| `GraphUpdateHandler.stop` | 2 | 0 | `observer.stop()`, `handler.stop()` |

Each edge carries `{"bare_call_target": "run", "receiver": "subprocess"}` in
`extra`, so the receiver is right there in the row. The two smaller drops are
the same story: `tools/query.py:174` is `store.get_impact_radius(...)` —
`GraphStore`'s method, not the tool function — and
`tests/test_watch_robustness.py:848` is `cli.main()`, which is
`code_review_graph/cli.py`, not `daemon_cli.py`.

So the unique-candidate heuristic gave up **correctly**: the wider evidence set
found several plausible `run`s and refused to guess, where before it had one
candidate only because the real ones were invisible. I therefore did not narrow
the evidence set or rank candidates — that would restore false positives. 61
edges withdrawn against 1 423 gained versus staging (340 functions gain, 19
lose). Resolving the receiver against the import map would recover the honest
half of `aggregate`, but that is a different fix to a different function and
does not belong on this branch.

### Weak tests, tightened

* `test_no_import_target_escapes_the_repository_root` passed on anything,
  because nothing in the fixture tried to escape. The fixture now writes real
  bait **above** the repository root (`outside.py`, `outside_pkg/__init__.py`)
  and imports it with `from ...outside import THING` (from `app/`) and
  `from ....outside import THING` (from `app/sub/`). The test asserts those
  three statements produced edges — so the loop really examined them — and a
  companion test asserts the bait and the statements exist, so the guard cannot
  quietly go vacuous again.
* `test_this_repository_has_no_missing_import_targets` read the gitignored
  `.code-review-graph/graph.db`: nondeterministic locally, inert in CI, and
  self-skipping whenever the sources were newer. It now copies
  `code_review_graph/` into `tmp_path`, `git init`s, builds its own graph and
  asserts against that — plus a second assertion that all 222 relative imports
  in our own package resolve correctly. Module-scoped fixture; the whole file
  runs in 2.4 s.

### The Linux-matrix claim was inverted — body corrected, code unchanged

I mounted a case-sensitive APFS volume and ran the same fixture through the
`origin/staging` parser on both:

| | case-insensitive APFS | case-sensitive APFS (= Linux, = CI) |
| --- | --- | --- |
| `from .registry import Registry` target | `.../app/Registry.py` | `Registry` (bare) |
| `missing_path_targets()` | 2 offenders — **fails** | `[]` — **passes** |
| `audit_imports()` | 3/3 wrong | 3/3 wrong |

Where the wrong spelling does not exist, staging never emits it, so the
missing-file guard has nothing to catch. CI is `ubuntu-latest`, so that
assertion is a macOS/Windows guard; the assertion the Linux matrix actually
proves is `test_every_relative_import_resolves_to_the_right_file`, which
compares against `ast` ground truth and also catches `from .cli import main`
landing on the sibling `main.py` — a wrong target that exists, and that the
missing-file guard cannot see on any filesystem. The code is right; the body
and the test module docstring now say this instead of the inverted claim.

### Audit re-run — the headline numbers still hold

`scripts/audit_python_imports.py`, full build + post-processing, same script
over three trees.

**`code_review_graph/` (this repository)**

| measurement | staging `81a1fac` | PR `c4ce127` | now `173ad4b` |
| --- | --- | --- | --- |
| relative imports with the correct existing file | 0 / 222 | **222 / 222** | **222 / 222** |
| absolute repo-local imports correct | 31 / 31 | 31 / 31 | 31 / 31 |
| cross-module CALLS through a relative import | 0 / 37 | **37 / 37** | **37 / 37** |
| path-shaped `IMPORTS_FROM` targets missing on disk | 7 | 0 | 0 |
| all edges whose target matches no node | 40.2 % | 37.9 % | **37.6 %** |
| build wall clock, median of 5 interleaved runs | — | 6.42 s | 6.71 s (+4.5 %) |

**`pallets/flask`** (`git clone --depth 30`, fresh clone today)

| measurement | staging | PR | now |
| --- | --- | --- | --- |
| relative imports with the correct existing file | 0 / 167 | **167 / 167** | **167 / 167** |
| cross-module CALLS through a relative import | 9 / 61 | 40 / 61 | 40 / 61 |
| path-shaped `IMPORTS_FROM` targets missing on disk | 17 | 0 | 0 |

(flask's call denominator is 61 rather than the body's 67 because this is a
newer `--depth 30` clone; the relative-import numbers are unaffected.)

The +4.5 % build time is the re-export following in (5a): each imported module
is parsed once for its export map, memoised per parser instance and again
process-wide. Restricting it to `__init__.py` files brings it to +1.4 % and
leaves 110 dangling targets instead of 105; I kept the general form, since a
re-export is a re-export wherever it is written.

### Gates

```
pytest tests/ -q                                 3280 passed, 9 skipped, 2 xpassed
ruff check code_review_graph/                    All checks passed
mypy code_review_graph/ --ignore-missing-imports --no-strict-optional
                                                 Success: no issues in 72 source files
bandit -r code_review_graph/ -c pyproject.toml   0 issues (High/Medium/Low)
```

tirth8205:
## Correction, and the receiver fix

**The earlier claim in this PR that "every withdrawn edge was a misattribution" was wrong.** I checked the withdrawn edges against the source instead of asserting it. Of the 59 edges this branch withdrew from the four named functions, **9 are genuine calls**, most of them module-qualified:

| withdrawn from | staging | genuine | why |
| --- | ---: | ---: | --- |
| `agent_baseline.py::run` | 44 | 3 | `agent_baseline.run(...)` after `from ...benchmarks import agent_baseline`; the other 41 are `subprocess.run`, `token_efficiency.run`, `impact_accuracy.run`, … |
| `agent_baseline.py::aggregate` | 8 | 4 | `agent_baseline.aggregate(...)`; the other 4 are `token_efficiency.aggregate` / `impact_accuracy.aggregate` |
| `GraphUpdateHandler.start` | 5 | 1 | `handler.start()`; the rest are `thread.start()`, `observer.start()`, `debouncer.start()` |
| `GraphUpdateHandler.stop` | 2 | 1 | `handler.stop()`; the other is `debouncer.stop()` |

## The bug this exposed

`graph.py::_resolve_bare_endpoints` resolved a bare `CALLS` target by name as soon as the call-site file imported one same-named node. The `receiver` already stored on every edge was never consulted. So `some_dict.get(...)` was attributed to `registry.py::ConnectionPool.get`.

That bug is on staging. **This branch made it worse**, because correct relative imports widen the import evidence, so more bare calls find a "unique" candidate:

```
inbound CALLS to code_review_graph/registry.py::ConnectionPool.get
  origin/staging   27
  PR head         124     receivers: result 24, <none> 23, pool 15, health 10, entry 9, …
```

The `<none>` 23 are `os.environ.get(...)`: a member call whose receiver is an expression, which the parser recorded with no receiver at all, making it indistinguishable from a plain `get(...)`.

## The rule

The parser now records what the call-site file itself says the receiver is, scoped the way Python is scoped, and the resolver restricts candidates accordingly. **No method-name list anywhere** — every rule reads syntax in the file.

| evidence | from | candidates restricted to |
| --- | --- | --- |
| `module` | an import binds the name to a module resolving to an indexed file | top-level names of that file |
| `class` | annotation, typed parameter, `x = Klass(...)` / `self.x = Klass(...)` where `Klass` is defined or imported here, or the receiver naming an imported symbol | members of that class |
| `builtin` | assigned a container/scalar literal or a comprehension | nothing — no repo node owns the method |
| `unknown` | contradictory bindings, a factory return value, or an unnamed receiver expression | nothing |

`builtin` and `unknown` attribute nothing: the target stays bare and the existing unresolved path explains it. Scope matters — `store` is a `GraphStore` in fifteen test methods and a `MagicMock` in three, and only the enclosing function says which.

## Measurements

One fixed corpus (`git archive origin/staging`) built three times, so all three parse byte-identical source.

| metric | staging | PR head | this pass |
| --- | ---: | ---: | ---: |
| inbound CALLS to `ConnectionPool.get` | 27 | 124 | **15** (true: 15) |
| distinct callers of `ConnectionPool.get` | 15 | 45 | **5** |
| `agent_baseline.py::run` | 44 | 0 | 3 |
| `agent_baseline.py::aggregate` | 8 | 0 | 4 |
| `GraphUpdateHandler.start` / `.stop` | 5 / 2 | 0 / 0 | 0 / 0 |
| path-shaped CALLS targets matching no node | 155 | 84 | 84 |
| …of those into `__init__.py` | 73 | 0 | 0 |
| resolved CALLS endpoints | 10633 | 11781 | 10407 |
| relative imports correct (this repo) | 0/222 | 222/222 | **222/222** |
| relative imports correct (fresh `pallets/flask`) | 0/167 | 167/167 | **167/167** |

The 15 that remain on `ConnectionPool.get` are exactly the 15 `self.pool.get(...)` calls in `TestConnectionPool`, reached through `self.pool = ConnectionPool(max_size=3)`.

All 7 recovered `agent_baseline` edges hand-verified against the source: every one is `agent_baseline.<fn>(...)` under `from code_review_graph.eval.benchmarks import agent_baseline`.

The two `GraphUpdateHandler` edges do **not** come back. `handler = _create_watch_handler(...)` is an unannotated factory, so nothing in the file says what `handler` holds. An honest unresolved beats a guess.

### `callers_of` precision, by hand

15 distinct callers opened and judged against the source at each call site.

| | staging | PR head | this pass |
| --- | ---: | ---: | ---: |
| `callers_of(ConnectionPool.get)` true | **5/15** | **0/15** (of 45 returned) | **5/5** |

Staging's 10 false ones are `data.get("repos")`, `entry.get("alias")` ×3, `call_args[1].get("timeout")`, `spec.get("tool")` ×2, `preview.get("refactor_id")`, `result.get("truncated")`, `f.get("steps")`. Every one of the 15 sampled at the PR head is a dict `.get(...)` on an ordinary local or on `os.environ`.

### New metric: attribution precision

The headline metric (`edges with no matching node`) only counts a target matching **no** node. It cannot see a target matching the **wrong** node — the failure a reviewer actually meets, because a wrong node looks perfectly resolved. Worse, it scores an honest bare "unresolved" exactly as badly as a confident lie, which is why it moves the wrong way here (37.7% → 40.2%).

`scripts/audit_python_imports.py` now re-reads every attributed Python call with CPython's `ast` and asks whether the source at that line gives any reason for the node it was attributed to. The test is deliberately **more permissive than any resolver**: a variable's class counts if its annotation or construction appears anywhere in the file, in any scope, and package re-exports are followed to their origin.

| corpus | staging | PR head | this pass |
| --- | ---: | ---: | ---: |
| this repo | 9235/10252 (90.1%) | 10264/11400 (90.0%) | **10022/10026 (100.0%)** |
| `pallets/flask` | 305/739 (41.3%) | 359/833 (43.1%) | **337/415 (81.2%)** |

Read together with the row above it: this pass drops 1 132 of the head's 1 141 unjustified attributions while keeping 10 022 of its 10 264 justified ones, and ends with 787 more justified attributions than staging.

Honest limits of the metric: it sets aside edges where `ast` finds no call with the target's leaf name at that line (1 edge here), and 4 of the 4 remaining unjustified on this repo are two oracle blind spots (a `sys.path`-relative fixture import, a `scripts/` import) plus two real residual misattributions from `resolve = resolve_spring_di_calls` aliasing.

## Tests and gates

`tests/test_python_receiver_evidence.py`, 14 cases pinning **exact** targets in both directions — the module-receiver rule (absolute, relative, function-scoped import, a same-named function in a sibling module, and `runner.start()` not reaching into `Runner`), the do-not-attribute rule (dict literal, list literal, unknown local, expression receiver, contradictory bindings), and the evidence that does justify an attribution (constructor, annotation, `self.field` constructed in another method, per-function scoping).

- `pytest tests/ -q` — 3294 passed, 9 skipped, 2 xpassed. No existing test weakened or changed.
- `ruff check code_review_graph/` — clean.
- `mypy code_review_graph/ --ignore-missing-imports --no-strict-optional` — clean, 72 files.

tirth8205:
## Blocker fixed: the precision gain now reaches `callers_of`

The review was right, and the diagnosis was exact. `tools/query.py::callers_of`
falls back to `store.iter_edges_by_target_name(node.name)`, which answers from
every `CALLS` edge whose target is the bare method name — the attribution
`_resolve_bare_endpoints` had just been taught to refuse. Because `3bca5cf`
deliberately converts wrong *qualified* targets into honest *bare* ones, the
fallback took them straight back. The graph got more honest; the tool did not.

Fixed in `74e626c`.

### Measured through the tool

`query_graph` called the way `main.py` calls it (`detail_level="minimal"`,
`max_results=500`, reading `result_count`), on one fixed corpus — the `3bca5cf`
tree, 324 files — built three times, once per parser version.

| `callers_of` returns | staging | after `3bca5cf` | after `74e626c` |
| --- | ---: | ---: | ---: |
| `ConnectionPool.get` | 421 | 421 | **5** |
| `GraphStore.close` | 343 | 365 | **189** |
| `GraphUpdateHandler.start` | 14 | 29 | **0** |
| `GraphUpdateHandler.stop` | 3 | 10 | **0** |
| `_sanitize_name` (unique-name control) | 28 | 28 | **28** |

My absolute numbers for the first two rows differ slightly from the review's
(421 vs 409, 343/365 vs 320/342) because I held the *corpus* fixed at the
`3bca5cf` tree rather than building each branch's own tree — its extra test
files add `.get()` and `.close()` call sites. The deltas are identical:
`ConnectionPool.get` +0 and `GraphStore.close` +22 from staging to `3bca5cf`,
and `start` 14→29 and `stop` 3→10 reproduce exactly.

`GraphStore.close` being *worse* than staging was the sharpest part of the
report and it is real: the resolver withdrew 113 wrong qualified attributions
(staging had resolved `self.tmp.close()` on a `NamedTemporaryFile` and
`conn.close()` on a `sqlite3.Connection` into `GraphStore.close`) and the
fallback re-absorbed them as bare ones.

### Hand verification

Every call site opened and read; stratified across the list where it is longer
than the sample.

| true callers / inspected | staging | after `3bca5cf` | after `74e626c` |
| --- | ---: | ---: | ---: |
| `ConnectionPool.get` | 5 / 15 | 5 / 15 | **5 / 5** |
| `GraphStore.close` | 9 / 15 | 15 / 27 | **16 / 16** |
| `GraphUpdateHandler.start` | 1 / 14 (all) | 2 / 29 (all) | **0 / 0** |
| `GraphUpdateHandler.stop` | 1 / 3 (all) | 2 / 10 (all) | **0 / 0** |
| `_sanitize_name` | 28 / 28 | 28 / 28 | **12 / 12** sampled of 28 |

The review's "2 of 29 true" for `start` reproduces exactly, and I can name them:
`incremental.py::watch` (line 2816) and
`tests/test_watch_rule_coverage.py::test_polling_observer_delivers_subsequent_deep_edits`
(line 133). `grep -rn 'handler\.start'` over the repository returns those two
and nothing else. `stop` is the same two callers.

### What it costs, stated plainly

Those two true callers are now **lost**, and so are `stop`'s two. Both bind the
receiver through an unannotated factory — `handler = _create_watch_handler(...)`
and `supervisor, handler = create_watch(...)` — so the recorded evidence is
`unknown`, and `_receiver_backed_candidates` attributes nothing to `unknown` by
design.

I measured the alternative before choosing. Admitting `unknown` (excluding only
builtin receivers and wrong-class/wrong-module ones) returns those 2 — and also
returns the 700 dict `.get()` edges, the 179 file-handle `.close()` edges and
the 33 `thread.start()` edges. `GraphStore.close` would stay at ~365 and
`ConnectionPool.get` in the hundreds: the blocker would not be fixed. So the
strict rule stands, and the real fix for the lost callers is inferring a return
type for an unannotated factory, which I have not attempted here and have filed
as follow-up.

### Where else the same fallback lives

* **`callers_of`** — changed. The measured one.
* **`tests_for`** — changed. `get_transitive_tests` has its own bare-*source*
  fallback, and `TESTED_BY` is minted from the test's own `CALLS` edge with the
  metadata copied, so 5 347 bare-source `TESTED_BY` edges in this graph carry
  receiver evidence. Without the guard, `'  a  '.strip()` in a test counts as
  coverage of a module-level `strip`. (Note it only reaches methods' *modules*:
  the fallback looks the bare source up by the qualified tail, so
  `ConnectionPool.get` is looked up as `"ConnectionPool.get"`, never `"get"`.)
* **`references_to`** — changed for consistency, **no measured effect**. The
  parser stamps receiver evidence on `CALLS` only (`TESTED_BY` inherits it); no
  `REFERENCES` edge in this graph carries any.
* **`get_impact_radius`** — **not changed, and there is no fallback there to
  change.** Both engines — the SQLite best-score relaxation and the NetworkX
  one — traverse exact `qualified_name` joins, so a bare target such as `get`
  matches no node and never enters the traversal.
* `refactor.py`'s rename-preview and dead-code detection call
  `search_edges_by_target_name` too. Left alone deliberately: there a bare match
  keeps a symbol *out* of the dead-code list, so over-inclusion currently fails
  safe. Narrowing it is a behaviour change that wants its own measurement, and
  it is filed as follow-up.

One helper, not two: `GraphStore.receiver_evidence_admits` runs the single
candidate being asked about through the existing `_receiver_backed_candidates`.

### Tests

`tests/test_receiver_evidence_read_path.py`, 12 cases, all asserting through
`query_graph` rather than through the store — the store was already right.
**10 of the 12 fail on `3bca5cf`**, each for the diagnosed reason. The two that
pass on both commits are the positive controls (a matching class receiver and a
uniquely named plain call must keep their callers).

### PR body

Row 2 of the "one fixed corpus built three times" table was labelled as what
`callers_of` returns and did not measure that. It is relabelled to what it does
measure — distinct source nodes of qualified `CALLS` edges in the store — and a
third row carries the real tool-level number. A new **Disclosures** section
folds in the four items the review asked for: the add/remove ledger (+787
justified, −1 013 unjustified, from 1 068 added and 1 294 withdrawn), the 280
previously-correct attributions withdrawn together with the −6.7% drop in
`TESTED_BY` edges whose source names a node (6 603 → 6 160), the fact that
`audit_call_attribution`'s denominator only counts edges that name a node — so a
resolver that attributes nothing scores 100% and the figure has to be read next
to the ledger — and the undocumented `IMPORTS_FROM` `extra` change (197 edges
lose `import_resolution` / `python_module` / `import_candidates` /
`import_candidate_count` / `import_candidates_truncated`; stamped total 1 159 →
962).

### Not fixed here, filed as follow-up

* The six new flask false positives: builtin `open()` reached through a
  `TYPE_CHECKING`-guarded relative import, and a stdlib `import json` shadowed
  by a repository package named `json`.
* The flow blow-up: `flow_memberships` 1 075 → 4 179, mean flow size 4.9 → 22.3
  nodes, max 72 → 730, `get_affected_flows` minimal response 1 523 → 5 143 bytes.
* Recovering the unannotated-factory callers lost above.
* Deciding whether `refactor.py`'s bare-name matches should take the same
  constraint.

### Gates

```
pytest tests/ -q                                 3306 passed, 9 skipped, 2 xpassed
ruff check code_review_graph/                    All checks passed
mypy code_review_graph/ --ignore-missing-imports --no-strict-optional
                                                 Success: no issues in 72 source files
bandit -r code_review_graph/ -c pyproject.toml   No issues identified
```

tirth8205:
Correction and final numbers.

**The earlier claim was wrong.** The 74e626c commit message said `GraphStore.close 343 -> 365 -> 189 (189 true; 176 false dropped)`. That is false: a large share of those 176 were genuine callers. The receiver filter was too strict, and the stated cause was also wrong. The dominant loss was not unannotated factories. It was `store, root = _get_store()`: the reader gave up on tuple-unpacking targets even though `_get_store` is annotated `-> tuple[GraphStore, Path]`.

**0eeb9bc teaches the reader four binding shapes it was skipping**: tuple and list unpacking through an annotated return type (including nested targets and `with ... as (a, b)`), dotted-path constructors, constructor-expression receivers, and module-local factories with a return annotation. No return type is inferred from a function body.

Measured on one fixed corpus built by each version:

| query | staging | before 0eeb9bc | after |
| --- | --- | --- | --- |
| callers_of(GraphStore.close) | 320 | 182 | 251 |
| tests_for(GraphStore), direct | 419 | 305 | 358 |
| callers_of(_sanitize_name), control | 28 | 28 | 28 |
| callers_of(ConnectionPool.get) | 409 | 5 | 5 |
| callers_of, repository total | 11198 | 7854 | 8237 |

**On the calls still dropped against staging.** I sampled 60 and opened each one. 55 are plainly wrong on staging: `subprocess.run(...)` counted as a benchmark's `run`, `some_dict.get(...)` as `ConnectionPool.get`, `tmp_path.resolve()` as `CodeParser.resolve`, `self.tmp.close()` as `GraphStore.close`. Two were real, three were unclear. So the remaining gap against staging is overwhelmingly the removal of wrong answers, not the loss of right ones.

Unchanged by this commit: relative imports 222/222, attribution precision 10487/10491, forget parity. Gates: 3335 passed, 9 skipped, 2 xpassed; ruff and mypy clean.

Known gaps, filed rather than fixed here: the two real callers of `GraphUpdateHandler.start` and `.stop` are still unreachable, because `handler = _create_watch_handler(...)` has no annotation to read (crg-6db); the flask false positives and the flow blow-up (crg-59w).

PR #1000, opened 2026-09-16
fix(gate): make the promotion coverage gate's rows survive a hand check

tirth8205:
## What this fixes

PR #996 cannot merge because the coverage gate inside it is measurably wrong.
Measured against `origin/main` on `origin/staging` (135 changed files, 639
changed nodes), it reported **212** changed functions with no test. **82 of
those rows were wrong**, and a further 16 named code nothing can test
directly.

Controlled A/B — same tree (`origin/staging` @ `9c13f4c`), same base, same
graph build, only the analyser swapped:

| | before | after |
| --- | --- | --- |
| changed files | 135 | 135 |
| changed nodes | 639 | 639 |
| test-gap rows | **212** | **114** |
| rows that are test code | 80 | 0 |
| rows added by the fix | — | 0 |

## The four defects

**1. `parser.py` never set `is_test` on Class nodes.** No `kind="Class"`
construction site set the flag, so every `TestX` in `tests/` read as changed
production code — 36 rows. `parse_bytes` now settles it once, after
extraction, for every language and every future extractor. The rule is
deliberately narrower than `_is_test_function`: location and an explicit
framework annotation only. `^Test` alone would mis-mark a production
`TestHarness` and hide it.

**2. Helpers and fixtures in test files.** 44 rows were `conftest` fixtures
and test helpers (`built_repo`, `_git`, `flaky`). `analyze_changes` now skips
any node whose **repo-relative** path is a test path — relative, because
`_is_test_file` searches for `test_` anywhere in the string and would answer
"yes" for every `.py` file under a CI workspace whose directory name contains
it.

**3. `TESTED_BY` edges stored under a bare source were invisible.**
`changes.py` called `get_edges_by_source(node.qualified_name)` with no
fallback. `detect_changes_func` itself was reported untested while having
three tests: `tests/test_changes.py` reaches it as
`from code_review_graph.tools import detect_changes_func`, a re-export, so the
import edge points at `tools/__init__.py` while the definition is in
`tools/review.py` — and endpoint resolution left the edge bare. The gap scan
now falls back to `get_transitive_tests` at depth 0, and that function's
evidence-gated bare-name fallback gained a one-hop re-export tier, consulted
only when the direct evidence is inconclusive so it can never overrule a
direct match.

**4. The gate failed open.** With the database deleted it printed
`changed_files=135 changed_functions=0 test_gaps=0` and exited **0**. It now
verifies that every changed file the indexer collects is in the graph at the
hash it currently has on disk, and exits 2 otherwise. Files the indexer
collects but that parse to nothing (a plain YAML workflow) are not counted as
missing — the check parses them to find out rather than guessing.

## Also fixed in the report

- It read `file_path` and `line` from rows that carry `file`, `line_start`
  and `line_end`, so every location printed `?`.
- It rendered `gaps[:50]` with a footer computed from a different number
  while the analysis defaulted to 25 rows. Rendered rows, total and footer
  now come from one set of numbers, and `CRG_MAX_CHANGED_FUNCS` is raised so
  the total is the real total (the 500 default reported 177 where the
  uncapped count was 212).
- 16 rows were closures — `walk`, `record`, `bind`, `sync_tested_by`.
  Nothing outside their enclosing function can call them. They are now
  reported once, against the enclosing function, which is in the same change
  by construction; nothing disappears.

## Trivially satisfiable tests: detected, reported, not blocking

A `@pytest.mark.skip` test and an assertionless `def test_x(): my_func()`
both clear a "does a test reach this function?" gate. The gate now finds
both and lists them in a separate section, and `--strict-assertions` makes
them fail the run. They do not block by default: "this test asserts nothing"
is a source-text heuristic, and a test that delegates its assertions to a
helper would be accused wrongly. Across 3051 test nodes in this repository it
finds 2 unconditionally skipped and 20 assertionless, nearly all of them
language fixtures rather than project tests. `skipif` and `skipUnless` are
deliberately *not* counted — a test skipped only where igraph is absent is a
real test everywhere else.

## Hand check

14 flagged functions checked by hand against the suite: **14 of 14** have no
test that calls them directly. 12 of 14 have no coverage at all within two
call hops; 2 (`_replace_file_data`, `get_file_hashes`) are reached
indirectly. Across all 114 remaining rows, 47 are reached by a test within
two hops — run, but not checked. The report now says which is which instead
of calling all 114 untested.

## Tests

`tests/test_gap_gate.py`, 26 tests, one or more per defect: class marking
(including that a production `TestHarness` stays unmarked and a `@Test`
class outside a test path is marked), the test-file skip, the bare-source
fallback (and that it stays evidence-gated — a name match alone is still a
gap), the closure rule, the report's columns and footer arithmetic, the
weak-test detector, and the fail-closed paths for a missing and a stale
graph.

## Gates

- `pytest tests/ -q` — 3361 passed, 9 skipped, 2 xpassed
- `ruff check code_review_graph/` — clean
- `mypy code_review_graph/ --ignore-missing-imports --no-strict-optional` — clean
- coverage 85.52% (threshold 65%)


tirth8205:

PR #1002, opened 2026-09-16
Make the demo safe: cap the flagship default, fix file:// visualize and phrase search, add docs/DEMO.md

tirth8205:
Material to make next week's live demo safe, and fixes for the four things
that would most likely break on stage. Everything here was measured on a
fresh clone of [django](https://github.com/django/django) at `8cbdd4a`
(3,005 parsed files) from a venv that had never seen the tool.

## 1. `get_impact_radius` at its default `detail_level`

`detail_level` defaults to `"standard"`, and the standard branch emitted
**every** connecting edge, changed node and impacted file with no cap.
`max_results` only ever reached the impacted-node list, and
`get_impact_radius_tool` did not expose it at all. Every `file_path`, every
`qualified_name`, both endpoints of every edge and the name of every File
node also carried the repository's absolute path.

One call, one changed file (`django/db/models/query.py`):

| | est. tokens | chars | edges | impacted files |
|---|---:|---:|---:|---:|
| before (`origin/staging`) | **410,834** | 1,643,333 | 1,437 | 481 |
| after, tool default (`max_results=100`) | **21,423** | 85,692 | 100 | 83 |
| after, same node budget (`max_results=500`) | **63,854** | 255,414 | 200 | 200 |

The third row holds `impacted_nodes` at 500, the same as before, so that
84% drop is the caps and the paths — not a smaller node list.

**1,083,461 of those 1.6M characters were the repeated absolute prefix**:
66% of the payload, at this checkout's depth. That is also why the same call
was reported at 178,244 tokens elsewhere — the number moved with how deep
the checkout sat on disk. The prefix is now emitted once, as `repo_root`.

That path change fixes a correctness bug too. Nine qualified names were
being cut off at the 256-character `_sanitize_name` limit *by the prefix
alone*; feeding one back into `query_graph` returned `result_count 0` or
`ambiguous`. Relative names round-trip (verified both directions).

- Edges, changed nodes and impacted files are each capped by `max_results`
  under their own hard ceiling (200/200/200; nodes 500), with `total_*` and
  `*_omitted` reported so nothing is silently dropped.
- Kept edges are ranked by their most-impacted endpoint, so capping drops
  the periphery rather than an arbitrary prefix.
- `max_results` is now on `get_impact_radius_tool` (default 100).

`get_impact_radius_tool` leaves `QUERY_OWNED_UNBOUNDED` in
`tests/test_token_budget.py` and gets a real worst-case ceiling — on that
fixture, a whole-repo diff at depth 5 measured **456,345 tokens before,
20,513 after**. `tests/test_impact_payload.py` (new, 8 cases) pins the caps,
the truncation counters, the relative-path contract and the round-trip.

## 2. `visualize` was blank offline from a `file://` URL

Confirmed, and it was the `integrity` attribute on the *local* script tag. A
page opened off disk gets an opaque origin per file, so the vendored D3 is
not eligible for integrity validation and is blocked; the `document.write`
CDN fallback then fires, which means the page only ever worked because it
silently reached d3js.org. Headless Chrome, generated `graph.html`, `file://`:

| | rendered edges | D3 came from |
|---|---:|---|
| with `integrity`, offline | **0** (blank) | nothing |
| with `integrity`, online | 6 | d3js.org |
| without `integrity`, offline | 6 | vendored file |
| without `integrity`, online | 6 | vendored file |

The bytes are still verified — `_write_d3_asset` checks them against
`D3_SRI_HASH` before writing, a stronger guarantee than a browser check on a
file this process just produced. The cross-origin CDN tag keeps its SRI hash
and `crossorigin`, so the CLAUDE.md security invariant is unchanged.
`visualize --serve` was never affected.

## 3. `semantic_search_nodes` returning nothing for "password hashing"

Confirmed, and the cause is not semantics. `_fts_search` wrapped the whole
query in one pair of quotes, which makes FTS5 read it as a *phrase* —
adjacent, in order. On django: `"password hashing"` → 0 results,
`"hash password"` → 5. Same index, same words.

Terms are now quoted individually and joined with AND (prefix-matched at ≥3
characters), with an OR retry when AND finds nothing. Injection is still
impossible: the query is split on non-identifier characters and every piece
is re-quoted, and the expression is still a bound parameter.

```
"password hashing"            -> make_password, render_password_as_hash,
                                 verify_password, check_password
"how are passwords hashed"    -> make_password, encode, _check_encode_args
"database connection pooling" -> test_connect_pool, test_get_connection_pool
```

This does not make the search semantic. It is BM25 over identifiers either
way; it just stops word order deciding whether the tool answers at all.

## 4. The GitHub Action

**Works end to end** — no fix needed, and **this PR is the proof**: the
`review` check passed and `github-actions[bot]` posted its sticky comment on
#1002 with correctly repo-relative symbol paths. It had already succeeded on
every non-fork PR through 2026-09-16 (#990, #993, #997). Running the Action's
own commands locally against django: `update` 7.7 s, `detect-changes` 0.3 s,
render 0.04 s.

One thing that *looks* like a failure and is not: pushing twice in quick
succession leaves the earlier `PR Review Comment` run red with *"PR head does
not match the analyzed commit; refusing to comment"*. That guard fired twice
on this branch. It is the workflow refusing to post an analysis of a
superseded commit — documented in DEMO.md so nobody debugs it on stage.

One papercut found and fixed: rendering the comment *by hand* (the
documented preview path) produced full absolute paths truncated mid-cell,
because `relativize_path` only knows `GITHUB_WORKSPACE`. It now also strips
an exact `cwd` prefix, so a local dry run renders what CI renders.

## 5. `docs/DEMO.md`

Ten minutes, seven steps, exact commands, expected output, a measured timing
and a fallback for each. Honest headline numbers:

| | measured |
|---|---|
| cold `pip install` | 14.8 s |
| build, 3,005 files | 78 s, once |
| incremental update after a one-file edit | 7.7 s |
| any query after the build | 0.12–0.54 s from the CLI |
| MCP stdio handshake, 30 tools, 5 prompts | 0.45 s |
| five-call agent review workflow, minimal detail | **≈1,479 est. tokens** |

That last row is deliberately not the number in CLAUDE.md, which targets
"≤5 tool calls, ≤800 total tokens". The measured workflow on django is 1,479
and the doc says so in those words. It also names the 1.18 GB graph database
and the 95 MB generated `graph.html` before an audience member does, and
carries a paragraph on what the tool does **not** do: structural not
semantic, conservative not precise, blind to dynamic dispatch and
reflection, weak flow detection outside Python/PHP, and not worth it under a
few hundred files.

## Gates

```
pytest        3353 passed, 8 skipped, 2 xpassed
coverage      85.68% (floor 65%)
ruff          clean on code_review_graph/
mypy          clean, 72 source files
```

Two existing assertions changed, both to the new intended contract:
`test_parser.py` and `test_multilang.py` asserted absolute paths in
`impacted_files`, and `test_visualization.py` asserted `integrity` on the
local D3 tag. `uv.lock` carries a one-line sync of the
`sentence-transformers` bound that `pyproject.toml` already has; `uv run`
rewrites it otherwise.

## Not fixed — carried into the demo doc as known rough edges

- The graph database is **1.18 GB** for django, dominated by edge indexes
  storing fully-qualified names. Shrinking it means a schema change.
- `visualize` on django produces a **95 MB** HTML file.
- Search is lexical unless the optional embeddings extra is installed.
- The Action installs from PyPI, so it analyses with the published release,
  not the branch under review.
- Absolute paths still leak from other tools. `query_graph` at standard detail
  measured **12,233 tokens with 175 repetitions of the absolute prefix** for a
  six-caller query on django; `semantic_search_nodes` returns the absolute path
  as the `name` of File nodes. Same class of waste, not swept here — the caps
  and the path change were scoped to impact analysis so the test churn stayed
  reviewable.
- `code-review-graph install` prefers `uvx code-review-graph serve` when `uv`
  is on PATH, so the MCP server an editor launches is the published release
  from a uv cache rather than the environment the user installed into. That is
  the documented priority order in `_detect_serve_command`, left alone here,
  but it surprises anyone who just ran `pip install -e .` — and on a cold cache
  it is a PyPI fetch at the first tool call. DEMO.md now says to check
  `.mcp.json` and warm it beforehand.


tirth8205:

PR #1035, opened 2026-09-17
Treat @Override and JUnit annotations as entry points in dead-code

### LINKED ISSUE #1034: [Bug]: dead-code reports @Override methods as dead, and JUnit lifecycle methods outside test-named files
### code-review-graph version

Measured on `staging` at `f0e4eb7` (post-2.3.8). The numbers below differ from what 2.3.8 shows,
because #1023 has since removed a large part of the same family — see *What #1023 already fixed*.

### Operating system

macOS

### Python version

3.13

### AI platform

claude-code

### Steps to reproduce

**Reproducer 1 — `@Override` in production code.**

`src/main/java/demo/Widget.java`

```java
package demo;

import java.util.Objects;

public class Widget {
  private final String id;

  public Widget(String id) { this.id = id; }

  public String getId() { return id; }

  @Override
  public int hashCode() { return Objects.hash(id); }

  @Override
  public boolean equals(Object other) {
    return other instanceof Widget && Objects.equals(id, ((Widget) other).id);
  }
}
```

`src/test/java/demo/WidgetTest.java`

```java
package demo;

import static org.junit.jupiter.api.Assertions.assertEquals;
import org.junit.jupiter.api.BeforeEach;
import org.junit.jupiter.api.Test;

class WidgetTest {
  private Widget widget;

  @BeforeEach
  void setUp() { widget = new Widget("a"); }

  @Test
  void keepsItsId() { assertEquals("a", widget.getId()); }
}
```

```bash
git init && git add -A && git commit -m repro
code-review-graph build --repo . --data-dir /tmp/dc-data
code-review-graph dead-code --repo . --data-dir /tmp/dc-data
```

**Reproducer 2 — a JUnit lifecycle method in a file the test-file exclusion does not cover.**
A shared harness deliberately lives in `main` so that several test modules can depend on it:

`src/main/java/demo/AbstractWidgetHarness.java`

```java
package demo;

import org.junit.jupiter.api.BeforeEach;

/** Shared harness, lives in main so several test modules can depend on it. */
public abstract class AbstractWidgetHarness {
  protected String fixture;

  @BeforeEach
  void prepareFixture() { fixture = "ready"; }
}
```

**Reproducer 1**, `staging` at `f0e4eb7`:

```
Dead code: 2 item(s); showing 2
  [Function] Widget    (src/main/java/demo/Widget.java:8)
  [Function] hashCode  (src/main/java/demo/Widget.java:16)
```

`hashCode` carries `@Override`. Its caller is `Object`'s contract — `HashMap`, `Objects.hash`,
anything that hashes a `Widget`. It is not dead, and **no hierarchy walk can rescue it**, because the
base is the JDK and therefore outside the graph. (`Widget` at line 8 is the constructor; that is a
separate defect, noted at the end.)

**Reproducer 2**, same build:

```
Dead code: 2 item(s); showing 2
  [Class] AbstractWidgetHarness   (src/main/java/demo/AbstractWidgetHarness.java:6)
  [Function] prepareFixture       (src/main/java/demo/AbstractWidgetHarness.java:9)
```

`prepareFixture` carries `@BeforeEach`. Its caller is the JUnit runner.

The annotation is already stored on the node in both cases:

```
sqlite> SELECT name, extra FROM nodes WHERE kind IN ('Function','Test') AND extra != '{}';
hashCode       | {"decorators": ["Override"]}
equals         | {"decorators": ["Override"]}
prepareFixture | {"decorators": ["BeforeEach"]}
```

`_is_entry_point` already consults `extra.decorators` through `_has_framework_decorator`. The
patterns list in `flows.py` covers Spring, Django, Click, Celery, Angular, Express and
`pytest.fixture` — but has no entry for `Override`, and none for JUnit or TestNG.

### What #1023 already fixed

Against stock **2.3.8** this reproducer reports **four** items, adding `[Class] WidgetTest` and
`[Function] setUp`. On current `staging` both are gone: `find_dead_code` now reads test-ness from the
path relative to the repository root, and `src/test/java/demo/WidgetTest.java` is recognised. So the
part of this family that lives in test-named files is already handled.

What remains is the part that does not: `@Override` in production code, and JUnit annotations on
methods that sit outside a test-named file — reproducer 2, which is a real pattern in multi-module
Java builds.

### Root cause and fix

Two patterns added to `_FRAMEWORK_DECORATOR_PATTERNS` in `flows.py`:

```python
# An override implements a supertype contract, so its caller is that contract:
# a supertype-typed reference, a framework, or the runtime itself for
# Object.hashCode/equals/toString. Anchored so it cannot match unrelated names
# such as override_settings, handled below.
re.compile(r"^Override$"),
# JUnit / TestNG lifecycle and test methods are invoked by the runner.
re.compile(
    r"^(Test|Before|After|BeforeEach|AfterEach|BeforeAll|AfterAll"
    r"|BeforeClass|AfterClass|ParameterizedTest|RepeatedTest|TestFactory)$"
),
```

Both reproducers drop from 2 findings to 1.

Measured on a real 125-file multi-module Java library, on `staging`:

| | `staging` | with the patch |
|---|---|---|
| symbols reported dead | **174** | **120** (−31 %) |
| of those, carrying `@Override` | 54 | **0** |
| findings *added* by the patch | — | **0** |

The removal set is exactly the 54 `@Override` methods, nothing else. Spot-checking them: `getHeader`
and `getStatusCode` implement the Azure Functions `HttpResponseMessage` interface;
`getSupportedAnnotationTypes` overrides `javax.annotation.processing.AbstractProcessor`. All are
called by a framework, none from the repository.

On this particular library the JUnit patterns change nothing — its tests all live in test-named
files, which #1023 already covers. Reproducer 2 is the case they do cover.

### Additional context

**What the patch deliberately does not fix.** Of the 120 that remain, most are `public` with no
`@Override`. In a library those are *unknowable*, not dead: the consumers are in other repositories
and a single-repository graph cannot see them. I verified one — a public test-helper class reported
dead, used by six other repositories in the same workspace. The fix there is not detection but
labelling, and there is a precedent in this codebase: the `confidence` key added by #884 explains why
a query result is empty instead of presenting the zero as proof. `dead-code` on a library could say
the same thing about its public surface.

**Two further false-positive families the reproducers surface, which this patch does not fix:**

1. **The harness class itself is still reported.** `AbstractWidgetHarness` has no class-level
   annotation, and the class-level exclusion only checks `_has_framework_decorator`. A class whose
   methods are all entry points is not dead.
2. **Constructors are reported.** `Widget` at line 8 is the constructor, and `new Widget("a")` is
   called in the test. The edge exists, but its target is the bare name `Widget`, never the
   constructor node `Widget.java::Widget.Widget`, which therefore has no incoming edge at all.

**A cheap detector for a related invention**, found while looking into this and worth its own check:
`SELECT COUNT(*) FROM edges WHERE kind='CALLS' AND source_qualified = target_qualified` — a method
calling itself. On the same library that is **39** edges, and of eight sampled at random seven are
not recursion: they are `super.toString()` and casts used as receivers (`((SomeType) x).getY()`),
where the parser records **no receiver at all** (`extra` is `{}`) and a target is still built on the
enclosing class. That is #984's invented target with zero evidence attached, and a self-loop is a
one-line way to find them.

### Related

- **#1023** — removed the test-file half of this family. This issue is the remainder.
- **#984** — call targets are invented when they cannot be worked out. The self-loop family above is
  an instance of it.
- **#884** — the `confidence` key on empty query results; the same idea applied to `dead-code` would
  cover the unknowable public surface.

### Patch

Against `staging` at `f0e4eb7`.

<details>
<summary><code>flows.py</code> — two patterns in <code>_FRAMEWORK_DECORATOR_PATTERNS</code></summary>

```diff
--- a/code_review_graph/flows.py
+++ b/code_review_graph/flows.py
@@ -42,8 +42,20 @@
     re.compile(r"receiver", re.IGNORECASE),
     re.compile(r"api_view", re.IGNORECASE),
     re.compile(r"\baction\b", re.IGNORECASE),
+    # An override implements a supertype contract, so its caller is that
+    # contract: a supertype-typed reference, a framework, or the runtime itself
+    # for Object.hashCode/equals/toString. It is never dead by virtue of having
+    # no explicit call site, and the base is often outside the graph (the JDK),
+    # so no hierarchy walk can rescue it either. Anchored so it cannot match
+    # unrelated names such as override_settings, handled below.
+    re.compile(r"^Override$"),
     # Testing
     re.compile(r"pytest\.(fixture|mark)"),
+    # JUnit / TestNG lifecycle and test methods are invoked by the runner.
+    re.compile(
+        r"^(Test|Before|After|BeforeEach|AfterEach|BeforeAll|AfterAll"
+        r"|BeforeClass|AfterClass|ParameterizedTest|RepeatedTest|TestFactory)$"
+    ),
     re.compile(r"(override_settings|modify_settings)", re.IGNORECASE),
     # SQLAlchemy / event systems
     re.compile(r"(event\.)?listens_for", re.IGNORECASE),
```

</details>

aala-conga:
Fixes #1034.

`dead-code` reports two families of method whose caller is a contract or a test runner, not a call
site in the repository:

- **`@Override` in production code.** An override implements a supertype contract, so its caller is
  that contract — a supertype-typed reference, a framework, or the runtime itself for
  `Object.hashCode`/`equals`/`toString`. No hierarchy walk can rescue these, because the base is
  often the JDK and therefore outside the graph.
- **JUnit/TestNG annotations outside a test-named file.** A shared harness deliberately lives in
  `main` so several test modules can depend on it, so #1023's test-file exclusion does not cover it.

Both annotations are already on the node — the parser stores them in `extra['decorators']` — and
`_is_entry_point` already consults them through `_has_framework_decorator`. The patterns list covered
Spring, Django, Click, Celery, Angular, Express and `pytest.fixture`, but had no entry for `Override`
and none for JUnit or TestNG. This adds two patterns, both anchored so they cannot match unrelated
names such as `override_settings`, which is handled by its own pattern further down the list.

### Relationship to #1023

Against stock 2.3.8 this looked like a four-item problem; on current `staging` it is a two-item one,
because #1023 already removed everything in the family that lives in a test-named file. What is left
is `@Override` in production code, and JUnit annotations on methods outside such files.

### Measured

A 125-file multi-module Java library, built on `staging` and again with this patch:

| | `staging` | with the patch |
|---|---|---|
| symbols reported dead | **174** | **120** (−31 %) |
| of those, carrying `@Override` | 54 | **0** |
| findings *added* by the patch | — | **0** |

The removal set is exactly the 54 `@Override` methods, nothing else moved. Spot-checking them:
`getHeader` and `getStatusCode` implement the Azure Functions `HttpResponseMessage` interface,
`getSupportedAnnotationTypes` overrides `javax.annotation.processing.AbstractProcessor` — all called
by a framework, none from the repository.

On this library the JUnit patterns change nothing, because its tests all live in test-named files.
The second reproducer in the issue is the case they cover.

### Tests

Seven cases, in the two modules that share the predicate — `detect_entry_points` in `flows.py` and
`find_dead_code` in `refactor.py`, which both reach it through `_has_framework_decorator`.

Five of them fail on `staging` and pass with the patch:

```
tests/test_flows.py::TestFlows::test_detect_entry_points_java_override
tests/test_flows.py::TestFlows::test_detect_entry_points_junit_lifecycle
tests/test_flows.py::TestFlows::test_detect_entry_points_junit_test_annotation
tests/test_refactor.py::TestFindDeadCodeJavaAnnotations::test_override_is_not_dead_code
tests/test_refactor.py::TestFindDeadCodeJavaAnnotations::test_junit_lifecycle_outside_a_test_file_is_not_dead_code
```

The other two pass in both directions on purpose — they are guards rather than proofs:
`test_override_pattern_is_anchored` (a decorator named `Overridable` must not match `^Override$`) and
`test_a_genuinely_unreferenced_method_is_still_reported` (the patch must narrow the false positives,
not silence the query).

Each annotated method in the `flows` tests is given an incoming `CALLS` edge, so that rule 1 of
`detect_entry_points` — no callers — cannot carry the assertion on its own and the annotation is the
only thing that can make the method an entry point. Note that `detect_entry_points` skips test files
unless `include_tests=True`; the JUnit cases are therefore written in ordinary files, which is also
the case the patch is for.

Full suite, fresh clone of `staging` at `f0e4eb7`, Python 3.13, `pip install -e ".[dev]"`:

| run | result |
|---|---|
| `staging`, untouched — baseline | `3963 passed, 773 skipped, 2 xfailed, 2 xpassed` (349 s) |
| + the patch + these tests | ``3970 passed, 773 skipped, 2 xfailed, 2 xpassed` (339 s)` |

### Not addressed here

Two further false positives the reproducers surface, both out of scope for this change: the harness
*class* is still reported, because the class-level exclusion only checks `_has_framework_decorator`
and not whether every method inside is an entry point; and constructors are reported, because a
`new Widget(...)` edge targets the bare name `Widget` rather than the constructor node, which
therefore has no incoming edge at all.


tirth8205:
A dead `@Override` on a base type outside the repo is no longer reportable at all, and that out-of-repo case is the rationale the PR body gives.

```java
public class MultiHandler extends com.external.BaseHandler {
    @Override public void handleLive()  { doWork(); }
    @Override public void handleStale() { doWork(); }
}
// App.main constructs MultiHandler and calls handleLive() only
```

```
staging:  [Class] App
          [Function] handleStale (MultiHandler.java:12)
this PR:  [Class] App
```

`handleStale` is dead and nothing reports it. `App` references the class so the class is alive, and the base is out of graph so there is no declaration side to fall back on. An in-repo base still surfaces the interface declaration, so that half is mitigated. The "Not addressed here" section should say this.

Half of the JUnit alternation cannot fire. `Test`, `ParameterizedTest`, `RepeatedTest` and `TestFactory` are already in `parser._TEST_ANNOTATIONS`, so the parser sets `is_test=1` and both consumers drop those nodes before `_has_framework_decorator` runs. `refactor.py` has `if node.is_test or _is_test_file(node.file_path): continue`, and `flows.py` has the same guard under the default `include_tests=False`. A bare `@Test` method in `src/main` is already absent from `dead-code` on staging, so there was no false positive to remove. The one JUnit form that does reach the predicate is excluded by the anchor. The parser strips the leading `@` and keeps the arguments, so `@Test(expected = IllegalStateException.class)` is stored as `Test(expected = IllegalStateException.class)` and `^Test$` misses it. It is still reported dead with the patch. `^Test\b` would catch it, and the same anchoring misses `@org.junit.jupiter.api.BeforeEach` and `@DataProvider(name = "rows")`. The comment says TestNG, but `BeforeMethod`, `AfterMethod`, `BeforeSuite`, `AfterSuite`, `BeforeTest`, `AfterTest` and `DataProvider` are not in the pattern. Only `Test`, `BeforeClass` and `AfterClass` overlap with TestNG.

`test_detect_entry_points_junit_test_annotation` is green against a node the parser cannot produce. `_add_func` defaults `is_test=False` while setting `decorators=["Test"]`, but a real build stores that method as `kind=Test, is_test=1`. So one of the five "fails on staging, passes with the patch" proofs shows nothing about real behaviour. The `Override` and `BeforeEach` tests are fine, both are genuinely `is_test=0`.

The flows-side cost is real and the body does not size it. Every `@Override` becomes a flow root even when it has callers. On apache/commons-lang flows go from 897 to 955 and `flow_memberships` from 5,666 to 6,264, with 58 new roots of which 32 are strict sub-paths of flows that already existed. In the default top-50 `list_flows` view, 8 real library entry points are evicted, among them `ConcurrentUtils.initializeUnchecked`, `ExceptionUtils.hasCause` and `ThreadUtils.getAllThreads`, replaced by `GenericArrayTypeImpl.equals`, `StrTokenizer.hasNext` and similar. On jackson-databind flows go from 1,909 to 2,270. `detect-changes` on a one-line edit to `ToStringStyle.java` returns 19 affected flows instead of 16, with `appendDetail`, `append` and `appendFieldStart` each listed twice under the same name and nothing in the output to tell them apart. The JSON payload grows 14.7%.

`flows.py:62` still carries `re.compile(r"@(Override|OnLifecycleEvent|Composable)", re.IGNORECASE)`, which can never match because the parser strips the `@` before storing. The file now has two `Override` patterns and one of them is dead.

The fix is Java-annotation-only while the body frames the problem generically. Kotlin `override fun`, C# `public override string`, TypeScript `override` and Python's `@override` produce identical `dead-code` output on both sides. The first three store no decorator at all and Python stores lowercase `override`.

The core claim reproduces well outside your own tests. On google/gson `dead-code` drops 8 unique symbols and adds none. On commons-lang it goes from 233 to 203, 30 removed and 0 added, and every removal opens to a real `@Override`. Non-JVM collateral is zero and this repo's own graph is byte-identical on both sides. It merges clean into `staging` with ruff, mypy and the full suite green, and reverting only `flows.py` in a merged tree makes exactly the five tests you name fail, so the patch is what moves them.

To get this mergeable: either drop `Test|ParameterizedTest|RepeatedTest|TestFactory` or loosen the anchor to `\b` so `@Test(expected=...)` is covered, and either add the TestNG names or drop TestNG from the comment. Rebuild `test_detect_entry_points_junit_test_annotation` from a parsed node instead of a hand-built one, or remove it. State in the body that an `@Override` on an out-of-repo base is no longer reportable at all, and that flows grow roughly 6% to 19% on JVM repos. Delete the dead `@(Override|OnLifecycleEvent|Composable)` line while you are in the file.

Issue #984, opened 2026-09-15
call targets are invented when they cannot be worked out, and the invented ones look exactly like real ones


stk0vrfl0w:
Found in a read-only audit of v2.3.8. Three languages, three different pieces of code, the same mistake.

## What happens

Several pieces of resolution added since 2.3.6 will rewrite a target to a name inside the repository that the source never actually referred to. Leaving the target as a plain name, as before, at least said honestly that it could not be worked out. An invented name says the opposite, and nothing later can tell the two apart.

In Java and Kotlin, `_resolve_typed_method_target` (`code_review_graph/parser.py:5470`) builds `<file>::<Type>.<method>` at `:5507` whenever it knows the type of the thing the method was called on, without checking that the type actually has that method.

In Rust, a call to something from the standard library is rewritten as if it lived in the local crate.

In Python, this is the second half of #981.

## Reproduction

Java, one file, no package, no imports:

    public class One {
        public int run() { Helper h = new Helper(); return h.hashCode(); }
    }
    class Helper { public int build() { return 1; } }

    CALLS -> One.java::Helper.hashCode   (matches no node)
             extra = {"receiver_resolution": "typed_receiver"}

`hashCode` comes from `Object`; `Helper` does not define it. The trigger is that `Helper` is declared at the top level of the same file, so the code knows the type -- with `Helper` in a separate file and no import, the target correctly stays a plain name.

Rust, a Cargo crate whose `src/lib.rs` contains `let v = Vec::new();`:

    CALLS -> src/lib.rs::new   (matches no node)

## Why it matters

`callers_of`, `get_impact_radius`, `tests_for` and `find_dead_code` all trust a target that has a file path in it. A plain name is something every one of them already treats as coming from outside the repository. A name with a file path in front of it passes every check they make, while pointing at nothing. In the Rust case the invented name is worse than what it replaced, because a target containing `::` is skipped by the later pass that repairs plain names.

## Suggested fix

When a target cannot be worked out with confidence, leave it as a plain name rather than building a name that looks right. `confidence_tier` already exists and would say this well, but only one of the four resolvers currently sets it. For the Java and Kotlin case specifically, checking that the type really does have the method -- or inherits it from something the graph knows about -- would be enough.

C# already avoids this: it writes `Class::method` for `scoped_resolver` to pick up, and that only rewrites when it finds a real node. That is the pattern the others want.

[...]

PR #1036, opened 2026-09-17
Stop inventing call targets: receiver types, and an INHERITS walk for this.method()

aala-conga:
Refs #984.

`callers_of` on a method returns the wrong answer in three different ways, all of which come from the
same habit: when a call target cannot be worked out, one is invented. This is five changes that each
replace an invention with either the right answer or nothing.

**This PR is stacked on #1033** and contains its commit, because part 5 cannot walk an inheritance
chain whose intermediate class is unindexed. Merge that one first and this becomes a one-commit diff;
if you would rather have a single PR, #1033 can be closed in favour of this one.

### Parts 1-4 — a target is only accepted when something supports it

| | |
|---|---|
| 1 | a bare target resolves only when **exactly one** candidate is supported by the call site's file or its imports, instead of any same-named node anywhere |
| 2 | the parser records the receiver's declared type as `extra.receiver_type` even when it cannot resolve a target itself |
| 3 | the resolver filters candidates by that type |
| 4 | a known type that names no candidate is **evidence the callee is external**, so nothing resolves, rather than falling back to same-file candidates |

Part 4 matters more than it looks. Without it, `Objects.equals(id, other.id)` inside a class that
declares its own `equals` resolved to `Widget.equals -> Widget.equals` — a self-loop, the invention
part 1 exists to prevent, coming back through the fallback.

### Part 5 — `this.method()` when the method is inherited

For a `this.` call the parser builds `<file>::<EnclosingClass>.<method>`. That is right when the class
declares the method and wrong when it inherits it: the target names no node at all, so the edge points
at a phantom while `callers_of` on the real declaration returns zero. Parts 1 and 4 cannot help —
part 1 deliberately exempts `this`/`self`, and part 4 would supply the enclosing class as the
receiver's type, which is exactly the wrong answer.

A post-build pass walks `INHERITS` upwards and retargets when **exactly one** ancestor declares the
method, tagging the edge `confidence_tier = INFERRED` with
`extra.inherited_self_resolution = "ancestor"`. Ambiguous or unreachable cases are left untouched.

It is called from `postprocessing.py::_resolve_bare_endpoints` and from both paths in
`tools/build.py`, which is where `resolve_bare_call_targets` is already called. All three matter: I
first wired only the `tools/build.py` pair, and a graph then held resolved or phantom targets
depending on whether it had last seen a full build or an incremental update.

### Measured

A 125-file multi-module Java library, and an 869-file Angular/TypeScript project, both built on
`staging` and again with this branch.

| | `staging` | this branch | truth |
|---|---|---|---|
| Java, resolvable CALLS edges | 743 (21.8 %) | **1064 (31.2 %)** | — |
| Java, edges carrying `receiver_type` | 71 | **1498** | — |
| Java, self-loop CALLS edges | 184 | **39** | — |
| `Payload.setErrorMessage` | 0 | **9** | 9 |
| `Holder.setErrorMessage` | 8 — **all 8 wrong**, all 3 real ones missed | **3** | 3 |
| synthetic repro, `Envelope.setLabel` | 0 | **2** | 2 |
| synthetic repro, `Mailbox.setLabel` | 2 — both fabricated | **0** | 0 |
| Angular, resolvable CALLS edges | 2571 (8.1 %) | **2928 (9.2 %)** | — |
| Angular, `BaseClient.handleFailure` | 0 — node absent | **4, all `INFERRED`** | 4 |
| part-5 retargets on the Java repo | — | **0** | — |

Four hand-built ground truths, all exact.

The `Holder` row is worth unpacking, because it is the failure in its purest form. The eight
callers `staging` reports are the eight `response.setErrorMessage(...)` calls *inside*
`Holder.java` — a `Payload` receiver, attributed to `Holder` because they
share a file. So all eight are wrong, and the three real callers, which live in a different file, are
missed. The same run reports 0 for `Payload.setErrorMessage`, whose truth is those same nine
calls. Two answers, both confidently wrong, from one missing receiver type.

The Angular one:

```
class BaseClient                         declares handleFailure       (base.ts:39)
  LeafOne extends BaseClient     2 calls (leaf-one.ts:68, :116)
  abstract class MiddleClient<U> extends BaseClient
    LeafTwo extends MiddleClient<Dto>
                                          2 calls (leaf-two.ts:99, :129)
```

The four resolved callers are exactly the four enclosing methods of those four lines. I also sampled
six of the 398 retargeted Angular edges at random and read the source for each: six of six are genuine
`this.method()` calls whose declaration is on an ancestor.

Note that `resolvable CALLS edges` **falls by design** wherever a fix makes the resolver decline, so it
is a floor rather than a score: the part-4 correction above cost 14 edges and was still an improvement.

### Tests

`tests/test_inherited_self_calls.py`, 9 cases, in the shape of `tests/test_python_receiver_evidence.py`
— a small real source tree, parsed, the resolver run, exact targets asserted in **both** directions:

- what must resolve: one level, two levels through the generic abstract intermediate, the `INFERRED`
  labelling, the returned count, and that a second run moves nothing
- what must be left alone, each one a way the pass could invent a target: a class that declares the
  method itself, **two same-named ancestors both declaring it** (the walk matches ancestors by simple
  class name, so a homonym is exactly where it must decline), a call on another object, and a class
  with no `INHERITS` edge
- the wiring, through `run_post_processing`

All 9 fail on `staging`. 7 of the 9 pass with part 5 alone, without the abstract-class commit; the two
that need it say so in their docstring, and I verified that by removing `abstract_class_declaration`
from `_CLASS_TYPES` and re-running.

Full suite, fresh clone of `staging` at `f0e4eb7`, Python 3.13, `pip install -e ".[dev]"`:

| run | result |
|---|---|
| `staging`, untouched — baseline | `3963 passed, 773 skipped, 2 xfailed, 2 xpassed` (349 s) |
| this branch | `3972 passed, 773 skipped, 2 xfailed, 2 xpassed` (323 s) |

### What I would still want your view on

`postprocessing.py` looks like the one true home for a post-build resolver, and the only reason there
are three call sites is that `tools/build.py` calls the resolvers inline in both of its paths instead
of going through `run_post_processing`. Collapsing that duplication is a larger change than this one
and I did not want to smuggle it in here.

Part 2 also reads `package` declarations off disk inside the resolver. It works, but it is a prototype
shortcut: the package belongs on the node at parse time. Say the word and I will move it.


tirth8205:
Part 5 invents targets in exactly the case it exists to prevent. `parents` and `declares` in `resolve_inherited_self_calls` are keyed by simple class name across the whole graph, so two unrelated classes that share a name pool their base classes.

```ts
// src/one.ts   Alpha declares only ping()
export class Widget extends Alpha { run() { this.handle(); } }
// src/two.ts   Beta declares handle()
export class Widget extends Beta { }
```

This branch emits:

```
src/one.ts::Widget.run -> src/beta.ts::Beta.handle
confidence_tier = INFERRED   extra.inherited_self_resolution = "ancestor"
```

staging leaves an honest phantom `one.ts::Widget.handle`. A file-qualified wrong answer tagged INFERRED is worse, because `callers_of`, `get_impact_radius`, `tests_for` and `find_dead_code` all trust it. `test_an_ambiguous_ancestor_name_resolves_to_nothing` guards the parent-side homonym. This is the child-side one, and nothing scopes either map to the caller's file or its import closure.

Part 1 breaks PHP. `has_recv` at graph.py:1755 is `receiver not in ("this", "self", "super", "base")`, which does not cover `$this`, so `$this->execute()` reads as a foreign receiver and part 1 strips `f"{context_file}::{caller_cls}.{bare_name}"`, the only correct candidate. tests/fixtures/sample.php:92 gives `ExtendedRepo.runQueries -> sample.php::ExtendedRepo.execute` on staging and a bare `execute` here. A minimal Repo class calling a private `execute()` through `$this->` drops from 2 resolvable CALLS to 1. The same list at graph.py:1505 turns part 5 off for PHP entirely, so PHP pays the cost and gets none of the benefit. tests/test_multilang.py:1375 misses this because it asserts parse-level target names, not post-resolution targets.

CI lint fails on the head. staging is clean under the identical config:

```
code_review_graph/graph.py:1677:65: E702 multiple statements on one line (semicolon)
code_review_graph/graph.py:1678:63: E701 multiple statements on one line (colon)
```

The title oversells what landed. `_resolve_typed_method_target` is untouched, and #984's own Java snippet still does this, byte-identical on both branches:

```
One.java::One.run -> One.java::Helper.hashCode   extra.receiver_resolution = "typed_receiver"
```

That matches no node. Interface-typed and type-alias-typed TypeScript receivers invent the same way. Please retitle or split so #984 does not get closed on this.

I checked the rest by building real repos on both branches rather than reading the diff. Holder/Payload is as you describe: staging produces two wrong `Holder.setErrorMessage` edges and zero callers for `Payload.setErrorMessage`, and this branch gets all three right. On the VS Code extension sources staging fabricates 11 self-loops from `this.item.dispose()`, `this.changedFiles.clear()`, `d.dispose()` and friends, and this branch declines all 11. On this repo's own 364-file corpus phantom CALLS go from 122 to 103 and all 19 part-5 retargets are correct. Parts 1 to 4 without the `$this` blind spot are a clear net win.

One cost the body does not mention: a TypeScript `any` or `unknown` receiver is now positive evidence the callee is external, since the type is absent from `class_file` and part 4 empties `supported`. `class Pair { label(); combine(other: any) { return other.label(); } }` resolves on staging and stays bare here. Declining honestly is defensible, but every any-typed receiver in a TS codebase stops resolving.

The graph.py conflict comes from base f0e4eb7, which predates the #1021 merge and three after it. Both hunks are textually keep-both, but the second sits inside `_resolve_bare_endpoints` where staging's new `_directory_member_files` expansion feeds the same `import_targets` set your `supported` filter consumes, so the receiver probes want re-running after the rebase rather than a diff read.

To make it mergeable: cover `$this` in both receiver lists, scope the part-5 maps so a simple-name collision cannot resolve, fix the two ruff errors, rebase and re-probe, and settle the #1033 sequencing given two of the nine tests need it. Splitting parts 1 to 4 into their own PR would get most of the value in now.

PR #1039, opened 2026-09-18
fix(opencode): emit OpenCode 2.x plugin definition with id and setup

### LINKED ISSUE #1038: [Bug]: opencode plugin template fails on OpenCode 2.x — V2 loader requires { id, setup }, template exports legacy function
## Summary

`code-review-graph install --platform opencode` (2.3.8) writes `~/.config/opencode/plugins/crg-plugin.ts` from `_opencode_plugin_content()` in `code_review_graph/skills.py`. On OpenCode v2.0.7 the plugin fails on every startup and never runs.

Server log (`~/.local/share/opencode/log/opencode.log`):

```text
message="failed to load plugin" target=.../crg-plugin.ts ref=err_230b6ff8 cause="Cause([Fail(PluginModule.LoadError: Plugin must export a default definition with an id and an effect or setup function. (cause: SchemaError(Expected object at [\"default\"])))])" role=server
```

`opencode plugin list` shows the entry with ID `-` and state `failed`.

## Root cause

The template exports a bare function (`export default (app: any) => { app.on(...) }`) using the removed `app.on` EventEmitter API. The V2 loader requires the default export to be an object with an `id` and a `setup` (or `effect`) function (`Plugin.define({ id, setup })`), so schema validation rejects the module before any hook runs.

Related: #947 covers the `app.on is not a function` failure; #949 ports the template to the V1 returned-hooks shape (`export default CodeReviewGraphPlugin` where the default is still a function). A function default export still fails the V2 loader with the `SchemaError(Expected object at ["default"])` above, so #949 does not fix OpenCode 2.x.

Secondary bug in the same template (also noted in #947): the `run($, cmd)` helper interpolates a full command string as one Bun template (`$\`${cmd}\``), which Bun escapes into a single executable name. Commands must be literal tagged templates or use `node:child_process`.

## Expected

Template emits a V2 plugin:

```ts
import { Plugin } from "@opencode/plugin"
export default Plugin.define({
  id: "code-review-graph",
  async setup(ctx) {
    // session.created via ctx.event.subscribe + `code-review-graph status`
    // file edits via ctx.tool.hook("execute.after") on edit/write/patch + `update --skip-flows`
    // git commit via ctx.tool.hook("execute.before") + `detect-changes --brief`
  },
})
```

using `node:child_process` for subprocesses (no Bun `$` string interpolation), scoped to the plugin instance directory, never blocking a commit.

## Local verification

Rewrote the installed plugin to the shape above: `bun build` passes, `opencode reload` shows `code-review-graph -> active` in `opencode plugin list` and `/api/plugin`, and no new `failed to load plugin` entries appear. `code-review-graph status` itself works (graph present).

## Proposed fix

Update `_opencode_plugin_content()` to emit the V2 shape and update `tests/test_skills.py` assertions (`app.on(` absent, `Plugin.define`, `id: "code-review-graph"`, `ctx.event.subscribe`, `ctx.tool.hook`) accordingly. PR to follow against `staging`.

ranokay:
Closes #1038.

## Problem

The OpenCode installer template (`_opencode_plugin_content()` in `code_review_graph/skills.py`) exports a bare function using the removed `app.on` API. On OpenCode 2.x (verified v2.0.7) the loader rejects it before any hook runs:

```text
PluginModule.LoadError: Plugin must export a default definition with an id and an effect or setup function. (cause: SchemaError(Expected object at ["default"]))
```

`opencode plugin list` shows the entry with ID `-` / state `failed`. #947 covers the `app.on` failure; #949 ports the template to the V1 returned-hooks shape, but a function default export still fails the V2 loader, so 2.x stays broken.

The template's `run($, cmd)` helper has a second bug (also noted in #947): interpolating a full command string into Bun's `$` template escapes it into a single executable name.

## Fix

- Template now emits a V2 definition: `Plugin.define({ id: "code-review-graph", setup(ctx) {...} })` from `@opencode/plugin`.
- `session.created` via `ctx.event.subscribe` + `status` (scoped to the plugin instance directory).
- File edits via `ctx.tool.hook("execute.after")` on `edit`/`write`/`patch` + `update --skip-flows` (V2 has no `file.edited` hook).
- Git-commit detection via `ctx.tool.hook("execute.before")` on `bash`/`shell`/`execute` + `detect-changes --brief`, never blocking.
- Subprocesses via `node:child_process` `execFile` with argv arrays.
- Tests updated to the V2 shape; CHANGELOG entry added.

## Verification

- `uv run pytest tests/test_skills.py -q`: 210 passed.
- `uv run ruff check code_review_graph/skills.py tests/test_skills.py`: clean.
- Generated template `bun build --external @opencode/plugin`: bundles clean.
- Live run on OpenCode v2.0.7: after `opencode reload`, `opencode plugin list` and `/api/plugin` report `code-review-graph -> active` (was `-`/failed), no new `failed to load plugin` entries; `code-review-graph status` works against a real graph.


tirth8205:
This no longer merges into `staging`. The one conflict is `CHANGELOG.md`, and the branch is 51 commits behind.

```bash
git fetch origin
git merge origin/staging
# resolve, then
uv run pytest tests/ -q
git push
```

PR #1061, opened 2026-09-20
fix: scope Codex graph hooks to event cwd

seonghobae:
Codex graph hooks currently discard the event JSON and choose a repository from inherited process cwd. When the launcher cwd differs from the event checkout, they can update the wrong graph.

Add `hook-update` and `hook-status` commands that resolve the repository from a literal absolute payload `cwd`. Missing, malformed, relative, nonexistent, and non-repository paths do not trigger updates. Generated hooks forward JSON and propagate failures when the executable exists. Status delegates to the existing console command, preserving its output and failures.

Validation:
- Focused owner suite: 22 passed, 194 deselected; syntax compilation passed. Full suite and formatting/lint were not verified.
- Coordinator independently ran hook regressions: 5 passed, 1 skipped before the status follow-up. The builder dependency test was skipped in that environment; a separate installed-package spy verified exact downstream update arguments without executing a graph build.
- Final status path was exercised against the real installed console command: an empty GraphStore produced normal zero-count status and exit 0; a missing graph emitted its diagnostic and propagated exit 1.

The package must be upgraded to support these new commands before regenerating hooks. No installed hook configuration was changed. Payload provenance relies on the agent hook channel; filesystem/Git validation is not a separate repository trust allowlist.


seonghobae:
Ready for maintainer review of e4b1d7e5ff64fc95dd5da6b661144a03e3366900. GitHub rejected my RequestReviewsByLogin attempt because this fork contributor lacks permission. Please review the payload cwd boundary and run the repository-required checks through the normal fork-PR path. Validation limits and the package-before-hook-regeneration requirement are recorded in the description.

